PR #72, opened 2013-11-08
exposes customer#orders

costford:
For https://github.com/Shopify/shopify/pull/10832

Exposes customer#orders to retrieve all Orders belonging to a specific Customer.

@csaunders @odorcicd


hrp:
I'm finding that this doesn't work anymore because the `customer_id` is ignored in the Order#find call.

PR #149, opened 2014-09-30
Protect from incidental mutation of URL

dsabanin:
Session.prepare_url is mutating and that makes Session.temp(domain, token) mutate domain without communicating that to the user in any way.


csaunders:
Where is there a production case where you saw this? The code isn't correct, either. From what I'm seeing this just messes around with the provided `url` during the initialization of the session.

It's probably better to just call that function with the URL and assign it to `self.url` in the session object.

PR #166, opened 2014-12-09
Adds policy model

csaunders:
We are going to be exposing the shops various policies (returns, refunds, etc.) through the API.

Please review @benjlcox @jamesmacaulay 

See https://github.com/Shopify/shopify/pull/34708 for more details.


dtKinger:
@csaunders  Have had trouble generating API docs given infrastructure changes but the Policy API doc is getting added to docs now :).

PR #175, opened 2015-03-21
Move pry and thor to dev dependencies

camilo:
Pry and thor bring along a bunch of "friends"  if they slip into a production
bundle it seems  unlikely that either is needed in production, moving them here
to be a dev dependency.

@kevinhughes27 @jamesmacaulay @ibawt makes sense?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,8 +3,6 @@ PATH
   specs:
     shopify_api (3.2.7)
       activeresource
-      pry (>= 0.9.12.6)
-      thor (~> 0.18.1)
 
 GEM
   remote: https://rubygems.org/
@@ -52,5 +50,7 @@ DEPENDENCIES
   fakeweb
   minitest (~> 4.0)
   mocha (>= 0.9.8)
+  pry (>= 0.9.12.6)
   rake
   shopify_api!
+  thor (~> 0.18.1)
--- a/shopify_api.gemspec
+++ b/shopify_api.gemspec
@@ -25,18 +25,18 @@ Gem::Specification.new do |s|
   s.license = 'MIT'
 
   s.add_dependency("activeresource")
-  s.add_dependency("thor", ["~> 0.18.1"])
-  s.add_dependency("pry", [">= 0.9.12.6"])
+
+  dev_dependencies = [['mocha', '>= 0.9.8'],
+                      ['fakeweb'], 
+                      ['minitest', '~> 4.0'],
+                      ['rake'],
+                      ['thor', '~> 0.18.1'],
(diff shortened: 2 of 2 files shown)


kevinhughes27:
yeah that makes sense. Looping in @benjlcox cause he is one of the main users of the thor modules

benjlcox:
Makes total sense :+1:

awd:
CC: @Shopify/api

csaunders:
Will this have an effect on people installing the gem from github and trying to use the `shopify` cli tool?

awd:
I will check that use-case and report back.

camilo:
> Will this have an effect on people installing the gem from github and trying to use the shopify cli tool?

Maybs but that should be dev no? not prod.

awd:
I pulled down master on my dev machine, configured a new connection, booted the console, said yes to `pry` when the console prompted me. Everything appears okay. 

Is there another use-case I should try before giving the API Team "Seal of Approval&trade;"?

csaunders:
Does a random machine count as development? I thought those dependencies were what you needed while building/developing.

`thor` is required and the `shopify` command was working but then I realized that it's because rails requires `thor` hence the dependency is satisfied. If we roll with this we are basically ensuring we have a hidden dependency that's going to result in tickets getting opened.



If we want to keep this dependency out, I say we'll need to do a few things:

- pull all the CLI stuff into it's own package
- bump the major version of `shopify_api`
  - include a post-install message in `shopify_api` telling people to install `shopify_cli` if they want those tools.

csaunders:
@camilo could you withhold releasing a new gem version for this? I'll take care of getting these issues resolved.

awd:
I think we should revert this PR, as it will break things for non rails applications (without `thor`).

camilo:
@csaunders sure can. 

I still thing bringing pry into a prod stack is a bad idea tho, would be great to not have that be a thing.

awd:
@camilo / @csaunders Agreed.

csaunders:
@awd That'll only be a problem if people are using this change in master. If they are using the version on Ruby gems they are fine. We can take care of this in a way that doesn't require a revert.

@camilo yeah. If we can kind of have "two" versions of the install of the gem (dev and prod) that would be awesome. We can pull it off with the `--dev` flag, but I don't like that solution because it's not standard operational procedure for installing a gem.

Adding _another_ gem might be the cleaner approach, but then we have more gems/projects to maintain...

kevinhughes27:
personally I have never understood the point of the cli - I always just use a minimal starting script and drop into pry. Thats my 2 c

PR #177, opened 2015-03-25
Allow options to be passed to shop fetching

anark:
This allows options to be passed when fetching the current shop.  For example

    ShopifyAPI::Shop.current(params: { fields: ['name', 'email'] })

### DIFF
--- a/test/shop_test.rb
+++ b/test/shop_test.rb
@@ -8,8 +8,6 @@ class ShopTest < Test::Unit::TestCase
   end
 
   def test_current_should_return_current_shop
-    fake "shop"
-    @shop = ShopifyAPI::Shop.current
     assert @shop.is_a?(ShopifyAPI::Shop)
     assert_equal "Apple Computers", @shop.name
     assert_equal "apple.myshopify.com", @shop.myshopify_domain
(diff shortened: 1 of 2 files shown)


csaunders:
ping @kevinhughes27 for review.

csaunders:
@anark this is mostly so you can filter out all the extra noise that shopify would send your way eh?

This sounds reasonable to me. Though if you could add some tests that would be fabulous.

kevinhughes27:
yeah looks good to me - a test would be great!

anark:
@kevinhughes27 @csaunders Added a test for Shop.current with params

test/shop_test.rb
```diff
@@ -37,6 +54,9 @@ def test_add_metafield
   end
 
   def test_events
+    fake "shop"
+    @shop = ShopifyAPI::Shop.current
```
kevinhughes27: why repeat these lines? they are in setup

anark: Thank @kevinhughes27 Sorry I am new to test/unit(or testing without Rspec).

I intended to drop setup since test_current_with_options_should_return_current_shop needs a different setup than the other tests.  Now I see that it would be better to leave them in setup and just override them for the options test.

I've updated the pull request.

kevinhughes27:
lgtm! @csaunders I'll leave you the honours of pushing the button

PR #180, opened 2015-04-15
Set errors on Bad Requests

maartenvg:
### Problem
With the introduction of stronger parameters, requests that send bad data (HTTP 400 Bad Request) will often still result a great error description on why the request was rejected. For example, when you send an non-money-formatted string when setting the variant price, the result is:
```
HTTP 400 Bad Request
{
 "errors": {
    "price": "expected String to be a Money"
  }
}
```
Unfortunately, ActiveResource only checks for the presence of the error data in the response body for HTTP codes >= 401, so it is impossible to see these errors via the ShopifyAPI gem.

### Solution
Money patch ActiveResource to look for errors on BadRequests.

@jamesmacaulay @awd 
/cc @EiNSTeiN-

### DIFF
--- a/lib/active_resource/json_errors.rb
+++ b/lib/active_resource/json_errors.rb
@@ -16,7 +16,7 @@ module ActiveResource
       clear unless save_cache
 
       messages.each do |key,errors|
-        errors.each do |error|
+        [errors].flatten.each do |error|
           add(key, error)
         end
       end
--- a/lib/active_resource/validations_ext.rb
+++ b/lib/active_resource/validations_ext.rb
@@ -1,0 +1,18 @@
+module ActiveResource
+  module Validations
+    def self.included(base)
+      base.class_eval do
+        alias_method_chain :save, :bad_request_handling
+      end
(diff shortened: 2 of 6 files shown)


EiNSTeiN-:
Would it be simpler to change the error code returned when UnpermittedParameterType is raised? ([here](https://github.com/Shopify/shopify/blob/master/app/controllers/concerns/admin_area_controller_exception_handler.rb#L35))

maartenvg:
It would definitely be simpler. IMHO, if it does become a 422 (which is the most fitting), the JSON structure should conform the normal validation structure

```json
{
  "errors": {
    "property": ["error description"]
  }
}
```

The difference is that the description is wrapped in an array.

The beauty is that the admin would probably be able to pick it up too, which would then display the error in the right way instead of a generic "something went wrong" message.

EiNSTeiN-:
Sorry, I didn't see this was the `shopify_api` repo. We can definitely make the error conform to the rest of the api.

PR #182, opened 2015-04-21
bump shopify version

csaunders:
All the deps are gone and we need to release a new gem.

Please review @camilo

### DIFF
--- a/lib/shopify_api/version.rb
+++ b/lib/shopify_api/version.rb
@@ -1,3 +1,3 @@
 module ShopifyAPI
-  VERSION = "3.2.7"
+  VERSION = "4.0.0"
 end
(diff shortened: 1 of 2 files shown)


lib/shopify_api/version.rb
```diff
@@ -1,3 +1,3 @@
 module ShopifyAPI
-  VERSION = "3.2.7"
+  VERSION = "4.0.0"
```
camilo: needs tests :trollface:

camilo:
:+1:

dylanahsmith:
@csaunders why was the major version bump done?  There is no entry for it in the changelog.

cc @clayton-shopify

csaunders:
We removed the CLI features from the gem.

PR #184, opened 2015-04-23
Bugfix/orders api order cancellation

csaunders:
Parameters being sent to the orders API were being
attached to the query string instead of added to the body.

As a result their truthiness was being lost and all calls to
cancel an order would result in an email being delivered to
the customer.

----

There is still an underlying issue with the API in that URL Encoded forms with values like `email=false` result in stringy values instead of boolean ones. We should probably patch Shopify anyway, and as a result _could_ throw out this PR. I'd like some thoughts on that as well.

for review @maartenvg @edward 

for context take a look at: https://github.com/Shopify/shopify/issues/38454

### DIFF
--- a/lib/shopify_api/resources/order.rb
+++ b/lib/shopify_api/resources/order.rb
@@ -7,7 +7,7 @@ module ShopifyAPI
     def open;   load_attributes_from_response(post(:open, {}, only_id)); end
 
     def cancel(options = {})
-      load_attributes_from_response(post(:cancel, options, only_id))
+      load_attributes_from_response(post(:cancel, {}, options.to_json))
     end
 
     def transactions
--- a/lib/shopify_api/version.rb
+++ b/lib/shopify_api/version.rb
@@ -1,3 +1,3 @@
 module ShopifyAPI
-  VERSION = "4.0.3"
+  VERSION = "4.0.4"
 end
--- a/test/detailed_log_subscriber_test.rb
+++ b/test/detailed_log_subscriber_test.rb
@@ -36,7 +36,7 @@ class LogSubscriberTest < Test::Unit::TestCase
   test "logging on #find with an error" do
     fake "pages/2", :method => :get, :body => nil, :status => 404
 
-    assert_raise ActiveResource::ResourceNotFound do
+    assert_raises ActiveResource::ResourceNotFound do
       ShopifyAPI::Page.find(2)
     end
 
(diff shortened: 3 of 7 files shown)


lib/shopify_api/resources/order.rb
```diff
@@ -7,7 +7,7 @@ def close;  load_attributes_from_response(post(:close, {}, only_id)); end
     def open;   load_attributes_from_response(post(:open, {}, only_id)); end
 
     def cancel(options = {})
-      load_attributes_from_response(post(:cancel, options, only_id))
+      load_attributes_from_response(post(:cancel, {}, options.to_json))
```
maartenvg: What is `only_id` and why can it be removed?

csaunders: [`only_id`](https://github.com/Shopify/shopify_api/blob/e729054363fd30e584a5fb2fb6b2e92cd41d6e73/lib/shopify_api/resources/base.rb#L81-L83) is a function that would encode only the records ID and make that the contents of the request body. (I only saw empty JSON in the request body)

The ID is encoded in the path so we don't need to worry about it, and our requests don't need to send that; only the data that is going to be acted upon.

csaunders:
Fixed an intermittent test failure. All tests should consistently run green now.

lib/shopify_api/resources/order.rb
awd: Just so I understand correctly, looking for confirmation on this.. The endpoint returns an empty body due to: https://github.com/Shopify/shopify/blob/8ded28a6ed7a791ca5e87c3fcb54cce2316d079d/app/controllers/admin/orders_controller.rb#L335-L343

csaunders: So in the request that was being submitted to Shopify it was an empty JSON object. The params were being encoded in the URL, which is weird for an action like this.

Running the code we just get `true` back, but that might be because of how the `load_attributes_from_response` behaves.

awd:
Nothing left to contribute here, this LGTM

awd:
@csaunders - Is this blocked, are you going to ship it?

csaunders:
Needs a rebase because there's definitely some conflicts going on. I think it's just the version number.

csaunders:
Alright, this will cleanly merge with master. Will merge and release a new gem in the morning.

PR #185, opened 2015-05-04
Add support for token revocation

nickhoffman:
## Problem

We don't provide an easy way to revoke an OAuth token.

## Solution

OAuth tokens can be revoked with `ShopifyAPI::OAuth.revoke` .

/related Shopify/facebook-commerce#2765
/review @awd, @ilikeorangutans


awd:
:sheep:

PR #187, opened 2015-05-04
Add a myshopify_domain class variable to use for building the site url.

dylanahsmith:
@clayton-shopify & @kevinhughes27 

This is an extra precaution to make sure requests always go to a .myshopify.com domain.

For development, there is still a ShopifyAPI::Session.myshopify_domain class variable that can be modified outside of the production environment.

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -6,10 +6,9 @@ module ShopifyAPI
   end
 
   class Session
-    cattr_accessor :api_key
-    cattr_accessor :secret
-    cattr_accessor :protocol
+    cattr_accessor :api_key, :secret, :protocol, :myshopify_domain, :port
     self.protocol = 'https'
+    self.myshopify_domain = 'myshopify.com'
 
     attr_accessor :url, :token, :name
 
@@ -21,12 +20,9 @@ module ShopifyAPI
 
       def temp(domain, token, &block)
         session = new(domain, token)
-        begin
-          original_domain = host_with_port(ShopifyAPI::Base.site.to_s)
-        rescue URI::InvalidURIError
-        end
-        original_token   = ShopifyAPI::Base.headers['X-Shopify-Access-Token']
-        original_session = new(original_domain, original_token)
+        original_site = ShopifyAPI::Base.site.to_s
+        original_token = ShopifyAPI::Base.headers['X-Shopify-Access-Token']
+        original_session = new(original_site, original_token)
 
         begin
           ShopifyAPI::Base.activate_session(session)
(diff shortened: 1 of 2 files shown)


fw42:
:+1:

lib/shopify_api/session.rb
```diff
@@ -6,10 +6,9 @@ class ValidationException < StandardError
   end
 
   class Session
-    cattr_accessor :api_key
-    cattr_accessor :secret
-    cattr_accessor :protocol
+    cattr_accessor :api_key, :secret, :protocol, :myshopify_domain, :port
     self.protocol = 'https'
+    self.myshopify_domain = 'myshopify.com'
```
kevinhughes27: I guess we can override this if we are using the api client against a local shopify. I added something similar to shopify_app recently and I think we could just patch this in in actual apps for development.

dylanahsmith: I think shopify_app could just use this, since it is already dependant on the shopify_api gem.

kevinhughes27:
lgtm

test/session_test.rb
```diff
@@ -18,6 +18,14 @@ class SessionTest < Test::Unit::TestCase
       assert session.valid?
     end
 
+    should "ignore everything but the subdomain in the shop" do
+      assert_equal "https://testshop.myshopify.com/admin", ShopifyAPI::Session.new("http://user:pass@testshop.notshopify.net/path", "any-token").site
+    end
+
+    should "ignore append the myshopify domain if not given" do
```
clayton-shopify: Typo. Delete "ignore".

clayton-shopify:
Apart from my nitpick, LGTM. :+1:

test/session_test.rb
dylanahsmith: removed "ignore"

vitalikdanchenko:
Why do you treat custom domains like `mysuperdomain.com` equivalent to  `mysuperdomain.myshopify.com`? You can't just extract subdomain and append to it `myshopify.com`. That is insane! For example `http://example.com` site isn't `http://example.myshopify.com` shopify shop. 

@dylanahsmith @clayton-shopify, please explain

dylanahsmith:
You don't want to send the client secret to anything but a myshopify.com domain.

vitalikdanchenko:
omg, why do you do that? but I think I don't want to know this.
Okay, I understood. Thank you.

dylanahsmith:
just don't set the myshopify_domain class variable and you have nothing to worry about.  It is just there for shopify employees to test with a dev instance of shopify

PR #188, opened 2015-05-12
Escape ampersand in params or equal in keys before signing with hmac.

dylanahsmith:
@csaunders & @clayton-shopify for review

This implements the same escaping that we added in Shopify for generating the signature to prevent delimiters from being added to tamper with params without changing the signature.  In the uncommon cases where these delimiters do appear in the keys or values, this pull request will make sure the signature is properly validated.

### DIFF
--- a/test/session_test.rb
+++ b/test/session_test.rb
@@ -171,6 +171,15 @@ class SessionTest < Test::Unit::TestCase
       assert_equal true, ShopifyAPI::Session.validate_signature(params)
     end
 
+    should "return true when validating signature of params with ampersand and equal sign characters" do
+      ShopifyAPI::Session.secret = 'secret'
+      params = {'a' => '1&b=2', 'c=3&d' => '4'}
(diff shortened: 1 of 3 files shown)


csaunders:
OK

test/session_test.rb
```diff
@@ -171,6 +171,15 @@ class SessionTest < Test::Unit::TestCase
       assert_equal true, ShopifyAPI::Session.validate_signature(params)
     end
 
+    should "return true when validating signature of params with ampersand and equal sign characters" do
+      ShopifyAPI::Session.secret = 'secret'
+      params = {'a' => '1&b=2', 'c=3&d' => '4'}
```
clayton-shopify: Should we throw some `%` signs in for good measure?

dylanahsmith: Added a test for escaping `%`

clayton-shopify:
:+1:

PR #190, opened 2015-05-26
brute force product attribute search - until api supports this

achadee:
A brute force product/variant search to search for useful things such as SKU. This doesn't work through the API, so the gem pulls down all products and searches through them.


lib/shopify_api/resources/product.rb
```diff
@@ -29,5 +29,42 @@ def add_to_collection(collection)
     def remove_from_collection(collection)
       collection.remove_product(self)
     end
+
+    def self.search(params)
```
csaunders: This will be kinda slow and I think you'll be limited to a single page of products. The function is also super massive and doesn't feel right.

I think we can perhaps take a better approach to this or simply make the API respond to more useful queries. If we can't take the query approach then let's fix up this function to be a bit easier to read, faster, etc.

lib/shopify_api/resources/product.rb
```diff
@@ -29,5 +29,42 @@ def add_to_collection(collection)
...
+      result = []
+      products.each do |product|
+        meets_criteria = true
+        temp = product.as_json
+        params.each do |key, value|
+          if !temp[key.to_s].nil?
+            if temp[key.to_s] != value.to_s
+              meets_criteria = false
+            end
+            puts "searching products"
```
csaunders: This shouldn't be here.

lib/shopify_api/resources/product.rb
```diff
@@ -29,5 +29,42 @@ def add_to_collection(collection)
...
+          else
+            product.variants.each do |variant|
+              v_temp = variant.as_json
```
csaunders: why do we need to reserialize every variant back into JSON?

test/product_test.rb
```diff
@@ -35,4 +35,19 @@ def test_update_loaded_variant
     variant.price = "0.50"
     variant.save
   end
+
+  def test_search
+    fake "products/search.json?query=sku%3AIPOD2008PINK", extension: false, body: load_fixture('products_search')
+
+    results = ShopifyAPI::Product.search(query: "sku:IPOD2008PINK")
+    # puts results.as_json
```
csaunders: Remove unnecessary comments.

test/product_test.rb
```diff
@@ -35,4 +35,19 @@ def test_update_loaded_variant
...
+
+    results.first.variants.each do |variant|
+      puts variant.sku
```
csaunders: don't print out to console in tests

test/product_test.rb
```diff
@@ -35,4 +35,19 @@ def test_update_loaded_variant
     variant.price = "0.50"
     variant.save
   end
+
+  def test_search
```
csaunders: This test isn't very good and doesn't verify that the only products you are getting back from the query are only those of the SKU you were expecting. You just verify that at least the pink iPod was in the result set. I could return all the products and this test would pass because it would eventually find that variant.

csaunders:
This might be a useful feature but in it's current manifestation we cannot accept this PR. If you'd like a hand improving it let me know.

ilikeorangutans:
I agree, the idea is good, but there are better approaches to this. I would prefer this to go into the actual API. It means less load for the servers and it would be faster too.

achadee:
Well this was a brute force method, and a hack (as the title suggests) until you guys get this working API side. Wasn't meant to be efficient. Feel free to contribute if you know a better way of searching products via sku. I really only have the need for SKU search but I just chucked in an all attributes search incase other people have a need for that.

achadee:
also, woops, wasn't aware products were paginated ... will be a bit more thorough in my testing next time.

IslamAzab:
Is there any updates about such feature?

csaunders:
This implementation of the feature wasn't really right.

We've been doing some experimentation on some internal apps that abstracts the pagination away from the user that allows `.each` fetch the data in batches. We'd still need to see if we can leverage the API for searching and anything that isn't API supported would need to be run through a filter.

The main problem with this approach is it requires requesting _everything_ from shopify to then filter everything out on the client. If we have a limited subset of queryable items that can be filtered via the API we will probably want to use those.

Ad Hoc queries may still end up being expensive though :(

achadee:
Would it be possible to index your products table additionally by SKU and make the result only return the first it finds, This would require a lot less computation and would resolve a lot of our issues. 

We are using the SKU is a unique identifier across multiple channels so this would suffice.

Alternatively, If there is a searchable field (a tag, metadata etc. etc) where we could place the SKU this would also be useful?

PR #199, opened 2015-07-06
Add missing helper method on ApplicationCharge

gavinballard:


kevinhughes27:
looks fine to me - did recurring charge have any tests?

ping @awd for another look

awd:
@gavinballard Can you add some test coverage? `RecurringApplicationCharge` has some to get an idea.

gavinballard:
Yes, there were a couple of tests for recurring charge but nothing for the regular application charge. This is probably a good opportunity to add some, I'll take a second swing at this when I get the chance.

awd:
thanks :+1:

gavinballard:
Ping @awd and @kevinhughes27 - tests added.

lib/shopify_api/resources/application_charge.rb
```diff
@@ -2,6 +2,12 @@ module ShopifyAPI
   class ApplicationCharge < Base
     undef_method :test
 
+    class << self
+      [:pending, :cancelled, :accepted, :declined].each do |status|
+        define_method(status) { (all || []).select { |c| c.status == status.to_s } }
```
awd: This comparison between `c.status == status.to_s` is case sensitive. If the API ever returned statuses in a different case, it would break the API client in a subtle way.

Instead can you do something like:
`c.status.casecmp(status.to_s).zero?`  

Also can you please add a test to cover this as well?

awd:
I've added one comment, otherwise it looks good. I'll merge after the comparison becomes case-insensitive and a test has been added.

lib/shopify_api/resources/application_charge.rb
gavinballard: Thanks @awd.

I'm happy to do this if you really think this is necessary, but it seems to me that if the API is returning statuses in a different case than provided in the documentation, then the API is broken (or has been updated to a new major version, in which case all clients should have been notified and given time to make the appropriate changes).

I'm all for defensive programming, but I don't feel that should really extend to a client handling an API behaving outside of its spec.

My two cents - like I said, this is your call.

gavinballard:
Ping @awd re: my response to your note.

lib/shopify_api/resources/application_charge.rb
awd: This type of problem has burned us before. But I agree the API should not change the case on important strings like these.

awd:
Thanks! Will get this released today!

gavinballard:
Thanks @awd :).

Issue #198, opened 2015-07-04
Missing helper methods on ApplicationCharge


gavinballard:
The `RecurringApplicationCharge` model has convenience methods for fetching charges based on their current state (`:pending`, `:accepted` et cetera). Is there any reason these couldn't be added to `ApplicationCharge` as well? Obviously the `current`, and `cancel` methods wouldn't make sense on this model. Ping @odorcicd as he was the original author of the status methods.

If it's a useful addition, I'll PR.

kevinhughes27:
I don't see why not.

dtKinger:
@awd Any update required to https://docs.shopify.com/api/applicationcharge#properties ?

Currently, valid statuses are listed as:
* pending
* accepted
* declined
* expired

If so I can take care of them - what should the list be?

awd:
Hmm doesn't appear `cancelled` is listed as a status in the docs, but has a helper in the codes.

@gavinballard Was this intentional?

gavinballard:
Hmm, I think that's probably an error on my part. As far as I'm aware, `ApplicationCharges` can't be cancelled?

dtKinger:
@gavinballard I just looked through our application_charge.rb file in Shopify, it lists the statuses as:

pending
accepted
declined
active
expired

The 5th one, active appears to do two things, create billing contract and email the merchant.

So I'm thinking [this line](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/resources/application_charge.rb#L6) should read:

`[:active, :pending, :expired, :accepted, :declined].each do |status|`

I don't imagine the active status lasts very long, so not sure if it should be included for a convenience method.

@awd sanity check?

awd:
This makes sense to me. @gavinballard mind, setting up another PR?

gavinballard:
Thanks @dtKinger and @awd, new PR created. I took your advice and ignored the `:active` state as you're right that that's a transitory stage that it doesn't seem worth being able to query for.

PR #202, opened 2015-08-07
Refs #198: Tweak statuses for ApplicationCharge convenience method.

gavinballard:


awd:
Thanks @gavinballard will release momentarily!

Issue #138, opened 2014-07-26
Deleting a Product Image resource fails API v 3.4.2


resistorsoftware:
Given a Product with Images we can iterate product.images

Given an instance of an image eg:

```
img = ShopifyAPI::Product.find(260391085).images.first
=> #<ShopifyAPI::Image:0x007fb4b9375bc0 @attributes={"created_at"=>"2014-03-08T20:37:26-05:00", "id"=>570863025, "position"=>1, "updated_at"=>"2014-03-08T20:37:26-05:00", "src"=>"https://cdn.shopify.com/s/files/1/0084/3892/products/calice1_2013.jpg?v=1394329046"}, @prefix_options={:product_id=>260391085}, @persisted=true>
```

We can see we get the parent product ID in the product_options attribute:

```
@prefix_options={:product_id=>260391085}
```

Trying to delete this image resource fails however.

```
ShopifyAPI::Image.delete(img.id)
ActiveResource::MissingPrefixParam: product_id prefix_option is missing
```

It would be nice to factor image resource deletes into the API.

PR #205, opened 2015-08-25
[amends https://github.com/Shopify/shopify_api/issues/138]

kevinmpowell:
Document alternate syntax for deleting an image with one api call.

I added a test for deleting an image using the syntax mentioned by:
@Hammadk 

In this comment: https://github.com/Shopify/shopify_api/issues/138#issuecomment-50498732

In my use case I wanted to delete an image with one API call, not fetch the image and then delete it using a subsequent call. The gem already supports this syntax, it's just buried in this github issue: https://github.com/Shopify/shopify_api/issues/138

With it added to the tests it should be easier for people to discover in the future.

### DIFF
--- a/test/image_test.rb
+++ b/test/image_test.rb
@@ -36,4 +36,10 @@ class ImageTest < Test::Unit::TestCase
     image = ShopifyAPI::Image.new(product_id: 632910392, id: 850703190)
     assert image.destroy
   end
+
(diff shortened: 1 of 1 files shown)


Hammadk:
:+1: 

boop @kevinhughes27 for review.

kevinhughes27:
++tests lgtm

PR #208, opened 2015-09-26
Closes #162 Use public_send instead of send on Session#setup

### LINKED ISSUE #162: Session#setup should be using public_send
Using send is dangerous/sketchy.

See #125 for an idea of how to solve this.

cefigueiredo:


kevinhughes27:
@csaunders

PR #218, opened 2015-10-26
Create CONTRIBUTING.md

kevinhughes27:
I'd like to add this file to help direct more questions to the forums rather than on Github where less people are watching.

This is what it will look like when people are submitting issues after this:
https://github.com/blog/1184-contributing-guidelines

Thoughts? I'd like to do the same for the python lib as well.
@ShayneP @joshubrown @atleeclark


ShayneP:
Love the idea, would just change the copy of the last point to something like : "When in doubt, post on the forum first. You'll likely have your questions answered more quickly if you post there; more people monitor the forum than Github."

kevinhughes27:
yeah that suggestion sounds great to me

kevinhughes27:
updated. @joshubrown can you look plz

PR #220, opened 2015-11-04
String errors should be added to :base

shawnfrench:
The fix in https://github.com/Shopify/shopify_api/pull/107 was wrong IMO. As the documentation states, `attribute should be set to :base if the error is not directly associated with a single attribute.`

@kevinhughes27 @maartenvg

### DIFF
--- a/lib/active_resource/json_errors.rb
+++ b/lib/active_resource/json_errors.rb
@@ -25,7 +25,7 @@ module ActiveResource
     def from_string(error, save_cache = false)
       clear unless save_cache
 
-      add("message", error)
+      add(:base, error)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


kevinhughes27:
Makes sense - how did this come up?

shawnfrench:
I was trying to output errors coming back from Shopify (eg. using `my_model.errors.full_messages.to_sentence`) and without this fix "Message " is pre-pended to the error.

Before: "Message Something went wrong"
After: "Something went wrong"

kevinhughes27:
:cool:

maartenvg:
:+1:

Seems like we don't have a test for it, since changing it didn't break anything. Can you add one?

shawnfrench:
Whoops ya I meant to fix those. Done.

shawnfrench:
Good to :ship: ?

maartenvg:
:boat:

kevinhughes27:
:+1:

PR #222, opened 2015-11-06
Add shipping zone resource

RichardBlair:
### Problem

We moved everyone over to shipping zones, and it has made all the rates that are showing on the countries resource invalid.

### Solution

I have a PR on shopify's side to add the shipping zone endpoint, this adds the resource.

@shawnfrench @csaunders


RichardBlair:
Can I get a review on this? :smile:

shawnfrench:
:ship:

PR #228, opened 2015-11-30
Add default timeout to ShopifyAPI::Base

### LINKED ISSUE #225: Timeout setting for ActiveResource
Is it possible to make use of the `timeout` setting available via `ActiveResource` ?

Adding `self.timeout = 5` (or similar value) at [lib/shopify_api/resources/base.rb](https://github.com/Shopify/shopify_api/blob/21338ab06bbecc93cb1352d465df248e15ff9c25/lib/shopify_api/resources/base.rb#L7) should help.

I understand that ActiveResource is internally based on `Net::HTTP`, which should have a `read_timeout` of 60 seconds. However, this change will help to fail-fast rather than have an `ActiveResource` call wait for a long time.

alexcoco:
@maartenvg @celsodantas @peterjm 

Closes #225

There is not default timeout on `ShopifyAPI::Base` (it's `nil` 😔). This introduces a ~~25~~ 90 second default timeout for API requests.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+== Version 4.1.1
+
+* Added explicit 90 second timeout to `ShopifyAPI::Base`
+
 == Version 4.0.7
 
 * Added `ShippingAPI::ShippingZone`
--- a/lib/shopify_api/resources/base.rb
+++ b/lib/shopify_api/resources/base.rb
@@ -4,6 +4,7 @@ module ShopifyAPI
   class Base < ActiveResource::Base
     class InvalidSessionError < StandardError; end
     extend Countable
+    self.timeout = 90
     self.include_root_in_json = false
     self.headers['User-Agent'] = ["ShopifyAPI/#{ShopifyAPI::VERSION}",
                                   "ActiveResource/#{ActiveResource::VERSION::STRING}",
--- a/lib/shopify_api/version.rb
+++ b/lib/shopify_api/version.rb
@@ -1,3 +1,3 @@
 module ShopifyAPI
-  VERSION = "4.1.0"
+  VERSION = "4.1.1"
 end


peterjm:
25 seconds is too short for some Shopify API requests, although I guess it's configurable per app to make it longer.

There's also going to be a patch that allows separate open and read timeouts for ActiveResource (https://github.com/Shopify/shopify/pull/60047), but relying on that would assume that the project is using the most up-to-date ActiveResource.

alexcoco:
If it's too short then we can increase it but leaving it blank is not a great idea.

peterjm:
Agreed that we shouldn't leave it blank, and I don't think there's going to be a single great timeout (I don't think we'd pick a timeout greater than a minute, and some API endpoints consistently take longer than a minute). So 25 seconds seems reasonable to me

celsodantas:
:+1: 
I agree that leaving it blank is not great.

maartenvg:
Is it maybe better to default to be slightly under Shopify's normal timeout? It's really up to the app to define there best timeout for their queries, and I fear users of this gem will run into unexplained timeouts breaking their app when they upgrade. Other option is documenting the change well and making it a breaking version bump to 4.1.x.

When we have open & read, it's different of course.

celsodantas:
Can we update the minor version to alert ppl that the gem update might break stuff?

alexcoco:
In that case then we should set the timeout to whatever the timeout would have been from Shopify anyway. That way it's not a breaking change but it still does help because if the network is just messed up and Shopify doesn't timeout the request, then the API client will timeout, preventing requests that would otherwise go on way longer than they should.

alexcoco:
@maartenvg @celsodantas @peterjm I've updated the timeout to be the same as in Shopify so that this will not change the behaviour (and, therefore, not be a breaking change). The upside to this change is that in the case that the timeout is not respected on the Shopify side, the client will instead fail with a timeout instead of being stuck for way longer than 90s.

celsodantas:
I'm cool with that. :+1: 

We would still need to bump the version to release a new version of the gem.

peterjm:
^ agreed

celsodantas:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shopify/shopify_api/228)*

alexcoco:
Version bumped. Any objections to merging?

celsodantas:
toasty-left

maartenvg:
@celsodantas ..... ?

maartenvg:
:+1:

celsodantas:
hehe that's a :+1:

alexcoco:
Rebased to fix version number conflict. Merging on :green_apple:

PR #240, opened 2016-01-26
[Refactoring] Use inheritance instead of Module#alias_method_chain

DanielVartanov:
Reasons:
1. `alias_method_chain` is obsolete.
2. It's easier to manage custom functionality by prepending modules.


DanielVartanov:
Travis made a good point, this will work with Ruby 2.0+. Is it acceptable? If yes, I will add this constraint to gemspec.

kevinhughes27:
I think that is probably okay. @ShayneP do you guys have any idea about the usage of ruby 1.9 amongst  apps?

swalkinshaw:
I was trying to track down a deprecation warning of `alias_method_chain` which led me here. Any movement on this?

DanielVartanov:
@swalkinshaw I am waiting for approval to drop support for Ruby <2.0. As soon as it is approved, this PR can be merged.

kevinhughes27:
Can you update travis to run with a bunch of 2+ rubies and remove the 1.9.3 rubies so we can get a green test run. 

@christianblais @dominiquesr are you both cool with bumping the required ruby for the latest api gem to 2.0 + and up?

maartenvg:
I'll do a quick check of the logs to see what our current distribution of ruby versions is.

kevinhughes27:
Thanks @maartenvg

christianblais:
> @christianblais @dominiquesr are you both cool with bumping the required ruby for the latest api gem to 2.0 + and up?

Yes, let's bump. And while we're bumping, I'd bump to 2.3. Older ruby can still lock their bundle to an previous version of the gem.

machty:
@DanielVartanov What's the status on this?

kevinhughes27:
This needs more work to force the test suite to run on the right ruby version and probably some readme notes as well.

swalkinshaw:
@DanielVartanov friendly reminder about this :)

Seems like the consensus is to only test 2.3 now. Here's what the `.travis.yml` file should look like:

```yml
language: ruby
sudo: false

rvm:
  - 2.3.1

gemfile:
  - Gemfile
  - Gemfile_ar30
  - Gemfile_ar31
  - Gemfile_ar32
  - Gemfile_ar40threadsafe
  - Gemfile_ar42threadsafe
  - Gemfile_ar50

before_script: echo $OLD_RAKE

matrix:
  include:
    - rvm: 2.3.1
      gemfile: Gemfile_ar30
      env: OLD_RAKE=1
    - rvm: 2.3.1
      gemfile: Gemfile_ar31
      env: OLD_RAKE=1
    - rvm: 2.3.1
      gemfile: Gemfile_ar32
      env: OLD_RAKE=1
```

@kevinhughes27 does that look correct? Not sure if the `OLD_RAKE` stuff is still needed.

And here's some `README` copy to address this change. Maybe add it under "Requirements"?

> Ruby version
This gem requires Ruby 2.3.1 as of version 4.3. If you need to use an older Ruby version then update your `Gemfile` to lock onto an older release than 4.3.

DanielVartanov:
Hey @swalkinshaw,
By coincidence, I was working on it today. I manually tested all configurations.  Here is what I have found:

* Ruby 2.1.5 is cumbersome to support because rack 2.x.x requires Ruby > 2.2.
* Ruby 2.2.5 and 2.3.1 are equally easy to support (if you say just keep 2.3.1 -- ok, I'll leave only it)
* It seems to make no sense to support `OLD_RAKE` (rake 0.94.x), can I get rid of it?
* `Gemfile_ar42threadsafe` no longer makes sense since AResource 4.1 has merged support for threadsafety. Can I replace it with `Gemfile_ar41`?
* Additional gem dependencies in `Gemfile_ar50` don't seem to make any difference, everything works without them. They were added to match gem depencencies of AResource, but it no longer seem to be required. Is it okay to remove them?
* What I did not figure out is what gem dependency of `minitest` is doing in each Gemfile. What is the purpose of it?

swalkinshaw:
* If we're going >= 2.3 then we should also update the gemspec to enforce that `rack` version. Right now there's no version requirements (this actually causes any new PR to fail right now since rack resolves to 2.0.1).

* If we're going to require 2.3 then it also seems like a good time to drop that `OLD_RAKE` support.

* re: `Gemfile_ar42threadsafe` looks like you're correct. @peterjm can you confirm this is no longer needed?

* `Gemfile_ar50`: do you mean the entire file? Or just [these extra 4](https://github.com/Shopify/shopify_api/blob/cc931e7f5414a42ea9a63101c0ecc58aad9a1024/Gemfile_ar50#L8-L11)?

* It looks like adding `minitest` was fixing some unknown TravisCI issue. See [this](https://github.com/Shopify/shopify_api/commit/3548341237deed2b9f2c33f032a11f0a3c9a7a66) for example. But I just put up a [branch](https://github.com/Shopify/shopify_api/compare/minitest-gemfile-fix?expand=1) that removed it all tests [passed](https://travis-ci.org/Shopify/shopify_api/builds/142308970).

@DanielVartanov I say go ahead with all the changes (2.3 only, remove `Gemfile_ar42threadsafe`, remove extra `minitest`, and `OLD_RAKE`).

One minor question is if we should test both `2.3.0` and `2.3.1`?  Not sure that matters.

`2.4.0-preview1` should be added as well and then set it under `allow_failures`.

Let's get the PR green and then hopefully it's all good :)

DanielVartanov:
* Thanks
* Thanks
* Thanks
* Yes, those extra 4, everything works without them
* Understood, thanks

It's deep night here, I'll do all this stuff in the morning, if you don't mind, see you soon in a green PR!

DanielVartanov:
@swalkinshaw looks like all done and it's green, please review

peterjm:
I'm not sure that `Gemfile_ar42threadsafe` should be removed, although it should probably be renamed to avoid confusion now that AR 4.1 is out.

Up until 4.0, ActiveResource version numbers renamed in lockstep with Rails version numbers, so ActiveResource 4.0 was compatible with Rails 4.0, etc.

For Rails 4.1 and 4.2, there weren't any core ActiveResource releases, but Shopify's fork was updated to be compatible with Rails 4.2 (which is what `Gemfile_ar42threadsafe` represents; we didn't make a version for 4.1).

Is ActiveResource 4.1 compatible with Rails 4.2? If so, we should drop the `Gemfile_ar42threadsafe` file, and tell people to stop using Shopify's forked version. If it's not compatible, we should rename the file to something horrible, like `Gemfile_ar40_threadsafe_rails42`

Either way, `Gemfile_ar50` should be renamed to `Gemfile_ar41`, since the official new ActiveResource version number is 4.1 and not 5.0 (which is what we expected it would be to match Rails).

swalkinshaw:
Thanks @peterjm.

> Is ActiveResource 4.1 compatible with Rails 4.2? If so, we should drop the Gemfile_ar42threadsafe file, and tell people to stop using Shopify's forked version. If it's not compatible, we should rename the file to something horrible, like Gemfile_ar40_threadsafe_rails42

According to its [gemspec](https://github.com/rails/activeresource/blob/c1f1e8ba1ed7a1b58440016d2f3cc4ef76d7aec9/activeresource.gemspec#L24-L25), it's Rails 4.2 compatible (and anything `~> 4.0`)

So we shouldn't have any Gemfiles that contain the Shopify fork anymore right?

I think this is the summary so far:

* Drop `Gemfile_ar40threadsafe`
* Drop `Gemfile_ar42threadsafe`
* Keep `Gemfile_ar41` (@DanielVartanov already added it)

Should we bump the main `Gemfile` from `~> 4.0.0` to `~> 4.1`? Not even sure `Gemfile_ar41` is needed if we do that. `Gemfile_ar50` can remain to track `master`.

Final list would look like:

```yml
  - Gemfile # `~> 4.1`
  - Gemfile_ar30
  - Gemfile_ar31
  - Gemfile_ar32
  - Gemfile_ar_master
```

peterjm:
> Should we bump the main Gemfile from ~> 4.0.0 to ~> 4.1?

Yes, let's do that.

> Not even sure Gemfile_ar41 is needed if we do that.

I think you're right, although then we should have a `Gemfile_ar40` to make sure we maintain compatibility with ActiveResources 4.0

> Gemfile_ar50 can remain to track master.

I'm okay with that, but we should configure it to not make builds fail. We don't want CI to fail if ActiveResource does something not backwards compatible.

swalkinshaw:
@DanielVartanov okay new list :)

```yaml
  - Gemfile # `~> 4.1`
  - Gemfile_ar30
  - Gemfile_ar31
  - Gemfile_ar32
  - Gemfile_ar40 # this is new. locked to `4.0.0`
  - Gemfile_ar_master # allow to fail
```

Thanks for your patience as we figure this out.

DanielVartanov:
@peterjm @swalkinshaw thank you guys for figuring this out. Versioning of ActiveResource is so confusing that I only now realised how wrong `_ar50` was.

.travis.yml
```diff
@@ -1,31 +1,33 @@
 language: ruby
+sudo: false
 
 rvm:
-  - 2.0.0
-  - 1.9.3
+  - 2.3.1
```
swalkinshaw: I think `2.4.0-preview1` needs to be here? There were no 2.4 tests run: https://travis-ci.org/Shopify/shopify_api/builds/142531600

swalkinshaw:
@DanielVartanov awesome, it looks great 👍  noticed one more minor thing then it should all be good.

@kevinhughes27 want a final look on this?

.travis.yml
```diff
@@ -1,31 +1,33 @@
...
   - Gemfile_ar30
   - Gemfile_ar31
   - Gemfile_ar32
-  - Gemfile_ar40threadsafe
-  - Gemfile_ar42threadsafe
-
-sudo: false
-
-before_script: echo $OLD_RAKE
+  - Gemfile_ar40
```
swalkinshaw: And add `Gemfile_ar_master` add to this list

.travis.yml
```diff
@@ -1,31 +1,33 @@
...
-    - rvm: 1.9.2
+    - rvm: 2.4.0-preview1
       gemfile: Gemfile_ar32
-      env: OLD_RAKE=1
+    - rvm: 2.4.0-preview1
+      gemfile: Gemfile_ar40threadsafe
+    - rvm: 2.4.0-preview1
+      gemfile: Gemfile_ar41
+    - rvm: 2.4.0-preview1
+      gemfile: Gemfile_ar50
```
swalkinshaw: This can be removed now (`Gemfile_ar50`)

.travis.yml
```diff
@@ -1,31 +1,33 @@
...
-    - rvm: 1.9.2
+    - rvm: 2.4.0-preview1
       gemfile: Gemfile_ar31
-      env: OLD_RAKE=1
-    - rvm: 1.9.2
+    - rvm: 2.4.0-preview1
       gemfile: Gemfile_ar32
-      env: OLD_RAKE=1
+    - rvm: 2.4.0-preview1
+      gemfile: Gemfile_ar40threadsafe
```
swalkinshaw: `Gemfile_ar40threadsafe` entry as well

.travis.yml
DanielVartanov: Sorry, I have very little experience with Travis. `Gemfile_ar_master` is supposed to be allowed to fail. Isn't adding it here makes Travis to test it without allowing to fail?

.travis.yml
DanielVartanov: Very right, I missed it, thanks for spotting.

.travis.yml
DanielVartanov: (the same question as about `Gemfile_ar_master`

.travis.yml
swalkinshaw: I believe it needs to be specified here and then the matrix acts as an override. The matrix is implicitly generated from the combination of Ruby versions, Gemfiles, etc.

DanielVartanov: Thanks, I'll try, let's see how it goes

DanielVartanov:
@swalkinshaw, great, it worked as expected, thanks again for your help. Please do a final review, did we miss anything this time?

kevinhughes27:
I'm happy with this. Maybe squash the travis related commits into 1 but otherwise 🚢

DanielVartanov:
Ok, 2 secs

.travis.yml
```diff
@@ -1,31 +1,31 @@
...
-  - Gemfile_ar40threadsafe
-  - Gemfile_ar42threadsafe
-
-sudo: false
-
-before_script: echo $OLD_RAKE
+  - Gemfile_ar40
+  - Gemfile_ar_master
 
 matrix:
```
swalkinshaw: one minor bonus. can you set `fast_finish: true` here?

```yml
matrix:
  fast_finish: true
  allow_failures:
    # etc
```

swalkinshaw:
We can ignore the other CI check that's going on btw. I pushed a test branch based on this one and GitHub is showing the status here because they had a commit in common.

This one is green 👍

peterjm:
@DanielVartanov thanks so much for sticking with this. We honestly didn't mean for this to turn into an "upgrade to Rails 5" PR, but we really appreciate it.

swalkinshaw:
@DanielVartanov thanks so much for all the work on this 👏

DanielVartanov:
Yep, that escalated too quickly, but I was happy to do it, thank you guys for helping me learning new things!

ch33sybr3ad:
Hey guys, this pr is breaking shopify_api related tests in core. Every request returns a `SystemStackError: stack level too deep` If I remove `ResponseCapture` and `RequestNotification` the tests are green again.

PR #241, opened 2016-01-29
Adds friendlier error message for missing site URI

tyler-boyd:
It came up on Slack #developers-talk and on [stack overflow](http://stackoverflow.com/questions/35091476/shopify-gem-create-products-from-database-script) that it is very common to forget to be in a Shopify session when using the Shopify API, and the error message isn't great, just `"Missing site URI"`.

It would help everyone if we make the error message better (the copy can change; IIRC in the context of this gem all that matters is `ShopifyAPI::Base.site` being set, whereas in the `shopify_app` gem you generally would use a `shop.with_shopify_session {}` wrapper)

### DIFF
--- a/lib/active_resource/connection_ext.rb
+++ b/lib/active_resource/connection_ext.rb
@@ -5,6 +5,15 @@ module ActiveResource
 
     attr_reader :response
 
+    alias_method :raw_initialize, :initialize
+
+    def initialize(site, *args)
+      if site.blank?
+        raise ArgumentError, 'Missing site URI. Make sure ShopifyAPI calls are wrapped in a Shopify session'
+      end
+      raw_initialize(site, *args)
+    end
+
     def handle_response_with_response_capture(response)
       @response = handle_response_without_response_capture(response)
     end
--- a/test/active_resource/connection_test.rb
+++ b/test/active_resource/connection_test.rb
@@ -1,0 +1,17 @@
+require 'test_helper'
+
+class ConnectionTest < Test::Unit::TestCase
+  test 'initialize with a nil site should raise an exception' do
+    error = assert_raises ArgumentError do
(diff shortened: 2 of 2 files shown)


Mavvie:
@benjlcox @ShayneP could you guys review this if you have time? :smile:

lib/active_resource/connection_ext.rb
```diff
@@ -5,6 +5,15 @@ class Connection
 
     attr_reader :response
 
+    alias_method :raw_initialize, :initialize
```
benjlcox: why the alias?

Mavvie: thought that was best practice for modifying a method without overriding it; I don't think `super` would work here.

I could copy the actual `initialize` body, but I think it's better to alias the old one and call the alias.

peterjm: Why wouldn't `super` work?

Mavvie: Yeah. What do you think, should we merge this? A lot of people wanted it when it was discussed on Slack #developers.

kevinhughes27: I am in favour. One other comment - I hit this error a lot in tests so maybe we should make a note about setting the shopify session for your test suite. You might even be able to make this extra message only show during tests

lulessa:
It would also be good if there was a query method telling whether a ShopifyAPI session is active, e.g., `ShopifyAPI::Base.active_session?`

jianghong:
Was just thinking about putting in a PR for this issue. @mavvie any chance to progress on this PR?

Mavvie:
Hey, once I'm back to a stable internet connection I'll take a look at rebasing this. 

It should just be a matter of getting it reviewed -- the code changes _should_ still be okay.

PR #242, opened 2016-02-10
bump gem to 4.1.0

ch33sybr3ad:
## Description
- bump gem to 4.1.0, can't deploy without changing version number

@Shopify/finance @dominiquesr


sluggoman:
:shipit:

Issue #194, opened 2015-06-17
Detect threadsafe ActiveResource


peterjm:
@csaunders 

We're using the ["threadsafe" PRE value to detect the threadsafe version of ActiveResource](https://github.com/Shopify/shopify_api/blob/7af53e19a24d03e4c4486df7bf6ed99359a9f92d/lib/shopify_api/resources/base.rb#L30)

We'll need to update this to detect a new version of ActiveResource, after https://github.com/rails/activeresource/pull/167 ships

gavinballard:
@peterjm Does this mean we can't flip projects using `shopify_api` over to using the `master` branch of `rails/activeresource` instead of using the Shopify fork tagged at `4.2-threadsafe`? We'll be bumping up to Rails 5 on a number of Shopify apps once it ships and it would be good to drop the dependency on the Shopify repo.

peterjm:
@gavinballard I think the future of the threadsafety functionality in ActiveResource might by in doubt, given some issues like https://github.com/rails/activeresource/issues/178. @methodmissing has made some attempts to fix it in https://github.com/rails/activeresource/pull/202, though.

We can probably change `shopify_api` to stop using a "threadsafe" value for `PRE` and instead just check to see if the object responds to something like `threadsafe_attribute` -- checking against version strings is always problematic.

gavinballard:
Thanks for the quick reply @peterjm!

I guess I'll keep an eye on Active Resource and this repository and see how things pan out. For now I'll stick to using the `4.2-threadsafe` branch.

onpaws:
Hey @peterjm and friends!
Rails dev here. Have a client interested in using this gem on a greenfield Rails 5 app.
New to this issue and just trying to get up to speed on the latest and if I might be able to help.

Piecing together the story across various GH issues, it seems like ActiveResource historically doesn't play nice with threading.
I see that rails/activeresource#167 got merged on the upstream side.
Then over on [better_detection_of_threadsafe_activeresource](/Shopify/shopify_api/tree/better_detection_of_threadsafe_activeresource) I see we're now doing feature detection vs version number checking (awesome!).

What's left at this point? (checked the wiki -- apologies if this is answered somewhere already.)
Is there a Slack/etc I should be asking these questions on?

peterjm:
@onpaws I think all that was necessary was to give me a reminder to finally ship that change. Hopefully you can use the 4.2.0 version of the shopify_api gem, and you'll be good to go with Rails 5

[...]

PR #244, opened 2016-02-16
Better detection of threadsafe ActiveResource

peterjm:
@alexaitken @kevinhughes27 @ch33sybr3ad

Fixes https://github.com/Shopify/shopify_api/issues/194

## Problem

`ShopifyAPI::Base.headers` behaves differently, depending on whether we've detected that we're using the threadsafe version of ActiveResource, or the non-threadsafe version of ActiveResource.

Right now, we're detecting that based on the version number of the `activeresource` gem. This isn't a sustainable solution, because:

1. It assumes that the "pre" string is "threadsafe", which is only defined in Shopify's fork of ActiveResource. Now that core ActiveResource contains the threadsafe changes, we should start to work with that instead of just our own fork.

2. It only works with major version 4, and core ActiveResource is now at version 5

## Solution

Instead of detecting the version string, directly check that the methods we want to use are available on `ActiveResource::Base`

## Food for Thought

This is still an unsatisfying solution. The best thing would be to not override the `headers` method at all. Is this change really necessary? If it is necessary, is this something that can be pushed to core ActiveResource?

Compared to the [`ActiveResource::Base.headers`](https://github.com/rails/activeresource/blob/c3cd2b535b7c5cdc12cc6a3c1e11be6c74ffa179/lib/active_resource/base.rb#L652-L659), it seems like the behaviour in ActiveResource is to merge the class's headers with the superclass's headers. The behaviour in ShopifyAPI is to return either the superclass's headers or the class's headers.

This was originally added here: https://github.com/Shopify/shopify_api/pull/19

Looks like it was to support oauth changes

### DIFF
--- a/Gemfile_ar50
+++ b/Gemfile_ar50
@@ -1,0 +1,11 @@
+source "https://rubygems.org"
+
+gemspec
+
+gem 'activeresource', git: 'git://github.com/rails/activeresource', ref: 'c3cd2b535b7c5cdc12cc6a3c1e11be6c74ffa179'
+
+gem 'minitest', "~> 5.1"
+gem 'activesupport', github: 'rails/rails'
+gem 'activemodel', github: 'rails/rails'
+gem 'arel', github: 'rails/arel'
+gem 'rails-observers', github: 'rails/rails-observers'
--- a/test/base_test.rb
+++ b/test/base_test.rb
@@ -88,7 +88,7 @@ class BaseTest < Test::Unit::TestCase
     end
   end
 
-  if ActiveResource::VERSION::MAJOR >= 4 && ActiveResource::VERSION::PRE == "threadsafe"
+  if ActiveResource::VERSION::MAJOR > 5 || (ActiveResource::VERSION::MAJOR >= 4 && ActiveResource::VERSION::PRE == "threadsafe")
     test "#headers set in the main thread affect spawned threads" do
       ShopifyAPI::Base.headers['X-Custom'] = "the value"
       Thread.new do
--- a/test/base_test.rb
+++ b/test/base_test.rb
@@ -88,7 +88,7 @@ class BaseTest < Test::Unit::TestCase
     end
   end
 
-  if ActiveResource::VERSION::MAJOR > 5 || (ActiveResource::VERSION::MAJOR >= 4 && ActiveResource::VERSION::PRE == "threadsafe")
+  if ActiveResource::VERSION::MAJOR >= 5 || (ActiveResource::VERSION::MAJOR >= 4 && ActiveResource::VERSION::PRE == "threadsafe")
     test "#headers set in the main thread affect spawned threads" do
       ShopifyAPI::Base.headers['X-Custom'] = "the value"
       Thread.new do
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ gemfile:
   - Gemfile_ar32
   - Gemfile_ar40threadsafe
   - Gemfile_ar42threadsafe
-  - Gemfile_ar50
 
 sudo: false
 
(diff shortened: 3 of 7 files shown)


peterjm:
ping @alexaitken @kevinhughes27 @ch33sybr3ad

I added testing for Rails 5 (and also testing for Rails 4.2, which had a Gemfile but travis wasn't running it)

Gemfile_ar50
```diff
@@ -0,0 +1,11 @@
...
+
+gemspec
+
+gem 'activeresource', git: 'git://github.com/rails/activeresource', ref: 'c3cd2b535b7c5cdc12cc6a3c1e11be6c74ffa179'
+
+gem 'minitest', "~> 5.1"
+gem 'activesupport', github: 'rails/rails'
+gem 'activemodel', github: 'rails/rails'
+gem 'arel', github: 'rails/arel'
+gem 'rails-observers', github: 'rails/rails-observers'
```
kevinhughes27: why rails-observers ..?

peterjm: These are here to match the [current dependencies of `activeresource`](https://github.com/rails/activeresource/blob/5c4b4a8d71e3a1ab4d1bc4fc1fbfb8eab26938aa/Gemfile#L3-L6)

I believe once Rails 5 actually ships, ActiveResource will remove the dependencies on the github versions, and then we can, too.

peterjm: As for why an API client library would depend on Rails Observers, or even Rails... :facepalm:

kevinhughes27: :+1:

kevinhughes27:
lgtm

peterjm:
I've also added a version bump to 4.2.0 to this PR

@alexaitken @ch33sybr3ad any concerns?

test/base_test.rb
```diff
@@ -88,7 +88,7 @@ def teardown
     end
   end
 
-  if ActiveResource::VERSION::MAJOR >= 4 && ActiveResource::VERSION::PRE == "threadsafe"
+  if ActiveResource::VERSION::MAJOR > 5 || (ActiveResource::VERSION::MAJOR >= 4 && ActiveResource::VERSION::PRE == "threadsafe")
```
alexaitken: Is that suppose to be > 5 ?   I think it should be >=

peterjm: Very good :eyes:

alexaitken:
LGTM

onpaws:
Thanks for working on this!
Is there something I need to do to get Rails 5 working?

Steps to reproduce:

- Create a [new](https://devcenter.heroku.com/articles/getting-started-with-rails5#write-your-app) Rails 5.0.0.beta3 app.
- Add `shopify_api` to the Gemfile
- Run `bunde install`

Expected:
Shopify API gem to install.
Actual:
Bundle  attached file. 

Seems there might be dependency issue here.
Bundle suggesting running `bundle update

onpaws:
Thanks for your efforts on this guys!

Do users need to do something special to get this going on Rails 5?
Bundler blocks this gem when I try on a new Rail 5 app, please see below attachments.

### Steps to repro:
- Create a [new](https://devcenter.heroku.com/articles/getting-started-with-rails5#write-your-app) Rails 5.0.0.beta3 app.
- Add `gem 'shopify_api', '>= 4.2'` to the Gemfile
- Run `bundle install`

### Expected
Shopify API gem to install.

### Actual
Bundler blocks install
Please see [output](https://github.com/Shopify/shopify_api/files/229023/output-issue.txt)



Apparently activemodel, activesupport, and thor have dependency issues. 
Bundler output suggests running `bundle update` but no difference.

Hopefully I'm posting this in the right place - apologies if not.

peterjm:
ActiveResource hasn't released a new gem in a while, which means trying to use rubygems out of the box will only pick up ActiveResource 4.0, which isn't compatible.

Try adding:

```
gem "activeresource", github: "rails/activeresource"                                                                                    
gem 'rails-observers', github: 'rails/rails-observers'    
```

to your gemfile, to use the recent code in these projects. Once Rails 5 releases, I expect ActiveResource and Rails Observers will release new compatible gems.

onpaws:
Ah, of course. 
Thanks Peter! That seems to have done the trick.

PR #247, opened 2016-03-19
Add tax service files

andrewpaliga:
/cc @dtKinger

### DIFF
--- a/lib/shopify_api/resources/tax_service.rb
+++ b/lib/shopify_api/resources/tax_service.rb
@@ -1,0 +1,4 @@
+module ShopifyAPI
+  class TaxService < Base
+  end
+end
--- a/test/fixtures/tax_service.json
+++ b/test/fixtures/tax_service.json
@@ -1,0 +1,9 @@
+{
+  "tax_service": {
+    "name": "My Tax Service",
+    "url": "https://mytaxservice.com",
+    "id": 123456,
+    "default_tax_code": "TX123",
+    "active": "true"
+  }
+}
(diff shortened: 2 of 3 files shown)


andrewpaliga:
@aprofeit @ch33sybr3ad any objections to merging this?

@dominiquesr anyone else I should ping for review?

ch33sybr3ad:
Wouldn't you also need an `activate` method for the tax_services?

andrewpaliga:
@ch33sybr3ad I don't think that's needed for json requests.

aprofeit:
> Wouldn't you also need an activate method for the tax_services?

@Mavvie would probably have a definitive answer, but the controller doesn't respond to `json` for that action, so I would assume it was never intended to be used via API.

ch33sybr3ad:
ahh, true. I just did a quick look at the controller. other than that LGTM :+1:

Mavvie:
@aprofeit is correct for now, but I'm working on some changes that add API access to the `activate` endpoint.

I would add it to this PR, even if you have to wait for me to merge mine.

aprofeit:
@Mavvie I don't think we should add code to the gem that we know won't work yet.

Also, why do you need a custom action for `activate`? Why not just use `update` with `active: false` in the params?

andrewpaliga:
Okay will merge this for now, once it's available in Shopify we can change.

Mavvie:
Yea I meant we'd hold off on the PR until it's merged.

But merging now and changing it later works too :+1: 

> Also, why do you need a custom action for activate? Why not just use update with active: true|false in the params?

We want `activate` to be separate from other things you can update. This way people can't accidentally activate/deactivate a tax service when they try to update, say, the name.

PR #248, opened 2016-03-24
Add Refund files

julionc:
/cc @andrewpaliga @ch33sybr3ad

### DIFF
--- a/test/fixtures/refund.json
+++ b/test/fixtures/refund.json
@@ -1,0 +1,112 @@
+{
+  "refund": {
+    "id": 509562969,
+    "order_id": 450789469,
+    "created_at": "2016-03-22T11:15:00-04:00",
+    "note": "it broke during shipping",
+    "restock": true,
+    "user_id": 799407056,
+    "refund_line_items": [
+      {
+        "id": 104689539,
+        "quantity": 1,
+        "line_item_id": 703073504,
+        "line_item": {
+          "id": 703073504,
+          "variant_id": 457924702,
+          "title": "IPod Nano - 8gb",
+          "quantity": 1,
+          "price": "199.00",
+          "grams": 200,
+          "sku": "IPOD2008BLACK",
+          "variant_title": "black",
+          "vendor": null,
+          "fulfillment_service": "manual",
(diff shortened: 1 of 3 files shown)


andrewpaliga:
I actually don't know enough about this to review. @dominiquesr?

test/refund_test.rb
```diff
@@ -0,0 +1,22 @@
...
+      :refund_line_items => [{ :line_item => 518995019, :quantity => 1 }]
+    )
+    assert_equal 703073504, refund.refund_line_items.first.line_item_id
+  end
+
+  test '#find should return a refund' do
+    fake "orders/450789469/refunds/509562969", :method => :get, :body => load_fixture('refund')
+    refund = ShopifyAPI::Refund.find(509562969, :params => {:order_id => 450789469})
+    assert_equal 509562969, refund.id
+  end
```
dominiquesr: Should there be a test for calculate as well?

test/fixtures/refund.json
```diff
@@ -0,0 +1,116 @@
...
+          "gift_card": false,
+          "name": "IPod Nano - 8gb - green",
+          "variant_inventory_management": "shopify",
+          "properties": [
+            {
+              "name": "Custom Engraving Front",
+              "value": "Happy Birthday"
+            },
+            {
+              "name": "Custom Engraving Back",
```
dominiquesr: I don't think you'd ever put Merry Christmas on the same item as Happy Birthday, but maybe I'm the only one that cares that this example doesn't make sense. ;)

julionc:
You're completely right, @dominiquesr :sweat_smile:
Maybe, someone should update the json response from documentation. (See [refund#show](https://docs.shopify.com/api/reference/refund#show) & [order#show](https://docs.shopify.com/api/reference/order#show)).
BTW, i just removed Merry Christmas property, and added calculate method.

dominiquesr:
LGTM but @kevinhughes27 has more experience in this repo.

kevinhughes27:
I'm fine with this.

julionc:
Bump

PR #249, opened 2016-03-29
Added support for Discounts (Shopify Plus)

tyler-king:
I asked for this at Unite and was told to just simply do a pull request so here it is :+1: 

Tests passed on my end with Bundler:
- Get a discount
- Get all discounts
- Delete a discount
- Enable a discount
- Disable a discount

JSON pulled from Shopify docs.


tyler-king:
Added newlines to JSON resources.

kevinhughes27:
Thanks!

PR #256, opened 2016-04-19
Add Customer#account_activation_url method

julionc:
According to the [API Announcements](https://ecommerce.shopify.com/c/api-announcements/t/new-api-to-retrieve-customer-account-activation-urls-305726) post.
I updated the gem with the new `account_activation_url` method for Customers resource.

/cc @kevinhughes27


lib/shopify_api/resources/customer.rb
```diff
@@ -1,3 +1,5 @@
+require 'active_resource'
```
kevinhughes27: is this require necessary?

julionc:
removed line

PR #260, opened 2016-04-28
allow scope to be string in ShopifyApi::Session::create_permission_url

Skaelv:
Hey!
I'm using the embedded SDK and would like to do something like : 
```ruby
session_shopify.create_permission_url(ShopifyApp.configuration.scope)
```
I could have processed `ShopifyApp.configuration.scope` to an array before passing it to create_permission_url but it seems useless as you are processing it back to a string in this method.


kevinhughes27:
Why are you creating the permission url yourself and not just letting omniauth take care of authentication for you?

PR #261, opened 2016-04-28
add api_key_params to ShopifyAPI::Session::create_permission_url

Skaelv:
Add possibility to add app_id manually in case session is initiated with a token.
I'm using the embedded SDK and I need to generate a redirection url but I get :
https://localhost.myshopify.com/admin/oauth/authorize?client_id=&scope=write_products
because I'm using (from Embedded App SDK)
```ruby
@sess = ShopifyApp::SessionRepository.retrieve(session[:shopify])
 ShopifyAPI::Base.activate_session(@sess)
```
to init my session. The session doesn't contain any api_key but just a token.

I'm not sure it is reasonable for security matters to add the possibility for developers to specify the api_key on their own when generating the redirect_url but that would solve my pb :)
Let me know

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -72,9 +72,10 @@ module ShopifyAPI
       self.token = token
     end
 
-    def create_permission_url(scope, redirect_uri = nil)
+    def create_permission_url(scope, redirect_uri = nil, api_key_param = nil)
       params = {:client_id => api_key, :scope => scope.join(',')}
       params[:redirect_uri] = redirect_uri if redirect_uri
+      params[:client_id] = api_key_param if api_key_param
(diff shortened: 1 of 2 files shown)


kevinhughes27:
same question as on #260. I highly recommend just letting omniauth take care of all of this and don't create your permission urls manually.

Skaelv:
Thanks kevin that seems to be the right way to do it. I'm still confused though. 
How can I force my users to authenticate again after I released a new version with new permissions required?

Skaelv:
Do you know how I can force my users to authenticate again when I release new version of my app with new permissions ?

kevinhughes27:
There is a lot of different ways you can do this. You could probably just change the scope and then rescue from a permission error by redirecting to auth again

PR #262, opened 2016-05-05
Add ability to delegate an access token

EiNSTeiN-:
Adds the `/admin/access_tokens/delegate` endpoint. The endpoint is documented here: https://help.shopify.com/api/guides/authentication/oauth (or will be after the documentation is deployed next).

@garethson @kevinhughes27 for review

### DIFF
--- a/test/access_token_test.rb
+++ b/test/access_token_test.rb
@@ -1,0 +1,19 @@
+require 'test_helper'
+
+class AccessTokenTest < Test::Unit::TestCase
+
+  def test_delegate_access_token
+    fake "access_tokens/delegate.json?delegate_access_scope%5B%5D=write_orders&" \
+      "delegate_access_scope%5B%5D=read_products&expires_in=",
+      method: :post,
+      status: 201,
+      body: load_fixture('access_token_delegate'),
+      extension: false
+
+    delegate_scope = ['write_orders', 'read_products']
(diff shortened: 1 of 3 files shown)


test/access_token_test.rb
```diff
@@ -0,0 +1,16 @@
+require 'test_helper'
+
+class AccessTokenTest < Test::Unit::TestCase
+
+  def test_delegate_access_token
+    fake "access_tokens/delegate.json?delegate_access_scope%5B%5D=write_orders&" \
+      "delegate_access_scope%5B%5D=read_products&expires_in=", :method => :post,
+      :status => 201, :body => load_fixture('access_token_delegate'), :extension => false
```
garethson: I'd probably put each hash key on a separate line. Also, hash rockets.

garethson:
Can you also squash the commits? LGTM!

test/access_token_test.rb
EiNSTeiN-: > hash rockets

This gem needs to be 1.9.3 compatible (at least travis runs tests on this platform.

EiNSTeiN-: Never mind, I think the new syntax was introduced in 1.9 so I believe I can use the new syntax here.

kevinhughes27: We've also talked about bumping the ruby dep here

kevinhughes27:
looks fine to me

PR #264, opened 2016-05-10
WIP: Checkout API support

ryanbalsdon:
An example of expected usage:
```
ShopifyAPI::Base.site = 'https://d6b30b8acab1950f36aad826b56646d2:d885f474951bb1e11baf01e2a0b8bab5@shop1.myshopify.io/admin'
ShopifyCS.site = 'https://cardsink.myshopify.io/'

checkout = ShopifyAPI::CheckoutAPI.create({
  email: "john.smith@example.com",
  line_items: [{
    variant_id: 3,
    quantity: 1
  }],
  billing_address: {
    first_name: "John",
    last_name: "Smith",
    address1: "126 York St.",
    city: "Ottawa",
    province_code: "ON",
    country_code: "CA",
    phone: "(123)456-7890",
    zip: "K1N5T5"
  },
  requires_shipping: false,
  requires_shipping_method: false,
  requires_shipping_address: false
})

session = ShopifyCS::Session.create({
  credit_card: {
    number: "4242424242424242",
    month: "12",
    year: "2019",
    verification_value: "123",
    first_name: "John",
    last_name: "Smith",
    brand: "visa"
  }
})

payment = checkout.pay_with(session)
```

Ping @maximebedard

### DIFF
--- a/lib/shopify_api/resources/api_checkout.rb
+++ b/lib/shopify_api/resources/api_checkout.rb
@@ -1,0 +1,32 @@
+module ShopifyAPI
+  class CheckoutAPI < Base
+    self.prefix = '/api/'
+    self.element_name = 'checkout'
+    self.timeout = 1000000
+
+    def id
+      @attributes[:token]
+    end
+
+    def pay_with(session)
+      Payment.create_with_checkout_and_session(self, session)
+    end
+  end
+
+  class Payment < Base
+    def self.create_with_checkout_and_session(checkout, session)
+      self.new.create_with_checkout_and_session(checkout, session)
+    end
+
+    def create_with_checkout_and_session(checkout, session)
+      load_attributes_from_response checkout.post(:payments, {
+        payment: {
+          source: {
+            session_id: session.id
+          }
+        }
+      })
+      self
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -5,3 +5,5 @@ gemspec
 gem "activeresource", '~> 4.0.0'
 
 gem 'minitest', "~> 4.2"
+
+gem 'byebug'
--- a/lib/shopify_api/resources/credit_card.rb
+++ b/lib/shopify_api/resources/credit_card.rb
@@ -1,0 +1,31 @@
+module ShopifyCS
+  module SessionFormat
+    extend self
+
+    def extension
+      "json"
+    end
+
+    def mime_type
+      "application/json"
+    end
+
+    def encode(hash, options = nil)
+      ActiveSupport::JSON.encode(hash, options)
+    end
+
+    def decode(json)
+      ActiveSupport::JSON.decode(json)
+    end
+  end
+
+  def self.site=(site)
+    Session.site = site
+  end
+
+  class Session < ActiveResource::Base
+    self.site = 'elb.deposit.shopifycs.com'
+    self.include_format_in_path = false
+    self.format = SessionFormat
+  end
+end


heliocola:
@ryanbalsdon do you know the status of this PR?
I am looking to implement this exact same feature!

lib/shopify_api/resources/api_checkout.rb
```diff
@@ -0,0 +1,32 @@
+module ShopifyAPI
+  class CheckoutAPI < Base
+    self.prefix = '/api/'
```
maximebedard: We could remove this and use the proper admin endpoints now.

Payments is not 100% there just yet, but we can make it point to `/admin/applications/checkouts/#{checkout_id}/payments.json` until we are fully able to move everything.

maximebedard:
@ryanbalsdon I've updated the PR to use the admin endpoints and tweaked a little how to actually process payments. Next steps:

- [ ] Add automatic polling
- [ ] Add coded errors support
- [ ] Create a separate resource for AbandonedCheckouts?

@dominiquesr @maartenvg could I get 👀 on this?

@ilikeorangutans @awd for sanity check as well, does the channel related api belong here?

ilikeorangutans:
This uses the checkout api? If the connecting api client is a channel app its sales channel will automatically get attributed.

maximebedard:
@ilikeorangutans Yep, I was just wondering if you were aware about another lib for channel specific apis?

lib/shopify_api/resources/checkout.rb
```diff
@@ -1,4 +1,31 @@
...
+    def pay(session_params, &block)
+      vault_session = VaultSession.create(session_params)
+      pay_with(vault_session, &block)
+    end
+
+    def pay_with(vault_session, &block)
+      Payment.create_from(self, vault_session).tap(&block)
+    end
+
+    class ShippingRate < Base
```
maartenvg: Shipping rate should live a level higher: it's not a `CheckoutShippingRate`/`Checkout::ShippingRate`, but simply `ShippingRate`. In the future other resources are likely to share the same shipping-rate interface.

lib/shopify_api/resources/vault_session.rb
```diff
@@ -0,0 +1,28 @@
+module ShopifyAPI
+  module VaultSessionFormat
```
maartenvg: How is this different from the norman JSON format? And this should live in its own file outside of resources.

test/checkout_test.rb
```diff
@@ -0,0 +1,87 @@
+require "test_helper"
+
+class CheckoutTest < Test::Unit::TestCase
+  def setup
+    super
+    ShopifyAPI::VaultSession.format = ShopifyAPI::VaultSessionFormat
```
maartenvg: Why is this necessary if you're already doing this in line 22 of `lib/shopify_api/resources/vault_session.rb`

test/fixtures/checkout_payment.json
```diff
@@ -0,0 +1,15 @@
+{
+  "payment": {
```
maartenvg: I thought this was going to be renamed (/cc @maximebedard )

bin/console
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env ruby
```
maartenvg: This file is unrelated to your PR, please create a separate PR for that.

test/fixtures/checkout_payment.json
maximebedard: No, this is not going to be renamed. It's going to stay like this.

lib/shopify_api/resources/vault_session.rb
maximebedard: This is to avoid the patch here: https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/json_format.rb#L8.

test/checkout_test.rb
maximebedard: Because I need to avoid this: https://github.com/Shopify/shopify_api/blob/master/test/test_helper.rb#L33-L39

lib/shopify_api/resources/checkout.rb
```diff
@@ -1,4 +1,31 @@
+require 'shopify_api/resources/shipping_rate'
+
 module ShopifyAPI
   class Checkout < Base
+    has_many :shipping_rates, class_name: ShopifyAPI::ShippingRate
```
maximebedard: @JakeCataford this seems to be unsupported by earlier versions of active resource :/

maximebedard:
I've updated the PR to remove the has_many and return to the original implementation as it's not supported in earlier version of active_resource. I didn't want to backport the feature, so I believe this make sense for now. We could also hide it behind version conditional and for people to upgrade, but that would add unnecessary complexity IMO. 

Any outstanding comments here? @ryanbalsdon @maartenvg ?

maximebedard:
🛎

lib/shopify_api/resources/vault_session.rb
```diff
@@ -0,0 +1,29 @@
+module ShopifyAPI
+  # avoid removing root_key as it activeresource object initialization
```
maartenvg: This sentence seems incomplete.

lib/shopify_api/resources/checkout.rb
maartenvg: This wasn't addressed or explained why not to do it.

lib/shopify_api/resources/vault_session.rb
```diff
@@ -0,0 +1,29 @@
...
+  # https://github.com/rails/activeresource/blob/master/lib/active_resource/formats/json_format.rb#L21
+  module VaultSessionFormat
```
maartenvg: I feel this should be in a generic class in a file on its own. It's used by, but not necessarily related to `VaultSession`.

lib/shopify_api/resources/vault_session.rb
maartenvg: That doesn't really address why you need the change. Why is that endpoint behaving differently then other places?

maximebedard: Because updating cardsink to include a `root` based on some criteria and/or update all the existing apps to use the new format is an enormous task.

I can extract this into a separate file, but since this was only used by this single class, I've never see any benefits to do so.

maartenvg: > Because updating cardsink to include a root based on some criteria and/or update all the existing apps to use the new format is an enormous task.

Thanks for the explanation 👍 

> I can extract this into a separate file

Please, I feel this doesn't belong in the resources folder, but with the other helper modules. If somebody ever needs it for something else, they wouldn't know to look in `vault_session.rb`. Which means it would be duplicated.

lib/shopify_api/resources/checkout.rb
```diff
@@ -1,4 +1,31 @@
...
+    def pay_with(vault_session, &block)
+      Payment.create_from(self, vault_session).tap(&block)
+    end
+
+    class Payment < Base
+      self.prefix = "/admin/checkouts/:checkout_id/"
+
+      def self.create_from(checkout, vault_session)
+        create(
+        checkout_id: checkout.id,
```
ryanbalsdon: Indenting looks a bit off here

ryanbalsdon:
> Pull request authors can't approve their own pull request

👍

maximebedard:
Should be all good with the comments, maybe another sanity check 👀  @Shopify/api

PR #277, opened 2016-06-22
Get rid of RuntimeError

vitalikdanchenko:
It's pretty unexpectable to receive a runtime error from third-party library

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -5,6 +5,8 @@ module ShopifyAPI
 
   class ValidationException < StandardError
   end
+  class SessionException < StandardError
+  end
 
   class Session
     cattr_accessor :api_key, :secret, :protocol, :myshopify_domain, :port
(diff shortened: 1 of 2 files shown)


lib/shopify_api/session.rb
```diff
@@ -91,8 +91,10 @@ def request_token(params)
 
       if response.code == "200"
         token = JSON.parse(response.body)['access_token']
+      elsif response.code == "400"
+        raise ShopifyAPI::ValidationException, "Invalid Code: Possible malicious login"
       else
-        raise RuntimeError, response.msg
+        raise ShopifyAPI::ValidationException, "Invalid Request: Unable to validate code"
```
kevinhughes27: I am okay with this change but can we rename the exception to `ShopifyAPI::SessionException` validation exception makes me think of other things.

Also can we keep the response message for the else case

vitalikdanchenko: I agree, `ValidationException` wasn't best here

PR #284, opened 2016-07-13
Adds FulfillmentRequest

kwestein:
This adds the FulfillmentRequest object and creates a new release. 

For review @kevinhughes27 @garethson

### DIFF
--- a/lib/shopify_api/resources/fulfillment_request.rb
+++ b/lib/shopify_api/resources/fulfillment_request.rb
@@ -1,0 +1,11 @@
+module ShopifyAPI
+  class FulfillmentRequest < Base
+    init_prefix :order
+
+    def order_id
+      @prefix_options[:order_id]
+    end
+
+    def cancel; load_attributes_from_response(post(:cancel, {}, only_id)); end
+  end
+end
--- a/test/fulfillment_request_test.rb
+++ b/test/fulfillment_request_test.rb
@@ -1,0 +1,29 @@
+require 'test_helper'
+
+class FulFillmentTest < Test::Unit::TestCase
+  def setup
+    fake "orders/450789469/fulfillment_requests/695890229", :method => :get, :body => load_fixture('fulfillment_request')
+  end
+
+  context "#cancel" do
+    should "be able to cancel fulfillment request" do
+      fulfillment_request = ShopifyAPI::FulfillmentRequest.find(255858046, :params => {:order_id => 450789469})
+
+      cancelled = ActiveSupport::JSON.decode(load_fixture('fulfillment_request'))
+      cancelled['failure_message'] = 'cancelled'
+      fake "orders/450789469/fulfillments/695890229/cancel", :method => :post, :body => ActiveSupport::JSON.encode(cancelled)
+
+      assert fulfillment_request.failure_message.blank?
+      assert fulfillment_request.cancel
+      assert_equal 'cancelled', fulfillment_request.failure_message
+    end
+  end
+
+  context "#find" do
+    should "be able to find fulfillment request" do
+      fulfillment_request = ShopifyAPI::FulfillmentRequest.find(255858046, :params => {:order_id => 450789469})
+      assert_equal 695890229, fulfillment_request.id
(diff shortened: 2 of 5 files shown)


test/fulfillment_request_test.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      cancelled = ActiveSupport::JSON.decode(load_fixture('fulfillment_request'))
+      cancelled['failure_message'] = 'cancelled'
+      fake "orders/450789469/fulfillments/695890229/cancel", :method => :post, :body => ActiveSupport::JSON.encode(cancelled)
+
+      assert fulfillment_request.failure_message.blank?
+      assert fulfillment_request.cancel
+      assert_equal 'cancelled', fulfillment_request.failure_message
+    end
+  end
```
kevinhughes27: I am sort of in favour of this new test syntax but for the record the rest of this project is old style

kevinhughes27:
👍

lib/shopify_api/resources/fulfillment_request.rb
```diff
@@ -0,0 +1,11 @@
+module ShopifyAPI
+  class FulfillmentRequest < Base
+    init_prefix :order
+
+    def order_id
+      @prefix_options[:order_id]
+    end
+
+    def cancel; load_attributes_from_response(post(:cancel, {}, only_id)); end
```
garethson: I'm wondering if `cancel` should maybe be `cancel_fulfillment`? Reason being that fulfillment services have a single callback url that currently defines `fetch_tracking`, `fetch_stock_levels`, soon `fulfill` (with fulfillment requests), and now this `cancel`. So I think being explicit might make sense.

garethson: Oh... I see. this is for cancelling a fulfillment *request*, not a fulfillment. Perhaps we should be more explicit about that as well, seeing as we'll actually be adding the ability to cancel the fulfillment itself. So `cancel_fulfillment_request`?

kwestein: Gotchya, so I should add a method to cancel the associated fulfillment as well?

Since the object we're operating on is a fulfillment request isn't it clear enough that ````fulfillment_request.cancel```` would cancel it (ie the fulfillment request)? And then we could add a ````fulfillment_request.cancel_fulfillment```` method

garethson: Ugh, wrong gem I'm thinking of. I"m thinking of the `ActiveFulfillment::ShopifyApi` adapter. This ... is good.

garethson:
🚢

garethson:
We'll want to dogfood this first IMO before we bump the gem. We're still pointing a commit ref in our gemfile and we'll probably be doing more work in here over the coming weeks, so I think we should just continue not publishing a new gem version until we've validated our changes. Thoughts?

lib/shopify_api/resources/fulfillment_request.rb
kwestein: So to clarify we don't want to add the ability to cancel a fulfillment from here? Does cancelling a fulfillment request cancel the associated fulfillment (in Shopify)?

kwestein:
Ya, that's probably a good call. I will remove the version changes until we're ready to go with this

lib/shopify_api/resources/fulfillment_request.rb
garethson: Cancelling a request is more of a limit action that simply marks a *request* as cancelled after the fact. This is only coming up because internally in Shopify we have a case where the fulfillment provider doesn't know initially that items can or can not be fulfilled.

Cancelling a fulfillment actually can be done already [here](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/resources/fulfillment.rb#L9).

kwestein: 👌

kwestein:
🆗 I removed the version update stuff. @kevinhughes27 am I good to 🚢 this?

kevinhughes27:
🆗

PR #286, opened 2016-07-20
Subclass ActiveResource::Connection instead of prepending

mikeyhew:
rails/activeresource#222 has been merged, so now it is possible to use our own subclass of `ActiveResource::Connection` instead of `prepend`ing extra functionality to it. This PR moves the logic in `lib/active_resource/connection_ext.rb` to the new `ShopifyAPI::Connection` class. At the moment it isn't compatible with any version of `activeresource` before the change was made, so it probably shouldn't be merged right away.


mikeyhew:
Closing in favour of #287

PR #287, opened 2016-07-20
use a custom subclass of ActiveResource::Connection if possible

mikeyhew:
This PR attempts to improve on #286 by only setting `ActiveResource::Base.connection_class` if it is available, and falls back to monkeypatching `ActiveResource::Connection` for older versions of `activeresource` that do not have that feature.

### DIFF
--- a/lib/shopify_api.rb
+++ b/lib/shopify_api.rb
@@ -4,7 +4,6 @@ require 'active_resource'
 require 'active_support/core_ext/class/attribute_accessors'
 require 'digest/md5'
 require 'base64'
-require 'active_resource/connection_ext'
 require 'active_resource/detailed_log_subscriber'
 require 'shopify_api/limits'
 require 'shopify_api/json_format'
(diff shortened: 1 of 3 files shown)


kevinhughes27:
LGTM. @peterjm you've been looking at this code recently can you have a look here too

lib/shopify_api.rb
```diff
@@ -22,3 +21,11 @@ module ShopifyAPI
 require 'shopify_api/countable'
 require 'shopify_api/resources'
 require 'shopify_api/session'
+require 'shopify_api/connection'
+
+# ActiveResource::Base.connection_class was introduced in https://github.com/rails/activeresource/pull/222
```
peterjm: @mikeyhew can you remove this comment? It's useful information for the pull request, but doesn't seem necessary in the codebase.

peterjm:
Yeah, this seems like an improvement to me.

lib/shopify_api.rb
mikeyhew: @peterjm sure thing

peterjm:
Thanks @mikeyhew!

PR #290, opened 2016-07-25
Fix for `TypeError: can't convert Hash into Integer` when calling `Product.count` with latest activeresource

mikeyhew:
The `Countable` mixin doesn't work with activeresource@master because of this pull request that was merged in January: rails/activeresource#96. This PR adds a test to expose this issue as well as a fix.

In short, the problem is that the call to `get(:count)` now no longer removes the root from the response properly, so the return value is now `{"count" => 16}` instead of `16` (assuming there are 16 of the resource). To fix it, I added some code to check if the value returned from `get(:count)` is a Hash instead of a number, and remove the root manually. I've also posted a comment on the offending PR about the new problem it caused here: https://github.com/rails/activeresource/pull/96#issuecomment-234991786

### DIFF
--- a/test/countable_test.rb
+++ b/test/countable_test.rb
@@ -1,0 +1,12 @@
+require 'test_helper'
+
+class CountableTest < Test::Unit::TestCase
+  def setup
+    fake "products/count", :body => '{"count": 16}'
+  end
+
+  def test_count_products
+    count = ShopifyAPI::Product.count
+    assert_equal 16, count
+  end
+end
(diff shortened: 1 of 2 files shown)


test/countable_test.rb
```diff
@@ -0,0 +1,14 @@
+require 'test_helper'
+
+class CountableTest < Test::Unit::TestCase
+  def setup
+    super
```
kevinhughes27: I don't think this call to super is required

kevinhughes27:
Can you squash? then we can ship

mikeyhew:
squashed

vrehm:
Hello, i've got the same error with a fresh install of rails 5 and shopify api, do i have to do some extra config ?

kevinhughes27:
looks like we haven't shipped the gem yet. For the time being use a github dependency

PR #291, opened 2016-07-26
Add the ability to attach an image to ShopifyAPI::Image from external URL

brettchalupa:
This adds a new method: `ShopifyAPI::Image#attach_image_from_url`.

The new methods allows for attaching an external image via the `src`
attribute. This is useful when adding images that are already available
on the web. It removes the need for downloading the image or opening
it in memory when it is attached with `#attach_image`.

Based on these docs: https://help.shopify.com/api/reference/product_image#create

```
Create a new product image using a source URL that will be downloaded by Shopify

POST /admin/products/#{id}/images.json
{
  "image": {
    "src": "http:\/\/example.com\/rails_logo.gif"
  }
}
```

### DIFF
--- a/lib/shopify_api/resources/image.rb
+++ b/lib/shopify_api/resources/image.rb
@@ -12,5 +12,11 @@ module ShopifyAPI
       attributes['attachment'] = Base64.encode64(data)
       attributes['filename'] = filename unless filename.nil?
     end
+
+    # Specify the +url+ of an image that Shopify will download for the
+    # product image.
+    def attach_image_from_url(url)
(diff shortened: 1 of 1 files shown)


maartenvg:
I don't think we need a separate method for this. You could just do `image.src = http://somewhere.com/cat.gif`

The reason we do have an explicit `attach_image` helper method, is because it wraps multiple actions you need to take (Base64, filename) when setting an image from a file.

brettchalupa:
That makes sense to me. Thank you for explaining!

PR #307, opened 2016-09-14
Unexpected Auth headers is so Basic

orenmazor:
@kevinhughes27 @maartenvg 

when you set `site=` in `ActiveResource::Base`, it parses out the password and username out of that and sets it internally. so when we clear the session here, we should also clear those flags, otherwise you end up with unexpected authentication headers down the line

### DIFF
--- a/lib/shopify_api/resources/base.rb
+++ b/lib/shopify_api/resources/base.rb
@@ -58,6 +58,8 @@ module ShopifyAPI
 
       def clear_session
         self.site = nil
+        self.password = nil
+        self.user = nil
         self.headers.delete('X-Shopify-Access-Token')
       end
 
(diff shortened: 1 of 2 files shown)


kevinhughes27:
Nice fine! Can you add this to the test suite as well please https://github.com/Shopify/shopify_api/blob/master/test/base_test.rb#L38

orenmazor:
Looks like none of those tests actually test for direct assignment of `ShopifyAPI::Base.site`, or am I missing it somewhere? I can add some of that.

kevinhughes27:
Yeah I don't see any either but there is a test for clear session. Adding a test that assigns directly should be straight forward

orenmazor:
@kevinhughes27 cool. added a very simple test to cover the case

PR #308, opened 2016-09-15
Relax ruby version requirement to 2.0

casperisfine:
`shopify_api` doesn't need anything more recent than `MRI 2.0`, so might as well not be too restrictive there.

Feel free to bump that requirement as needed later, but I think it's better that way for now.

@swalkinshaw @camilo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: ruby
 sudo: false
 
 rvm:
+  - '2.0'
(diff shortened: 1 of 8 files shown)


camilo:
👍 thanks!

O-kasso:
awesome! 🎉

swalkinshaw:
🚀

PR #315, opened 2016-10-19
Adding code and tests for online-mode access tokens

EiNSTeiN-:
This adds documentation on how to use this gem with a per-user access token. It exposes `expires_in`, `expires_at` and `expired?` which can be called when the `extra` hash is populated in `ShopifyAPI::Session`. The intended use for this `extra` hash is that the caller will populate it with the data that Shopify provides, to have it available through the entire lifetime of the session. A few apps serialize the `ShopifyAPI::Session` object so it is a good place to store this information.

@clayton-shopify @ilikeorangutans @Shopify/api

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,9 +67,11 @@ ShopifyAPI uses ActiveResource to communicate with the REST web service. ActiveR
 
    For a partner app you will need to supply two parameters to the Session class before you instantiate it:
 
-  ```ruby
-  ShopifyAPI::Session.setup({:api_key => API_KEY, :secret => SHARED_SECRET})
-  ```
+   ```ruby
+   ShopifyAPI::Session.setup(api_key: API_KEY, secret: SHARED_SECRET)
+   ```
+
+   Shopify maintains [`omniauth-shopify-oauth2`](https://github.com/Shopify/omniauth-shopify-oauth2) which securely wraps the OAuth flow and interactions with Shopify (steps 3 and 4 above). Using this gem is the recommended way to use OAuth authentication in your application.
 
 3. In order to access a shop's data, apps need an access token from that specific shop. This is a two-stage process. Before interacting with a shop for the first time an app should redirect the user to the following URL:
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -133,10 +136,28 @@ ShopifyAPI uses ActiveResource to communicate with the REST web service. ActiveR
    token = session.request_token(params)
    ```
 
-   This method will save the token to the session object and return it. For future sessions simply pass the token in when creating the session object:
+   This method will save the token to the session object and return it. All fields returned by Shopify, other than the access token itself, are stored in the session's `extra` attribute. For a list of all fields returned by Shopify, read [our oauth documentation](https://help.shopify.com/api/guides/authentication/oauth#confirming-installation). If you requested an access token that is associated with a specific user, you can retreive information about this user from the `extra` hash:
```
clayton-shopify: "oauth" -> "OAuth"

README.md
```diff
@@ -133,10 +136,28 @@ ShopifyAPI uses ActiveResource to communicate with the REST web service. ActiveR
...
+
+   ```ruby
+   # a list of all granted scopes
+   granted_scopes = session.extra['scope']
+   # a hash containing the user information
+   user = session.extra['associated_user']
+   # the access scopes available to this user, which may be a subset of the access scopes granted to this app.
+   active_scopes = session.extra['associated_user_scope']
+   # the time at which this token expires, this is automatically converted from 'expired_in' returned by Shopify
```
clayton-shopify: "expired_in" -> "expires_in"

Also, I'd suggest replacing the comma with a semicolon or making the second part a separate sentence.

README.md
```diff
@@ -133,10 +136,28 @@ ShopifyAPI uses ActiveResource to communicate with the REST web service. ActiveR
...
+   expires_at = session.extra['expires_at']
+   ```
+
+   For the security of your application, after retrieving an access token you must validate the following:
+   1) The list of scopes in `session.extra['scope']` is the same as you requested.
+   2) If you requested an online-mode access token, `session.extra['associated_user']` must be present.
+   Failing either of these tests means the end-user may have tampered with the query parameters during the OAuth authentication phase. You should avoid using this access token and revoke it immediately. Using [`omniauth-shopify-oauth2`](https://github.com/Shopify/omniauth-shopify-oauth2) gem these checks are done automatically for you.
```
clayton-shopify: Suggestion: "Using ..." --> "If you use the ..."

test/session_test.rb
```diff
@@ -173,18 +213,16 @@ def setup
   end
 
   test "return true when the signature is valid and the keys of params are strings" do
-    now = Time.now
-    params = {"code" => "any-code", "timestamp" => now}
-    sorted_params = make_sorted_params(params)
-    signature = OpenSSL::HMAC.hexdigest(OpenSSL::Digest::SHA256.new(), ShopifyAPI::Session.secret, sorted_params)
-    params = {"code" => "any-code", "timestamp" => now, "hmac" => signature}
```
clayton-shopify: Who needs assertions? 😄

clayton-shopify:
Some minor comments, otherwise LGTM.

EiNSTeiN-:
Thanks for the corrections

humancopy:
Is there a reason why this hasn't progressed since October? Would be happy to help move this forward, it's quite a needed feature. ;)

EiNSTeiN-:
It has not progressed because I didn't get the 2nd review. I guess we'll do without it.

EiNSTeiN-:
This is released in v4.3.5: https://rubygems.org/gems/shopify_api

PR #317, opened 2016-11-07
Add module for only sending params that were mutated

JakeCataford:
For some API's (in particular the checkout api) we only validate params if they are sent in the request. Because active resource will send all of the params it syncs from the response, we will being validating a lot of attributes that are not yet set, and given that these endpoints are designed to be form objects that are filled incrementally we need to only send the attributes that were changed locally.

I do this in this pr by providing a module that overrides `method_missing` to catch attribute setters before passing them down to the active resource magic. Then I slice the params on the primary key + any params changed since the last result. Resetting the params after the save completes.

Couple of thoughts, this might not be completely robust, ie: update_attributes, but its a push in the right direction.

/cc @louisvanier @maximebedard @maartenvg


justinstern:
related: it would be awesome to have something like this for the Orders endpoint. I've noticed that my update orders requests often fail do to a reason like "invalid/missing shipping address" even though all I want to do is update the order note. IIRC trying to perform the update with just the single attribute (e.g. order note) still failed, so I figure it's something that would need to be changed on Shopify's side.

lib/shopify_api/only_send_mutated_params.rb
```diff
@@ -0,0 +1,34 @@
...
+        before_save :slice_attributes_on_changed
+        after_save :clear_changed_attributes
+      end
+    end
+
+    def method_missing(method_symbol, *args)
+      method_name = method_symbol.to_s
+
+      if method_name =~ /(=)$/
+        changed << $`
```
louisvanier: I know this is what was used in the source for `ActiveResource` but the intent would be cleaner with a capture group, i.e. `/(.*)(=)$/`

lib/shopify_api/only_send_mutated_params.rb
```diff
@@ -17,6 +10,13 @@ def method_missing(method_symbol, *args)
       super
     end
 
+    def save
+      slice_attributes_on_changed
+      base = super
+      clear_changed_attributes
```
louisvanier: should we only clear changed attributes if the save was succesful?

JakeCataford: Ill write a test to prove it but afaik `super` here will throw an exception and the flow will stop

JakeCataford: actually a good case to handle here would be that if the transaction is unsuccessful that we restore the original attributes to their before sliced state

lib/shopify_api/only_send_mutated_params.rb
```diff
@@ -0,0 +1,34 @@
+module ShopifyAPI
+  module OnlySendMutatedParams
```
maximebedard: I would call this `DirtyParams` or something similar?

test/only_send_mutated_params_test.rb
```diff
@@ -0,0 +1,73 @@
+require "test_helper"
+
+class OnlySendMutatedParamsTest < Test::Unit::TestCase
+  class Fake < ShopifyAPI::Base
+    include ShopifyAPI::OnlySendMutatedParams
+  end
+
+  def setup
+    super
+  end
```
maximebedard: ✂️

lib/shopify_api/only_send_mutated_params.rb
JakeCataford: Hmm, I would like to differentiate between this and something that gives you access to  dirty attributes like `ActiveModel::Dirty`, how about something like `OnlySyncDirtyAttributes` which removes the language thats http request specific?

JakeCataford:
🔔  @maximebedard @louisvanier

lib/shopify_api/resources/base.rb
```diff
@@ -4,6 +4,7 @@ module ShopifyAPI
   class Base < ActiveResource::Base
     class InvalidSessionError < StandardError; end
     extend Countable
+    self.format = :json
```
maximebedard: Why is this necessary? I thought in the tests we set this to json, but in the wild it could totally be using XML here no?

lib/shopify_api/only_sync_dirty_attributes.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      if match_data = /(?<attribute_name>.+)(?:=)$/.match(method_name)
+        changed << match_data[:attribute_name]
+      end
+
+      super
+    end
+
+    def save
+      slice_attributes_on_changed
```
maximebedard: I would block this

```rb
def save(*)
  with_dirty_attributes { super }
end

def with_dirty_attributes
  @old_attributes = @attributes
  @attributes.slice(...)
rescue # ???
  # could potentially restore the state in case something goes awfully wrong, nothing I can't think of right now, but it would be fairly easy to add.
ensure
  @changed.clear
end
```

lib/shopify_api/resources/base.rb
JakeCataford: If you are using the gem then you dont care what the return format is because AR handles the serialization and deserialization of the request/response. All of shopify's api's support JSON so we might as well only request in JSON.

The reason it fixes things is because in active resource 3.0 it seems like the default format was xml and therefore some http mocks weren't working

lib/shopify_api/only_sync_dirty_attributes.rb
JakeCataford: `ensure @changed.clear` wouldn't work because we dont want to clear our dirty state if the request did not succeed (did not throw an exception). I like the idea otherwise

JakeCataford:
ping @maximebedard you cool with this?

lib/shopify_api/only_sync_dirty_attributes.rb
```diff
@@ -0,0 +1,43 @@
...
+    private
+
+    def with_dirty_attributes
+      old_attributes = @attributes.dup
+      slice_attributes_on_changed
+      yield
+      clear_changed_attributes
+    rescue ActiveResource::ClientError,
+      ActiveResource::ConnectionError,
+      ActiveResource::TimeoutError => e
```
maximebedard: `=> e` is not necessary. Also, if you just re-raise the stack trace will be preserved properly

maximebedard:
Tiny detail on my part. Can we get more 👀  from someone else in @Shopify/api ?

lib/shopify_api/resources/base.rb
maartenvg: Since this changes the default for all apps, it might break apps that currently do anything with the response body directly. This is an actual thing that developers do, unfortunately, so this change might break apps.

Since it is just to fix some tests, I wouldn't take the risk and find some other solution. If really necessary, we'll have to treat it as a breaking change, so that would mean a major version bump.

maartenvg:
@rafaelfranca: as the maintainer of active resource, you probably have an opinion on whether this is the best way to do this. Could you spare some 👀 ?

JakeCataford:
🔔  @rafaelfranca @maartenvg?

lib/shopify_api/only_sync_dirty_attributes.rb
```diff
@@ -0,0 +1,43 @@
+module ShopifyAPI
+  module OnlySyncDirtyAttributes
```
rafaelfranca: I think it is possible to use `ActiveModel::Dirty` instead of doing a custom implementation.

maartenvg:
@JakeCataford I won't be able to look at this until tomorrow late, or next week early.

maximebedard:
@rafaelfranca @maartenvg I see a few other ways this could be done, I'd like to have your input on it :)

1- Instead of hooking into save & update_attributes, we could hook ourselves in the encode method. This way any method that tries to persist the remote record will get only the changed attributes. 

```rb
def encode(options = {})
  super(options.merge(only: changes)
end
```

2- We could use ActiveModel::Dirty, but we would still need to use `method_missing` to trigger `[attr_name]_will_change!` and let the observer know that the state as changed. We also need to hook ourselves into the module to let ActiveModel::Dirty know that the attributes are stored in `@attributes` instead of having the caller do `define_attribute_methods` for every attributes they want to watch. 


I feel like the simpler approach here would be to simply record changes through method_missing (like what Jake did here), then send only the changed values `def encode(opts={})`.

Thoughts?

rafaelfranca:
Yeah, that could work.

maximebedard:
I've updated to use `#encode` instead, seems a little simpler and achieve exactly the same thing. 

@JakeCataford @rafaelfranca Mind 👀 ?

Also something seems wrong with CI and Gemfile_ar40, I'm able to reproduce locally but not really sure why this would happen here.

test/dirty_attributes_encoder_test.rb
```diff
@@ -0,0 +1,93 @@
...
+      attr_other: "some other attr",
+    }.to_json
+
+    fake_resource = Fake.new
+    fake_resource.save
+    fake_resource
+  end
+
+  def assert_last_request_body(expected)
+    assert FakeWeb.last_request.body, expected
```
rafaelfranca: should not this be `assert_equal`?

maximebedard:
@rafaelfranca @maartenvg I've updated the PR to use `encode` instead and removed a hack to always include the root in json during the encoding and do this only in the as_json call. 

Let me know what you think 🙇

maartenvg:
`DirtyAttributesEncoder` looks good, but having a hard time figuring out whether the changes in `lib/shopify_api/resources/base.rb` is a no-op refactor in all situations for all users.

maximebedard:
> lib/shopify_api/resources/base.rb is a no-op refactor in all situations for all users.

I feel like it should be a no-op, I just can't understand why the root was always injected in the `#encode` method :/

maximebedard:
@maartenvg would you like me to extract the changes to base into a separate PR?

maartenvg:
If they're not required for this PR: yes please, always :)

maximebedard:
Yes they are unfortunately. Although could ship the refactor first and address this PR then

maartenvg:
I am a strong proponent of PRs that only do one thing and having (required) refactors that take away from the actual change. So yea, splitting is pretty much always my preference.

rafaelfranca:
Making the commits atomic also help to explain why a change is necessary. But if you can make that changing without the changes on this PR, splitting make sense.

JakeCataford:
Closing this because its stale, we can always circle back

PR #321, opened 2016-11-30
Added StorefrontAccessToken to the gem

yomexzo:
👀  @Shopify/channels-dev-pod, @Shopify/api-documentation, @Shopify/api


yomexzo:
👀 @lreeves , @Matt888

lreeves:
I think this is good provided that these are just the basic operations and we're (well, @Matt888 I mean ;-) are trying to the common use case of channels requiring a per-shop access token that can be created upon install.

PR #323, opened 2017-01-09
Add product/collection listings to ShopifyAPI gem

Matt888:
This is the first step towards moving the documentation from https://help.shopify.com/api/sdks/sales-channel-sdk/reading-products to the API reference at https://help.shopify.com/api/reference.  

If we're going to move the docs over, we should also make this officially part of the rest API in the gem.

For: https://github.com/Shopify/channels-platform/issues/572

@kevinarcher @ilikeorangutans @swalkinshaw


test/collection_listing_test.rb
```diff
@@ -0,0 +1,25 @@
...
+    collection_listings = ShopifyAPI::CollectionListing.find(:all, params: { application_id: 999 })
+
+    assert_equal 1, collection_listings.count
+    assert_equal 1, collection_listings.first.collection_id
+    assert_equal 'Home page', collection_listings.first.title
+  end
+
+  def test_get_collection_listing_for_collection_id
+    fake "applications/999/collection_listings/1", method: :get, status: 201, body: load_fixture('collection_listing')
+    fake "applications/999/collection_listings//1/product_ids", method: :get, status: 201, body: load_fixture('collection_listing_product_ids')
```
Matt888: For some reason this property calls the API with a double slash.  Any ideas how this can be resolved?

swalkinshaw:
A lot of our apps were defining these resources themselves.

Example: https://github.com/Shopify/facebook-commerce/blob/ebd66d0352b18356c902f8f4b3812ea7281c17f6/lib/shopify_api/product_listing.rb

It looks like this shouldn't conflict with those?

Matt888:
Yeah that's a hack that our apps do.  I can open separate PRs to remove these hacks and use the official gem update.  They shouldn't conflict because they don't define anything that conflicts AFAIK.

Matt888:
//cc @lreeves

PR #325, opened 2017-01-20
Remove automatic encoding of the root key on ::Base

maximebedard:
Instead of always encoding the root key, we will reuse the `include_root_in_json` parameter and set it to true instead. Because in the end, it's exactly what we are doing, except that this will allow us to customize serializations in some very specific cases during the encoding phase without having to worry if the root key is present or not.

Let me know what you think 👍  @maartenvg @Shopify/api

### DIFF
--- a/lib/shopify_api/resources/base.rb
+++ b/lib/shopify_api/resources/base.rb
@@ -9,7 +9,7 @@ module ShopifyAPI
     self.headers['User-Agent'] = ["ShopifyAPI/#{ShopifyAPI::VERSION}",
                                   "ActiveResource/#{ActiveResource::VERSION::STRING}",
                                   "Ruby/#{RUBY_VERSION}"].join(' ')
-    self.site = "https://maxime-bedard.myshopify.com"
+    # self.site = "https://maxime-bedard.myshopify.com"
 
     def encode(options = {})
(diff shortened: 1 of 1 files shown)


maartenvg:
There don't seem to be any tests in `base_test.rb` that cover any of this root logic (for either JSON & XML). That seems useful to add.

maartenvg:
I ran this against Shopify's internal tests and got a whole bunch of failures. You might want to look into that.

lib/shopify_api/resources/base.rb
```diff
@@ -9,22 +9,11 @@ class InvalidSessionError < StandardError; end
     self.headers['User-Agent'] = ["ShopifyAPI/#{ShopifyAPI::VERSION}",
                                   "ActiveResource/#{ActiveResource::VERSION::STRING}",
                                   "Ruby/#{RUBY_VERSION}"].join(' ')
+    # self.site = "https://maxime-bedard.myshopify.com"
```
cefigueiredo: You forgot a probable a personal setting commented here...

cefigueiredo: You forgot commented something that looks a debugging setting here

PR #330, opened 2017-02-13
Accept the extra argument in Session.temp

humancopy:


lib/shopify_api/session.rb
```diff
@@ -67,10 +67,10 @@ def encoded_params_for_signature(params)
       end
     end
 
-    def initialize(url, token = nil, extra = {})
+    def initialize(url, token = nil, extra = nil)
       self.url = self.class.prepare_url(url)
       self.token = token
-      self.extra = extra
+      self.extra = extra || {}
```
cefigueiredo: It would not be necessary check if `extra` was set if you let the `initialize` method the way it was.

lib/shopify_api/session.rb
```diff
@@ -19,11 +19,11 @@ def setup(params)
         params.each { |k,value| public_send("#{k}=", value) }
       end
 
-      def temp(domain, token, &block)
-        session = new(domain, token)
+      def temp(domain, token, extra = nil, &block)
```
cefigueiredo: if you want set a default value for `extra` why not use an empty hash as the `initialize` does already?

humancopy: Just thought to not repeat the code, but I agree with you, will fix it in next commit ... ;)

humancopy:
I've added a class attribute `extra` to `ShopifyAPI::Base` so it could be retrieved later when needed (as `original_extra`). Tests are also in place now :)

PR #332, opened 2017-02-16
Fix the endpoint for CustomerSavedSearch#customers

### LINKED ISSUE #331: CustomerSavedSearch#customers hits the wrong url
When I call `CustomerSavedSearch#customers` right now, the library calls a url like:
```
/admin/customers/search.json?saved_search_id=2067224913
```
This doesn't work - it ends up returning all of the customers for the shop since `saved_search_id` is not a [listed parameter](https://help.shopify.com/api/reference/customer#search).

The api should be calling this url instead:
```
/admin/customer_saved_searches/2067224913/customers.json
```

which will actually return the customers for the saved_search.

JamesHageman:
Fixes #331

### DIFF
--- a/lib/shopify_api/resources/customer_saved_search.rb
+++ b/lib/shopify_api/resources/customer_saved_search.rb
@@ -3,8 +3,7 @@ require 'shopify_api/resources/customer'
 module ShopifyAPI
   class CustomerSavedSearch < Base
     def customers(params = {})
-      url = "/admin/customer_saved_searches/#{id}/customers.json"
-      Customer.find(:all, from: url, params: params)
+      Customer.search(params.merge({:customer_saved_search_id => self.id}))
     end
   end
 end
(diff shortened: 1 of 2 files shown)


jamiemtdwyer:
This looks sane to me, but I'd have to defer to someone else for a formal review as I'm not super well-versed in activeresource

@maartenvg @Matt888 👀

maartenvg:
Using `Customer.search` was actually correct. `/admin/customer_saved_searches/1231231/customers.json` is just a convenient URL that maps directly to the search with the `customer_saved_search_id` set. The only problem is the ID, it should be `customer_saved_search_id`, not `saved_search_id`.

But if you want to change it, you shouldn't hard-code the path. You can look at `DraftOrder#send_invoice` for some direction.

JamesHageman:
@maartenvg I can change it to `customer_saved_search_id`, I think that's a simpler change anyway.

JamesHageman:
@maartenvg do you think I could merge this?

JamesHageman:
@jamiemtdwyer @maartenvg Would we be able to cut a release with this commit in it? The shopify core has tests that run against CustomerSavedSearch that I am trying to fix [here](https://github.com/Shopify/shopify/blob/master/test/api/unit/customer_saved_search_api_test.rb#L210)

maartenvg:
Go for it.

PR #333, opened 2017-02-22
Added complete endpoint to draft order

tanner-rutgers:
This adds the `complete` endpoint for Draft Orders, now that the Draft Orders API has exposed this endpoint.


tanner-rutgers:
[[Complete] Update Shopify API gem](https://trello.com/c/rj0QB443/459-complete-update-shopify-api-gem)

lib/shopify_api/resources/draft_order.rb
```diff
@@ -6,5 +6,9 @@ def send_invoice(draft_order_invoice = ShopifyAPI::DraftOrderInvoice.new)
       resource = post(:send_invoice, {}, draft_order_invoice.encode)
       ShopifyAPI::DraftOrderInvoice.new(ShopifyAPI::DraftOrder.format.decode(resource.body))
     end
+
+    def complete(options = {})
+      load_attributes_from_response(patch(:complete, {}, options.to_json))
```
maartenvg: This should be `put`, since `patch` wasn't added until [ActiveResource 4.1.0](https://github.com/rails/activeresource/commit/55cfbee6f7671c966be9135e7a13f1ef1b0f710a) and we have support for AR 3.x too.

tanner-rutgers: Unfortunately sending a `put` won't work, since the endpoint is a `patch` (and yes, I've tried sending `put` and get a 406 (?))

tanner-rutgers: However, I did run into this issue (see failing tests below).

I'm thinking the only way to properly fix this is to change the endpoint itself to a put.

maartenvg: Oh so `put` in the route accepts both `put` and `patch`, but an explicit `patch` only accepts `patch` and not `put`? That's kinda an important detail. 

But yea, changing the endpoint itself sounds good to me.

tanner-rutgers: >Oh so put in the route accepts both put and patch, but an explicit patch only accepts patch and not put? That's kinda an important detail.

Exactly.  Odd, I know, but that's what I've experienced.

Since a `put` will accept both patch and put, that definitely sounds like the right thing to do.

test/fixtures/draft_order_completed.json
```diff
@@ -0,0 +1,159 @@
...
+    "taxes_included": false,
+    "currency": "CAD",
+    "subtotal_price": "1007.41",
+    "total_tax": "0.00",
+    "total_price": "1027.41",
+    "invoice_sent_at": null,
+    "created_at": "2017-02-02T13:14:38-05:00",
+    "updated_at": "2017-02-02T13:14:38-05:00",
+    "tax_exempt": false,
+    "completed_at": null,
```
dominiquesr: This shouldn't be null; and should also be asserted in the tests above. This code gets used as a reference so we should reflect reality here.

tanner-rutgers:
Ready for 👀 again

test/draft_order_test.rb
```diff
@@ -111,4 +111,30 @@ def test_get_metafields_for_draft_order
...
+
+  def test_complete_draft_order_with_params
+    completed_fixture = load_fixture('draft_order_completed')
+    completed_draft = ActiveSupport::JSON.decode(completed_fixture)['draft_order']
+    complete_params = { payment_pending: true }
+    fake 'draft_orders/517119332/complete', method: :put, status: 200, body: completed_fixture
+
+    @draft_order.complete(complete_params)
+
+    assert_request_body complete_params.to_json
```
maartenvg: Interesting, I didn't really expect this to be part of the body, I thought they were going to be simple query params.

tanner-rutgers: The endpoint will support either one, but the API gem will send them in the body

maartenvg: Can you make the gem send it as query params? That's what's we should be documenting too. If we ever want to change what body we want to accept on that endpoint, we'd have to change the gem again, and have to deal with users using older versions of the gem.

tanner-rutgers:
Modified gem to use params instead of body, ready for 👀 again :)

PR #334, opened 2017-03-07
add checkout settings model

zacharyreynochiasson:
Adds checkout settings api model


jamiemtdwyer:
@maartenvg @Matt888 mind having a 👀 guys?

Matt888:
What are checkout settings?

maartenvg:
I believe this is a bit premature. Let's decide on whether we're going to expose this first.

zacharyreynochiasson:
No immediate need.

PR #339, opened 2017-03-17
Adding search and auto_paging_each

iloveitaly:
The `Customer` object has a `search` class method, but none of the other objects do. I think adding a `search` class method to the base class would make the library much more intuitive.

Also, adding a search pagination convenience method would be helpful to new users as well. The `auto_paging_each` pattern has been ripped from Stripe's ruby library.

### DIFF
--- a/lib/shopify_api/resources/base.rb
+++ b/lib/shopify_api/resources/base.rb
@@ -28,6 +28,33 @@ module ShopifyAPI
     end
 
     class << self
+      def self.list(params = {})
+        find(:all, params: params)
+      end
+
+      def self.auto_paging_each(opts = {})
+        opts[:limit] ||= 50
+        page_limit = opts[:limit]
+
+        total_count = self.count(opts)
+        pages = total_count/page_limit + 1
+        current_page = opts[:page] || 0
+        records = []
+
+        while current_page != pages
+          self.list(opts.merge(page: current_page)).each do |record|
+            records << record
+
+            yield(record) if block_given?
+          end
+
+          current_page += 1
+        end
+
+        records
+      end
+    end
+
       if ActiveResource::Base.respond_to?(:_headers) && ActiveResource::Base.respond_to?(:_headers_defined?)
         def headers
           if _headers_defined?


iloveitaly:
Friendly ping here! Any interest in merging this in?

jamiemtdwyer:
Adding search seems sane to me, but we should remove it from customer.rb if we add it here.

As for auto paging, seems like the API call limit might cause difficulties here?

@maartenvg or @kevinhughes27 any opinions here? 🙏

lib/shopify_api/resources/base.rb
```diff
@@ -28,6 +28,32 @@ def as_json(options = nil)
...
+            records << record
+
+            yield(record) if block_given?
+          end
+
+          current_page += 1
+        end
+
+        records
+      end
```
kevinhughes27: I usually write this to keep incrementing the page until no records are returned.

lib/shopify_api/resources/base.rb
```diff
@@ -28,6 +28,32 @@ def as_json(options = nil)
     end
 
     class << self
+      def search(params = {})
+        find(:all, params: params)
+      end
```
kevinhughes27: this isn't a very efficient search - If I remember correctly won't this fetch records until something is found?

iloveitaly: @kevinhughes27 This was a poor name for this: I updated the PR to name it `list`. The `find :all` call doesn't actually search it grabs a list from the shopify endpoints

iloveitaly:
> As for auto paging, seems like the API call limit might cause difficulties here?

I've run into API limits when paging through results, here's the solution I've implemented:

https://gist.github.com/iloveitaly/cf0566bd598732c1e81bad0de1be4ff7

Happy to submit a PR for this as well—would this be something you'd include in the API behind an optional configuration flag?

maartenvg:
I don't think this auto paging behaviour belongs in the gem, especially not on all resources. How, when and how far you want to page is very application specific, and doing the wrong thing can be very expensive (for the user and for Shopify). By making it part of the general library it promotes a behaviour that is unwanted in many cases, which can unintentionally cause lots of unnecessary calls to Shopify.

IMHO this kind of logic belongs to a wrapping layer in your own code.

iloveitaly:
@maartenvg That makes sense! I think it would be helpful to have this semi-generic pagination and request limit backoff code available to the larger community. I'll wrap this into a standalone gem and release on GitHub.

Thanks again for your help and review here!

iloveitaly:
I've pulled this, and some other convenience utilities, into a separate gem: https://github.com/iloveitaly/shopify_api_extensions

nwtn:
Closing because the change was extracted into a separate gem

PR #340, opened 2017-03-21
Add ResourceFeedback and Bump Version to 4.3.8

jessicaxiejw:
Add ResourceFeedback and bump gem version to 4.3.8

Issue: https://github.com/Shopify/product-publishing-feedback/issues/193

### DIFF
--- a/lib/shopify_api/resources/base.rb
+++ b/lib/shopify_api/resources/base.rb
@@ -81,6 +81,7 @@ module ShopifyAPI
     end
 
     private
+
     def only_id
       encode(:only => :id, :include => [], :methods => [])
     end
--- a/lib/shopify_api/resources/resource_feedback.rb
+++ b/lib/shopify_api/resources/resource_feedback.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module ShopifyAPI
+  class ResourceFeedback < Base
+    class ExistingFeedbackSaved < StandardError; end
+    include DisablePrefixCheck
+
+    STATE_SUCCESS = 'success'
+    STATE_REQUIRES_ACTION = 'requires_action'
+
+    conditional_prefix :product, false
+
+    EXISTING_FEEDBACK_SAVED_ERROR_MESSAGE = 'WARNING: Attempted call to ShopifyAPI::ResourceFeedback#save' \
+      'after it was already persisted. Updating an existing ShopifyAPI::ResourceFeedback is not supported.'
+
+    def save
+      return super unless persisted?
+      raise ExistingFeedbackSaved.new(EXISTING_FEEDBACK_SAVED_ERROR_MESSAGE)
+    end
+  end
+end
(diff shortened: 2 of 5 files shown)


jessicaxiejw:
I am not sure if we need to ping someone else.

test/resource_feedback_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def test_save_with_product_id_resource_feedbacks_endpoint
+    body = { resource_feedback: {} }.to_json
+    fake 'products/42/resource_feedbacks', method: :post, body: body
+    ShopifyAPI::ResourceFeedback.new(product_id: 42).save
+    assert_request_body body
+  end
+
+  def test_save_notify_via_bugsnag_when_already_persisted
```
jamiemtdwyer: is the name of this test accurate?

lib/shopify_api/resources/resource_feedback.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module ShopifyAPI
+  class ResourceFeedback < Base
+    class ExistingFeedbackSaved < StandardError; end
+    include DisablePrefixCheck
+
+    STATE_SUCCESS = 'success'
+    STATE_REQUIRES_ACTION = 'requires_action'
```
jamiemtdwyer: we've never used this pattern of string constants before, but i'm not opposed to it

lib/shopify_api/resources/base.rb
```diff
@@ -81,6 +81,7 @@ def persisted?
     end
 
     private
+
```
jamiemtdwyer: was this whitespace added to follow some convention or by accident?

test/resource_feedback_test.rb
jessicaxiejw: Good catch! If they are 3P, the exception may or may not be show on bugsnag. I updated the function name so it's more correct.

lib/shopify_api/resources/base.rb
jessicaxiejw: Yeah, in Shopify core, you are supposed to have a new line after private. I saw it and fixed it here assuming gem should follow core's coding style guideline.

lib/shopify_api/resources/resource_feedback.rb
chrisbutcher: Yeah we don't need to add these constants. We did it thinking it might be a convenience, but it's not going to save much typing, given how long the namespace on each constant would be `ShopifyAPI::ResourceFeedback::STATE_...`

We also added these at a time when there were a lot more states and topics still existed.

jessicaxiejw: I removed the constants.

jessicaxiejw:
@jamiemtdwyer I tested it by checkout the branch locally then change the gem file in facebook-commerce to point to the path where I clone the shopify_api and do a bundle install.

jessicaxiejw:
Should I wait on merging the gem? I am not sure how the release works.

PR #343, opened 2017-03-22
Singularize ResourceFeedback collection_name

chrisbutcher:
Singularize ResourceFeedback `collection_name`, to coincide with API path change to rename endpoint to `/admin/resource_feedback`

### DIFF
--- a/lib/shopify_api/resources/resource_feedback.rb
+++ b/lib/shopify_api/resources/resource_feedback.rb
@@ -5,6 +5,7 @@ module ShopifyAPI
     class ExistingFeedbackSaved < StandardError; end
     include DisablePrefixCheck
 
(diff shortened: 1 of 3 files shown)


jamiemtdwyer:
If this is a breaking change, we should bump the version and re-deploy

maartenvg:
Considering our chat yesterday, I assume this is no longer necessary?

chrisbutcher:
@maartenvg I think this change still applies. We'll keep index returning an array of (by default 1 feedback record). This change is simply to make the name less awkward to outside parties, since the word `feedback` is already plural.

chrisbutcher:
@jamiemtdwyer Yep, for sure

maartenvg:
> since the word feedback is already plural.

:+1:

PR #346, opened 2017-03-29
Add support for customer invites

elsom25:
Adds `ShopifyAPI::CustomerInvite` and exposes `Customer#send_invite`

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,8 @@
+== Version 4.4.0
+
+* Added `ShopifyAPI::CustomerInvite`
+* Support for Customer#send_invite endpoint
+
 == Version 4.3.8
 
 * Added `ShopifyAPI::ResourceFeedback`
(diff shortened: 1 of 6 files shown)


CHANGELOG
```diff
@@ -1,3 +1,8 @@
+== Version 4.4.0
```
dominiquesr: Why 4.4 and not 4.3.9?

elsom25: because it changes the public API? VS being a non-public facing change that `_._.X` denotes?

lugray: 👍 to 4.4: https://github.com/Shopify/shopify_api/blob/master/RELEASING#L3
> Given a version number MAJOR.MINOR.PATCH, increment the:
MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible manner, and
PATCH version when you make backwards-compatible bug fixes.

PR #352, opened 2017-04-26
Fix Rails 5 deprecation warning with_indifferent_access

andrewle:
When running the shopify_api with Rails 5, you'll see the following
deprecation warning:

```
DEPRECATION WARNING: Method with_indifferent_access is deprecated and will be
removed in Rails 5.1, as `ActionController::Parameters` no longer inherits from
hash. Using this deprecated behavior exposes potential security problems. If
you continue to use this method you may be creating a security vulnerability in
your app that can be exploited. Instead, consider using one of these documented
methods which are not deprecated:
http://api.rubyonrails.org/v5.0.2/classes/ActionController/Parameters.html
(called from validate_signature at /../shopify/shopify_api/lib/shopify_api/session.rb:54)
```

This patch checks for and calls `to_unsafe_hash` on the `params`
argument to convert the params to a hash before calling
`with_indifferent_access`. It supports working with
`ActionController::Parameters` as well as plain hashes because the tests
use plain hashes.

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -51,7 +51,7 @@ module ShopifyAPI
       end
 
       def validate_signature(params)
-        params = params.with_indifferent_access
+        params = (params.respond_to?(:to_unsafe_hash) ? params.to_unsafe_hash : params).with_indifferent_access
         return false unless signature = params[:hmac]
 
         calculated_signature = OpenSSL::HMAC.hexdigest(OpenSSL::Digest::SHA256.new(), secret, encoded_params_for_signature(params))


andrewle:
@ShayneP @jamiemtdwyer 👋 👋 👋 😃

jamiemtdwyer:
A bit out of my scope but I'm trying to get 👀  on it!

bdewater:
Hey 👋  thanks for your contribution!

> It supports working with ActionController::Parameters as well as plain hashes because the tests use plain hashes.

I'm not familiar with this codebase, but if in production this code always works on AC::Parameters and the tests use plain hashes, it sounds like the tests are not representative and should be changed so `validate_signature` can always call `params.to_unsafe_hash`. Correct me if I'm wrong!

If that assumption is correct, a small nitpick 😄  [`to_unsafe_hash`](http://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-to_unsafe_h) already returns a HashWithIndifferentAccess.

lib/shopify_api/session.rb
```diff
@@ -51,7 +51,7 @@ def prepare_url(url)
       end
 
       def validate_signature(params)
-        params = params.with_indifferent_access
+        params = (params.respond_to?(:to_unsafe_hash) ? params.to_unsafe_hash : params).with_indifferent_access
```
rafaelfranca: The correct way to fix this deprecation is calling `to_h`, but that is going to show another deprecation message unless the application already fixed it or making sure that the argument of this method is a hash with indifferent access (since `Parameters` is already one and you don't need to make another one again) or a hash with symbols (since it is all we want). That being said have we tried to remove the `with_indifferent_access` call?

yjaspar:
Has this been abandoned? I'm running into the issue right now 😄

lib/shopify_api/session.rb
yjaspar: Has this been abandoned? I'm running into the issue right now 😄 
In 5.2.1 it's no longer a deprecation warning, they simply removed the method 🎉 

Would it be crazy to just use `.permit!` instead?

@rafaelfranca @bdewater @andrewle

rafaelfranca: That has the same as `to_unsafe_hash`, which BTW is the same as `permit!.to_h`. I think the proper fix would remove the `with_indifferent_access` call from here.

yjaspar: I don't even think we need `.to_h`. Why would we need it?

rafaelfranca: because `with_indifferent_access` is not defined in `Parameters` instance.

rafaelfranca: oh, yeah. We can, but I really prefer if we don't call neither of those methods. `permit!` is not required since we access the hash using `[]` and we don't need `with_indifferent_access`.

yjaspar: @rafaelfranca 
Created a new PR as I don't have access to push to this one -- I guess this one can then be closed. I can make updates to my branch or give access to people if needed.

https://github.com/Shopify/shopify_api/pull/468

PR #355, opened 2017-04-27
Bump gem to version 4.7.0

peterjm:
Bumping the version before releasing a new gem.


CHANGELOG
```diff
@@ -1,5 +1,9 @@
 == Version 4.6.0
```
kenrose: I'm late to the party, but shouldn't this line have been `4.7.0`?

peterjm: It sure should have been. I got impatient to release the new gem version and merged without review, figuring this small change wouldn't need one. Shows that even the smallest changes should be properly reviewed 😄 

Fixing it in https://github.com/Shopify/shopify_api/pull/356

PR #356, opened 2017-04-27
Fix bug with listing classes

peterjm:
The listing classes were broken, because they were trying to use the `id` field as a primary key, which doesn't exist for listings. This tells ActiveResource to use the underlying object id as the primary key.

This was the reason the URL for fetching the product ids for collections was "/admin/collection_listings//123/product_ids.json" (note the double "//") -- ActiveResource was trying to fill in an id between the slashes, but just ended up rendering an empty string (`nil.to_s`)

### DIFF
--- a/test/collection_listing_test.rb
+++ b/test/collection_listing_test.rb
@@ -12,14 +12,30 @@ class CollectionListingTest < Test::Unit::TestCase
     assert_equal 'Home page', collection_listings.first.title
   end
 
-  def test_get_collection_listing_for_collection_id
+  def test_get_collection_listing
     fake "collection_listings/1", method: :get, status: 201, body: load_fixture('collection_listing')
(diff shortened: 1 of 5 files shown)


test/collection_listing_test.rb
```diff
@@ -12,14 +12,30 @@ def test_get_collection_listings
     assert_equal 'Home page', collection_listings.first.title
   end
 
-  def test_get_collection_listing_for_collection_id
+  def test_get_collection_listing
     fake "collection_listings/1", method: :get, status: 201, body: load_fixture('collection_listing')
-    fake "collection_listings//1/product_ids", method: :get, status: 201, body: load_fixture('collection_listing_product_ids')
```
kenrose: Lol.  This bug was actually codified in a test too.

PR #357, opened 2017-04-27
Add support for marketing event engagements

kingscott:
Hey Shopify team!

With the new announcement, and 1.0, of the Marketing Events API we are about to starting using it at Sweet Tooth to show real engagement statistics to merchants. 

As documented, marketing events have associated engagements which track views, clicks, etc. The endpoint is as follows:

image

Right now there is no official support for engagement via this gem. And ideally, for best practice, we want to avoid directly using `ActiveResource` REST methods from within our codebase. This PR introduces a small wrapper, that allows us to nicely create engagements. 

The specs are included as well.

### DIFF
--- a/lib/shopify_api/resources/engagement.rb
+++ b/lib/shopify_api/resources/engagement.rb
@@ -1,0 +1,4 @@
+module ShopifyAPI
+  class Engagement < Base
+  end
+end
--- a/lib/shopify_api/resources/engagement.rb
+++ b/lib/shopify_api/resources/engagement.rb
@@ -1,4 +1,0 @@
-module ShopifyAPI
-  class Engagement < Base
-  end
-end
--- a/lib/shopify_api/resources/marketing_event.rb
+++ b/lib/shopify_api/resources/marketing_event.rb
@@ -1,7 +1,7 @@
 module ShopifyAPI
   class MarketingEvent < Base
     def add_engagements(engagements)
-      engagements = { engagements: Array(engagements) }
+      engagements = { engagements: Array.wrap(engagements) }
       post(:engagements, {}, engagements.to_json)
     end
   end
--- a/lib/shopify_api/resources/marketing_event.rb
+++ b/lib/shopify_api/resources/marketing_event.rb
@@ -1,4 +1,8 @@
 module ShopifyAPI
   class MarketingEvent < Base
+    def add_engagements(engagements)
+      engagements = {engagements: Array(engagements)}
+      post(:engagements, {}, engagements.to_json)
+    end
   end
 end
--- a/lib/shopify_api/resources/marketing_event.rb
+++ b/lib/shopify_api/resources/marketing_event.rb
@@ -1,7 +1,7 @@
 module ShopifyAPI
   class MarketingEvent < Base
     def add_engagements(engagements)
-      engagements = {engagements: Array(engagements)}
+      engagements = { engagements: Array(engagements) }
       post(:engagements, {}, engagements.to_json)
     end
   end
--- a/test/fixtures/engagement.json
+++ b/test/fixtures/engagement.json
@@ -1,0 +1,15 @@
+{
+  "engagements": [
+    {
+      "occurred_on": "2017-04-20",
+      "impressions_count": null,
+      "views_count": null,
+      "clicks_count": 10,
+      "shares_count": null,
+      "favorites_count": null,
+      "comments_count": null,
+      "ad_spend": null,
+      "is_cumulative": true
+    }
+  ]
+}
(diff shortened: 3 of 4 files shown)


jamiemtdwyer:
Thanks for contributing, Scott! 🙌 

@manasa91 would you mind doing a review here?

test/marketing_event_test.rb
```diff
@@ -46,4 +46,23 @@ def test_count_marketing_events
...
+      impressions_count: nil,
+      views_count: nil,
+      clicks_count: 10,
+      shares_count: nil,
+      favorites_count: nil,
+      comments_count: nil,
+      ad_spend: nil,
+      is_cumulative: true
+    )
+    marketing_event.post(:engagements, {}, {engagements: Array(engagement)}.to_json)
```
lulessa: @kingscott thank you for adding this action to MarketingEvent resource!

should this be `marketing_event.add_engagements(engagement)` to test the `MarketingEvent#add_engagements` method?

kingscott:
Good point, thanks @lulessa! 🐐

lib/shopify_api/resources/marketing_event.rb
```diff
@@ -1,4 +1,8 @@
 module ShopifyAPI
   class MarketingEvent < Base
+    def add_engagements(engagements)
+      engagements = {engagements: Array(engagements)}
```
manasa91: add spaces inside curly braces

kingscott:
Done @manasa91 👍

test/fixtures/engagement.json
```diff
@@ -0,0 +1,15 @@
+{
+  "engagements": [
```
manasa91: can we call this marketing_engagement, because there can be different attributes for other engagemenrs such as content engagements

lib/shopify_api/resources/engagement.rb
```diff
@@ -0,0 +1,4 @@
+module ShopifyAPI
```
manasa91: Can we remove this file because we do not have a corresponding resource in shopify called enagement

manasa91: Looks like the engagements resourse i.e. ContentEngagement is deprecated and we just write to a kafka stream instead

kingscott: So for now, creating engagements will just be the `add_engagements` method along with a JSON object?

manasa91: yep!

kingscott:
Update @manasa91 👍

kingscott:
The build is failing for reasons out of my control...

kingscott:
Could this get another look over @manasa91? 😄

meunierd:
@kingscott LGTM we've got partners asking about it.

test/marketing_event_test.rb
```diff
@@ -46,4 +46,23 @@ def test_count_marketing_events
     marketing_events_count = ShopifyAPI::MarketingEvent.get(:count)
     assert_equal 2, marketing_events_count
   end
+
+  def test_add_engagements
```
manasa91: you could refactor this to 

` def test_add_engagements
     fake "marketing_events/1", method: :get, body: load_fixture('marketing_event')
     marketing_event = ShopifyAPI::MarketingEvent.find(1)
     engagement_fixture = load_fixture('engagement')
     engagements = ActiveSupport::JSON.decode(engagement_fixture)['engagements']
     fake "marketing_events/1/engagements", method: :post, status: 201, body: engagement_fixture
    
     assert marketing_event.add_engagements(engagements[0])
   end`

kingscott: I'm not sure this is an ideal refactor - mainly regarding the `assert`. This would be asserting the existence of the returned result of the post, correct? For testing purposes, if we got a 404, that assert would return true because it would have a returned response. 

I would understand loading the fixture instead, since that would be what the request body _would_ be. Any ideas on reaching a common ground @manasa91?

manasa91: that makes sense, and im not sure if checking equality of returned fields in the body is any different from the current implementation so this is good to go 👍

kingscott:
Are there timelines for when these sort of things are released? Not trying to bother anyone, just curious before we ship this to prod if I should keep the fork in our Gemfile, or if it'll be merged to master. :) Thanks!

meunierd:
This should hit master ASAP. 👍

kingscott:
Woohoo!

jamiemtdwyer:
Without a version bump, gem consumers won't see these changes downstream. I'll bump the version in a separate PR and deploy to Rubygems.

jamiemtdwyer:
Nvm, I see it was added in a separate PR - kcco :+1:

Appreciate the contribution @kingscott !

PR #407, opened 2018-01-08
Test build on multiple Ruby versions

nicolasleger:
Note that `rails-observers` (`0.1.5`) requires Ruby version `>= 2.2.2`.


nicolasleger:
@jamiemtdwyer Tests now pass. I suggest to squash before merging it.

nwtn:
This was recently updated to test on the versions we’re currently targeting.

PR #410, opened 2018-01-31
Fetch order with custom properties

rebeccajfriedman:
We are seeing an error where `ActiveResource` tries to (incorrectly) create a resource to load an order's line_item property hash value. The workaround in this PR is to rescue the `NameError` and return the attributes as a hash (and not a resource).

### DIFF
--- a/test/fixtures/order_with_properties.json
+++ b/test/fixtures/order_with_properties.json
@@ -1,0 +1,373 @@
+{
+  "order": {
+    "buyer_accepts_marketing": false,
+    "cancel_reason": null,
+    "cancelled_at": null,
+    "cart_token": "68778783ad298f1c80c3bafcddeea02f",
+    "checkout_token": null,
+    "closed_at": null,
+    "confirmed": false,
+    "created_at": "2008-01-10T11:00:00-05:00",
+    "currency": "USD",
+    "email": "bob.norman@hostmail.com",
+    "financial_status": "authorized",
+    "fulfillment_status": null,
+    "gateway": "authorize_net",
+    "id": 450789469,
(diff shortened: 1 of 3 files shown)


rebeccajfriedman:
My solution to override the `initialize` method for the `Property` class is only working for activeresource 3.1 and above because the method params changed.

jamiemtdwyer:
This gem already has a dependency for a higher version of activeresource, so I don't think that should be an issue.

``` 
 gem "activeresource", "~> 4.1"
```

rebeccajfriedman:
Awesome :) What are your thoughts about this solution? And what's the timeline to get something like this approved and pushed?

jamiemtdwyer:
I don't see anything wrong with this workaround, but I wouldn't mind seeing a test for creating an order with line item properties.

I'd be happy to 👍 once tests are passing.

test/fixtures/order_with_properties.json
```diff
@@ -0,0 +1,373 @@
...
+        "sku": "",
+        "taxable": true,
+        "title": "Create Your Own Coffee 3 Pack (8oz bags)",
+        "variant_id": 4107806867497,
+        "variant_title": "",
+        "vendor": "OMC",
+        "name": "Create Your Own Coffee 3 Pack (8oz bags)",
+        "variant_inventory_management": null,
+        "properties": [
+          {
```
rebeccajfriedman: This is the ill-formatted property that we are getting in several channels.

rebeccajfriedman:
This PR will clean up the allowable failures: https://github.com/Shopify/shopify_api/pull/411

PR #415, opened 2018-02-23
add inventory api

jtgrenz:
Added:
- InventoryItem: https://help.shopify.com/api/reference/inventoryitem
- InventoryLevel: https://help.shopify.com/api/reference/inventorylevel

[Multi-location Migration Guide](https://help.shopify.com/api/tutorials/inventory-migration-guide)

First time contributing to a gem and I'm not 100% if I've mocked these tests correctly. Would really appreciate a solid 🎩


lib/shopify_api/resources/inventory_level.rb
```diff
@@ -0,0 +1,49 @@
+
+# frozen_string_literal: true
+module ShopifyAPI
+  class InventoryLevel < Base
+    def self.element_path(prefix_options = {}, query_options = nil)
+      prefix_options, query_options = split_options(prefix_options) if query_options.nil?
+      "#{prefix(prefix_options)}#{collection_name}.#{format.extension}#{query_string(query_options)}"
```
celsodantas: Can you explain why do we need this?

lib/shopify_api/resources/inventory_level.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    def destroy
+      element_path
```
celsodantas: useless call? 😬

goncalopereira:
Will this be released soon ? Already saw the deprecation warning on the API documentation :)

lib/shopify_api/resources/inventory_level.rb
jtgrenz: `element_path` is part of ActiveResource::Base and it helps construct the url for the custom method. The built in `[element_path](https://github.com/rails/activeresource/blob/b71ded7e1b9004c7a24f28203c35c7ad24da6617/lib/active_resource/base.rb#L769)` method is 

```ruby
def element_path(id, prefix_options = {}, query_options = nil)
  check_prefix_options(prefix_options)

  prefix_options, query_options = split_options(prefix_options) if query_options.nil?
  "#{prefix(prefix_options)}#{collection_name}/#{URI.parser.escape id.to_s}#{format_extension}#{query_string(query_options)}"
end
```

Which adds the id of resource in the path. So this becomes `/admin/inventory_levels/#{ inventory_level.id }.json?#{ params }`. We don't expose inventory_level ids so we need to send DELETE to the base `/admin/inventory_levels.json` endpoint with id params. Redefining element_path here removes the id and sends DELETE to the correct path. 

I was stuck trying to get this to go to the right URL, but our asset resource also needs a special path and I lifted the code from [asset.rb](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/resources/asset.rb#L41).

In hindsight  though, I don't need to explicitly call it from destroy since its called by default.

celsodantas: oh ok. What do you think about adding a comment with the same explanation you added here to this method? So just other devs can understand the need of it (and that is only needed become of the 'delete' call).

I saw that you were doing the same as in `asset.rb` but couldn't understand why.

jtgrenz:
@celsodantas explanation added. I also added an inventory_levels method to locations.rb

lib/shopify_api/resources/location.rb
```diff
@@ -1,4 +1,8 @@
 module ShopifyAPI
   class Location < Base
+
+    def inventory_levels
+      ShopifyAPI::InventoryLevel.find :all, from: "/admin/locations/#{id}/inventory_levels.json"
```
Soleone: Nitpick, but I prefer to consistently use parentheses on method calls.

Exceptions to that rule include declarative style method calls in Rails like "has_many :things" or `assert...` calls. Seems like we do this with the `fake` calls in this gem as well. But maybe everything else should stay consistently called with parentheses.

test/inventory_level_test.rb
```diff
@@ -0,0 +1,59 @@
+require 'test_helper'
+
+class InventoryLevelTest < Test::Unit::TestCase
+  def setup
+    super
+    @inventory_level_response = ActiveSupport::JSON.decode load_fixture('inventory_level')
+    @inventory_level = ShopifyAPI::InventoryLevel.new(@inventory_level_response['inventory_level'])
+  end
+
+  test 'Retrieve inventory levels for the specified inventory_items and locations' do
```
Soleone: Nitpick, but typically our test descriptions start lower case.

Normally we start with the method name that is being tested like:

```
test ".find with inventory_item_ids and location_ids returns expected inventory levels"
```

* `#` prefix for instance methods
* `.` prefix for class methods

lib/shopify_api/resources/inventory_level.rb
```diff
@@ -0,0 +1,56 @@
...
+    # /admin/inventory_levels.json?#{ params } instead.
+    #
+    def self.element_path(prefix_options = {}, query_options = nil)
+      prefix_options, query_options = split_options(prefix_options) if query_options.nil?
+      "#{prefix(prefix_options)}#{collection_name}.#{format.extension}#{query_string(query_options)}"
+    end
+
+    def destroy
+      load_attributes_from_response(
+        self.class.delete('/', location_id: location_id,
```
Soleone: This should still fit into a 120 char line.

test/inventory_level_test.rb
```diff
@@ -0,0 +1,59 @@
...
+          status: 200, body: load_fixture('inventory_levels')
+    inventory_levels = ShopifyAPI::InventoryLevel.find(:all, params: params)
+
+    assert inventory_levels.all? { |item|
+      params[:location_ids].include?(item.location_id) &&
+      params[:inventory_item_ids].include?(item.inventory_item_id)
+    }, message: 'Response contained inventory_items or locations not requested.'
+  end
+
+  test 'Adjust an inventory level by 5' do
```
Soleone: For example I would have named this test something like:

```
test "#adjust sets new inventory availability"
```

lib/shopify_api/resources/inventory_level.rb
celsodantas: the data being sent can be added to the payload body, not necessarily the query params. The backend doesn't care where the content is being sent. 

So if we are close to the 120 char limit, we should move it to the body.

test/fixtures/inventory_level.json
```diff
@@ -0,0 +1,7 @@
+{
+  "inventory_level" : {
+      "inventory_item_id": 808950810,
+      "location_id": 905684977,
+      "available": 1
+  }
+}
```
celsodantas: nitpick: extra line required here.

test/fixtures/inventory_levels.json
```diff
@@ -0,0 +1,24 @@
...
+      "location_id": 487838322,
+      "available": 9
+    },
+    {
+      "inventory_item_id": 39072856,
+      "location_id": 905684977,
+      "available": 3
+    }
+  ]
+}
```
celsodantas: nitpick: extra line required here.

lib/shopify_api/resources/inventory_item.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+module ShopifyAPI
+  class InventoryItem < Base
+  end
+end
```
vnbrs: To follow the same pattern other .rb files are following, I'd consider adding an extra line over here.

lib/shopify_api/resources/inventory_level.rb
```diff
@@ -0,0 +1,56 @@
...
+        location_id: location_id,
+        inventory_item_id: inventory_item_id,
+        available_adjustment: available_adjustment
+      }
+      load_attributes_from_response(
+        self.class.post(:adjust, {}, body.to_json)
+      )
+    end
+  end
+end
```
vnbrs: And another new line here

lib/shopify_api/resources/inventory_item.rb
jtgrenz: Most of these actually have new lines at the end. Github for some reason truncates the last line of the file if its blank (which is really weird?). I did forget the new line on another file though and github displays those with a little red warning icon 

[alt](https://screenshot.click/07-03-18-11-07-77evs-vi1y9.jpg)

lib/shopify_api/resources/inventory_level.rb
jtgrenz: can't _quite_ get it to 120 
[alt](https://screenshot.click/07-03-18-11-11-i206l-kxo9i.jpg

next best thing? 
[alt](https://screenshot.click/07-03-18-11-12-rcpxg-b1slg.jpg)

@celsodantas `#destroy` will take params or body, but its difficult to send a body with DELETE in ActiveResource. I figure if they both work, stick with query params here and if someone wants to hit the endpoint with cURL or manually, then that's fine too. We can only ever send one location and one item at a time anyway, so its not like the query string will ever be unreasonably long.

jtgrenz:
@celsodantas @jordanliddle  does it make sense to add convenience methods to location and inventory_item for connect?

```ruby
 class InventoryItem < Base

    def connect(location_id, relocate_if_necessary: nil)
      inventory_level = ShopifyAPI::InventoryLevel.new(inventory_item_id: id, location_id: location_id )
      inventory_level.connect(relocate_if_necessary)
    end
  end
```

so if you are already working with an inventory_item object, you can use use `inventory_item.connect(123456789, relocate_if_necessary: true)` or conversely with a location, 
`location.connect(123456789, relocate_if_necessary: true)`  or is that just unencesary? I only ask because I've already had to use the a similar method to get inventory_levels working on location and these classes are already coupled.

PR #420, opened 2018-03-02
examples added

ayazahmadtarar:
Examples added for Shopify

### DIFF
--- a/examples/shopify.rb
+++ b/examples/shopify.rb
@@ -1,0 +1,71 @@
+class Shopify
+  attr_reader :client
+
+  def initialize(account)
+    @client = account
+  end
+
+  def shop
+    ShopifyAPI::Session.temp(@client.uid, @client.access_token){ ShopifyAPI::Shop.current }
+  end
+  
+  def revoke_shop
+    ShopifyAPI::Base.clear_session
+  end
+
+  def fetch_all_orders
+    ShopifyAPI::Session.temp(@client.uid, @client.access_token) { ShopifyAPI::Order.find(:all) }
+  end
+
+  def fetch_specific_order(order_id)
+    ShopifyAPI::Session.temp(@client.uid, @client.access_token) { ShopifyAPI::Order.find(order_id) }
+  end
+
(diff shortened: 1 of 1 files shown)


Hamza-Irfan:
Its really helpful Ayaz bro. Nice work done. (y)

examples/shopify.rb
```diff
@@ -0,0 +1,71 @@
...
+    
+    fulfill = ShopifyAPI::Fulfillment.new( :params => {:order_id => order_id})
+    fulfill.tracking_numbers = tracking_numbers
+    fulfill.tracking_company = shipping_company
+    fulfill.tracking_urls = tracking_links
+    fulfill.prefix_options = { :order_id => order_id }
+    fulfill.save
+    fulfill.id.present ? fulfill.id : ''
+
+  rescue Exception => e
```
vnbrs: `Exception` class should never be rescued. Instead, you could use `StandardError`. This is going to swallow even `NameError` or `NoMethodError` inside the specs, not the code.

examples/shopify.rb
```diff
@@ -0,0 +1,71 @@
+class Shopify
+  attr_reader :client
+
+  def initialize(account)
+    @client = account
+  end
+
+  def shop
+    ShopifyAPI::Session.temp(@client.uid, @client.access_token){ ShopifyAPI::Shop.current }
```
drewish: Will the Shop be valid outside the temp call? It seems like this should accept a block and run the result in that context instead.

examples/shopify.rb
ayazahmadtarar: alright I will refactor it

examples/shopify.rb
ayazahmadtarar: @drewish 

not it will not for that you have to create a session and thuen create obj and use them like 
`session = ShopifyAPI::Session.new(@client.uid, @client.access_token)
ShopifyAPI::Base.activate_session(session)

shop = ShopifyAPI::Shop.current`

andyw8:
Closing as stale.

Issue #439, opened 2018-06-06
CONTRIBUTING.md is missing the development environment setup


vnbrs:
I didn't find any documentation about how to run shopify_api specs, install the development dependencies, or, most importantly, how to really setup the development environment.

This is going to allow more PRs to be created, since this is an open source gem. :smile:

PR #441, opened 2018-06-06
Add more details to "Using Development Version"

vnbrs:
Referring to #439.

Running `rake test` were failing. To overcome that, I had to run `bundle exec rake`, just as it is on the CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,7 +232,8 @@ If you were previously using Shopify's [activeresource fork](https://github.com/
 Download the source code and run:
 
 ```bash
-rake install
+bundle install --jobs=3 --retry=3
+bundle exec rake
 ```
 
 ## Additional Resources
--- a/README.md
+++ b/README.md
@@ -232,8 +232,8 @@ If you were previously using Shopify's [activeresource fork](https://github.com/
 Download the source code and run:
 
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -232,7 +232,8 @@ If you were previously using Shopify's [activeresource fork](https://github.com/
 Download the source code and run:
 
 ```bash
-rake install
+bundle install --jobs=3 --retry=3
```
jamiemtdwyer: These arguments are only needed when running in the CI environment to make sure that the build doesn't die if there's an issue reaching Rubygems.

`--jobs=3` is used for parallelization and mileage might vary on different workstations. Best to omit it as to not cause confusion.

README.md
```diff
@@ -232,7 +232,8 @@ If you were previously using Shopify's [activeresource fork](https://github.com/
...
+bundle exec rake
```
jamiemtdwyer: For completeness, I think this should be `bundle exec rake test`.

vnbrs:
There you go, @jamiemtdwyer. Thank you for pointing it out!

jamiemtdwyer:
Nice! Thanks a lot @vnbrs.

Issue #438, opened 2018-05-30
Uninstall app from a shop using the API


mksm:
https://help.shopify.com/api/getting-started/authentication/oauth/uninstalling-applications

Could this be supported?

jamiemtdwyer:
Hey @mksm, I just added this in the linked PR.

If you use bundler, you can follow [this guide](https://bundler.io/guides/git.html) to pull the changes into your application. Otherwise, you'll need to wait for the next stable release.

PR #448, opened 2018-06-26
Add support for ApiPermission#destroy

jamiemtdwyer:
Fixes https://github.com/Shopify/shopify_api/issues/438.

Adds support for ApiPermission#destroy, which allows applications to remotely uninstall themselves.

~Removes the OAuth class as it's now deprecated.~

### DIFF
--- a/lib/shopify_api/resources/o_auth.rb
+++ b/lib/shopify_api/resources/o_auth.rb
@@ -1,8 +1,16 @@
+# frozen_string_literal: true
+
+# This resource is deprecated and will be removed in a future version of this gem.
+# Use ShopifyAPI::ApiPermission.destroy instead
+
 module ShopifyAPI
   class OAuth < Base
     self.collection_name = 'oauth'
 
     def self.revoke
+      warn '[DEPRECATED] ShopifyAPI::OAuth#revoke is deprecated and will be removed in a future version. ' \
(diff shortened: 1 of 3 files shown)


jamiemtdwyer:
Actually, the OAuth#revoke action still seems to be working, so I don't think we want to 🔥it here. Maybe we can just add a comment to let folks know that it is deprecated in favour of ApiPermission#revoke.

lib/shopify_api/resources/o_auth.rb
```diff
@@ -1,3 +1,8 @@
+# frozen_string_literal: true
+
+# This resource is deprecated and will likely be removed in a future version of this gem.
+# Use ShopifyAPI::ApiPermission.destroy instead
```
jtgrenz: You probably want to log this when someone uses the deprecated method.

```ruby
def self.revoke
  warn "[DEPRICATION] ShopifyAPI::OAuth is deprecated and will likely be removed in a future version of this gem. Use ShopifyAPI::ApiPermission.destroy instead"
  
  delete(:revoke)
end
```
maybe?

jamiemtdwyer:
Added the kernel warning as @jtgrenz suggested. I've also changed the class method on ApiPermission from `self.revoke` to `self.destroy` to be more descriptive.

PR #449, opened 2018-06-26
Update .rubocop.yml

jamiemtdwyer:
Disable `Style/Documentation` cops

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,3 +6,6 @@ AllCops:
 
 Rails:
   Enabled: false
+
+Style/Documentation:
+  Enabled: false


jtgrenz:
Im a little confused. Shouldn't that have been covered by the Shopify cops we added?

jamiemtdwyer:
You're right, all cops are disabled by default in Shopify's Ruby style guide and this particular cop is never explicitly enabled... not sure what's going on there.

jamiemtdwyer:
Yeah, forget this. Looks like we just need to rebase our branches to pull in the Rubocop config.

PR #460, opened 2018-08-15
[WIP] Force use of checkout api latest version

manumilou:
Clients need to send a specific header version to be routed to the latest version of the Checkout controller. This gem does not allow sending this header, thus locking all clients using it on the legacy version.

This PR allows sending the version header.

Note: this version of code actually forces sending the header, which might cause problem. Well, it's WIP.

### DIFF
--- a/lib/shopify_api/resources/checkout.rb
+++ b/lib/shopify_api/resources/checkout.rb
@@ -1,4 +1,5 @@
 module ShopifyAPI
   class Checkout < Base
+    headers['X-Shopify-Checkout-Version'] = '2017-09-03'
   end
 end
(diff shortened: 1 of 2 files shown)


jamiemtdwyer:
Nice work 👍 

Unfortunately this gem doesn't currently support using the Checkout API without some additional monkey patching.

https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/resources/checkout.rb

For the "Checkout API" proper, the primary key would need to be `token`, this resource will currently only work with the Abandoned Checkouts API as is.

manumilou:
@jamiemtdwyer I should have been more precise. The goal is to make the Checkout resource work with the `Admin::CheckoutsController` controller. The `admin/checkouts` endpoint is now split into 2 controllers: the checkout API and the actual abandoned checkout API (which actually also serves the Checkout API), the branching being done at the routing level based on the version header. I feel you were thinking about the public checkout API ?

PR #462, opened 2018-08-29
Add an abandoned checkout ActiveResource

manumilou:
# TL;DR

The Abandoned Checkout ActiveResource is mapped to the same RESTful resource as the Checkout ActiveResource.

The current Checkout resource is used to serve two different APIs: Abandoned checkouts API and the actual Checkout API. I would like to have one resource per API:

* `ShopifyAPI::Checkout`: Checkout API (*Admin::CheckoutsController* in core)
* `ShopifyAPI::AbandonedCheckout`: Abandoned Checkouts API (*Admin::AbandonedCheckoutsController* in core)

Note that both APIs are still served on the same URL for the moment, but I plan on creating a new endpoint `admin/abandoned_checkouts` routed the same way.

# Migration plan

This should be released as a new version.

It is a breaking change for clients who uses the Abandoned Checkouts API. These clients can either stay on their current version, or update. If they choose the latter, they'll need to replace `ShopifyAPI::Checkout` by the `ShopifyAPI::AbandonedCheckout` resource in their codebase.
Clients already using the Checkout API through the `ShopifyAPI::Checkout` resource will have nothing to do.

Please refer to the unicorn project for more context: https://unicorn.shopify.io/projects/5352

### DIFF
--- a/lib/shopify_api/resources/checkout.rb
+++ b/lib/shopify_api/resources/checkout.rb
@@ -8,5 +8,20 @@ module ShopifyAPI
     def complete
       post(:complete)
     end
+
+    def ready?
+      return false unless persisted?
+
+      reload
+      [200, 201].include?(ShopifyAPI::Base.connection.response.code.to_i)
+    end
+
+    def payments
+      Payment.find(:all, params: { checkout_id: id })
+    end
+
+    def shipping_rates
+      ShippingRate.find(:all, params: { checkout_id: id })
+    end
   end
 end
--- a/lib/shopify_api/resources/checkout.rb
+++ b/lib/shopify_api/resources/checkout.rb
@@ -1,4 +1,8 @@
+# frozen_string_literal: true
+
 module ShopifyAPI
   class Checkout < Base
+    self.primary_key = :token
+    headers['X-Shopify-Checkout-Version'] = '2016-09-06'
   end
 end
--- a/lib/shopify_api/resources/checkout.rb
+++ b/lib/shopify_api/resources/checkout.rb
@@ -4,5 +4,9 @@ module ShopifyAPI
   class Checkout < Base
     self.primary_key = :token
     headers['X-Shopify-Checkout-Version'] = '2016-09-06'
+
+    def complete
+      post(:complete)
+    end
   end
 end
--- a/test/abandoned_checkouts_test.rb
+++ b/test/abandoned_checkouts_test.rb
@@ -5,7 +5,8 @@ require 'test_helper'
 class AbandonedCheckoutsTest < Test::Unit::TestCase
   test ":get lists all abandoned checkouts indexed by id" do
     fake 'checkouts', method: :get, status: 200, body: load_fixture('checkouts')
-    abandoned_checkout = ShopifyAPI::AbandonedCheckout.all
-    assert_equal 450789469, abandoned_checkout.first.id
+    abandoned_checkouts = ShopifyAPI::AbandonedCheckout.all
+    assert_equal 450789469, abandoned_checkouts.first.id
+    assert_equal JSON.parse(load_fixture('checkouts'))['checkouts'].size, abandoned_checkouts.size
   end
 end
--- a/test/abandoned_checkouts_test.rb
+++ b/test/abandoned_checkouts_test.rb
@@ -1,8 +1,10 @@
+# frozen_string_literal: true
+
 require 'test_helper'
 
 class AbandonedCheckoutsTest < Test::Unit::TestCase
-  test "get all should get all orders" do
-    fake 'checkouts', :method => :get, :status => 200, :body => load_fixture('checkouts')
+  test ":get lists all abandoned checkouts indexed by id" do
+    fake 'checkouts', method: :get, status: 200, body: load_fixture('checkouts')
     abandoned_checkout = ShopifyAPI::AbandonedCheckout.all
     assert_equal 450789469, abandoned_checkout.first.id
   end
--- a/test/abandoned_checkouts_test.rb
+++ b/test/abandoned_checkouts_test.rb
@@ -1,0 +1,9 @@
+require 'test_helper'
+
+class AbandonedCheckoutsTest < Test::Unit::TestCase
+  test "get all should get all orders" do
+    fake 'checkouts', :method => :get, :status => 200, :body => load_fixture('checkouts')
+    abandoned_checkout = ShopifyAPI::AbandonedCheckout.all
+    assert_equal 450789469, abandoned_checkout.first.id
+  end
+end
--- a/test/abandoned_checkouts_test.rb
+++ b/test/abandoned_checkouts_test.rb
@@ -1,12 +1,29 @@
 # frozen_string_literal: true
-
 require 'test_helper'
 
 class AbandonedCheckoutsTest < Test::Unit::TestCase
-  test ":get lists all abandoned checkouts indexed by id" do
-    fake 'checkouts', method: :get, status: 200, body: load_fixture('checkouts')
-    abandoned_checkouts = ShopifyAPI::AbandonedCheckout.all
-    assert_equal 450789469, abandoned_checkouts.first.id
-    assert_equal JSON.parse(load_fixture('checkouts'))['checkouts'].size, abandoned_checkouts.size
+  def setup
+    super
+
+    @expected_checkouts = JSON.parse(load_fixture('abandoned_checkouts'))['checkouts']
+    @expected_checkout_id = JSON.parse(load_fixture('abandoned_checkout'))['checkout']['id']
+  end
+
+  test ":create creates a checkout" do
+    fake 'checkouts', method: :post, status: 201, body: load_fixture('abandoned_checkout')
+
+    checkout = ShopifyAPI::AbandonedCheckout.create
+
+    assert_equal @expected_checkout_id, checkout.id
+    assert_equal true, checkout.attributes.include?(:abandoned_checkout_url)
+  end
+
+  test "get all checkouts indexed by token" do
+    fake 'checkouts', method: :get, status: 200, body: load_fixture('abandoned_checkouts')
+
+    checkouts = ShopifyAPI::AbandonedCheckout.all
+
+    assert_equal @expected_checkout_id, checkouts.first.id
+    assert_equal @expected_checkouts.size, checkouts.size
   end
 end
--- a/test/checkouts_test.rb
+++ b/test/checkouts_test.rb
@@ -56,7 +56,7 @@ class CheckoutsTest < Test::Unit::TestCase
 
     fake "checkouts/#{@expected_checkout_id}/payments", method: :get, status: 202, body: load_fixture('payments')
 
-    assert_equal 10.00, checkout.payments.first.attributes['amount']
+    assert_equal 10.00, checkout.payments.first.amount
   end
 
   test ":shipping_rates returns shipping rates for a checkout" do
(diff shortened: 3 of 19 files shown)


manumilou:
Agreed on the version bump. This should be released as `5.0.0` version then.

Now, to your points: 

* Overload the primary_key for the Checkout resource to be :token

I don't know if this is necessary, unless I am missing something here. Many clients using the `Checkout` resource are already using the Checkout API behind the scene if they are requesting one of the `show`, `create` or `update` actions (see this branching logic: https://github.com/Shopify/shopify/blob/master/components/checkouts/web/app/controllers/admin/abandoned_checkouts_controller.rb#L307).

* Add support for non-CRUD methods complete and shipping_rates

Yes, but right now those actions are not even supported by this gem. Only Frenzy uses the `complete` endpoint because they did monkey-patch `ShopifyAPI`: https://logs.shopify.io/en-US/app/search/search?sid=1535568836.4810_0C7CF512-2D40-4AC5-8FDE-E56474880ED3 

* Consider adding some helper methods to deal with remote polling

Yes, it could be done in a subsequent PR. Throttling support also.

manumilou:
This would be pretty much ready for a first-pass review: @Shopify/wholesale @Shopify/frenzy @jamiemtdwyer

Hammadk:
Hi @manumilou. The `ShopifyAPI::Checkout` resource currently returns both regular and abandoned checkouts. Will this functionality be maintained in the future?

I am asking this because the [current flow to create a checkout (link)](https://help.shopify.com/en/api/guides/sell-through-the-checkout-api) with the API is:
1. Create a checkout (an abandoned checkout on Shopify)
2. Update shipping information on checkout
3. Vault card
4. Create payment.

As a client, we can use the [`ShopifyAPI::Checkout` resource to complete](https://github.com/Shopify/frenzy_server/blob/master/app/services/remote_checkout_service.rb#L15-L20) all of these steps. If further restrictions are added, I am concerned that this will further complicate using the rest API since you will have to use different resources

test/checkouts_test.rb
```diff
@@ -1,9 +1,19 @@
 require 'test_helper'
 
 class CheckoutsTest < Test::Unit::TestCase
-  test "get all should get all orders" do
-    fake 'checkouts', :method => :get, :status => 200, :body => load_fixture('checkouts')
+  test "get all checkouts indexed by token" do
+    fake 'checkouts', method: :get, status: 200, body: load_fixture('checkouts')
     checkout = ShopifyAPI::Checkout.all
-    assert_equal 450789469, checkout.first.id
+    assert_equal '2a1ace52255252df566af0faaedfbfa7', checkout.first.id
```
jamiemtdwyer: shouldn't this be `checkout.first.token`? Or is `id` aliased to the primary key? Either way, I think `token` is more explicit

test/shipping_rate_test.rb
```diff
@@ -0,0 +1,17 @@
...
+require 'test_helper'
+
+class ShippingRateTest < Test::Unit::TestCase
+  test ":get lists all shipping rates for a given checkout" do
+    fake 'checkouts', method: :get, status: 200, body: load_fixture('checkouts')
+    checkouts = ShopifyAPI::Checkout.all
+
+    fake "checkouts/#{checkouts.first.id}/shipping_rates",
+      method: :get, status: 200, body: load_fixture('checkouts')
+    shipping_rates = ShopifyAPI::ShippingRate.find(:all, params: { checkout_id: checkouts.first.id })
```
jamiemtdwyer: same thing here with `checkout_id`, shouldn't this be `checkout_token`?

jamiemtdwyer:
LGTM although I have some questions about `token` vs `id` usage for Checkout.

I think we could also add the [Payment](https://help.shopify.com/en/api/reference/sales_channels/payment) resource, it's very simple so it could just be an empty class which inherits from ActiveRecord::Base.

It would be nice to have convenience methods on Checkout for accessing the shipping rates and payments for a Checkout, similar to [this helper method](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/resources/order.rb#L18-L20) on the Order resource for fetching the associated transactions.

Another thing that might be helpful is a `.ready?` instance method for Checkout which returns true if that status code is in `200..201`.

I would be happy to do this work in a follow-up PR or commit directly on top of this. Let me know 👍

test/checkouts_test.rb
manumilou: `ActiveResource` objects use `id` variable as primary key, mapped by default to an `id` attribute of the mapped object. But in this case, I overrode the property to be mapped to `token` attribute: https://github.com/Shopify/shopify_api/pull/462/files#diff-7d86954f571e1b61789ade0a470543b1R5.

I'd prefer keep using `id` when manipulating `ActiveResource` objects, as I think it is more consistent, even though it might be less explicit.

jamiemtdwyer: Ok, makes sense 👍

manumilou:
@jamiemtdwyer Yes, I think it would be nice to add these methods as well in the release. I can work on the `Payment` resource and the helpers methods. Would you be able to implement the `.ready?` method you are talking about?

jamiemtdwyer:
Sure, I'll give it a go. Hopefully it's not too tricky to access the status code from the response in ActiveResource.

test/abandoned_checkouts_test.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class AbandonedCheckoutsTest < Test::Unit::TestCase
+  test ":get lists all abandoned checkouts indexed by id" do
+    fake 'checkouts', method: :get, status: 200, body: load_fixture('checkouts')
+    abandoned_checkout = ShopifyAPI::AbandonedCheckout.all
+    assert_equal 450789469, abandoned_checkout.first.id
```
louisvanier: Since we load a list we should assert that `abandoned_checkouts.size == load_fixture('checkouts').size`

also naming nitpick: `abandoned_checkout` should be pluralized since it is a list of abandoned checkouts

Hammadk:
@manumilou Re: `I would think that, given that you are sending the X-Shopify-Checkout-Version header, you are already creating checkouts through the Checkout API, and not the Abandoned Checkouts API. What makes you think otherwise ?`

I learned this while working on Frenzy. If you don't complete the [fourth 'payment' step](https://help.shopify.com/en/api/guides/sell-through-the-checkout-api), the Checkout shows up in the `Abandoned checkouts` section under orders.

jamiemtdwyer:
^ Shouldn't this be expected? If a customer enters their shipping and billing details but doesn't complete payment, the checkout is abandoned.

manumilou:
> I learned this while working on Frenzy. If you don't complete the [fourth 'payment' step](https://help.shopify.com/en/api/guides/sell-through-the-checkout-api), the Checkout shows up in the `Abandoned checkouts` section under orders.

This step: https://help.shopify.com/en/api/guides/sell-through-the-checkout-api#completing-the-purchase ? An abandoned checkout is a regular checkout that has not yet completed, so this is actually expected. But it does not mean that you are using the Abandoned Checkouts API.

Hammadk:
> This step: https://help.shopify.com/en/api/guides/sell-through-the-checkout-api#completing-the-purchase ? 

Yeah, that step.

> An abandoned checkout is a regular checkout that has not yet completed, so this is actually expected. But it does not mean that you are using the Abandoned Checkouts API.

My point is that by adding the `ShopifyAPI::AbandonedCheckout` we are creating an inconsistency.  If we do this clients will fetch abandonded checkouts with the `ShopifyAPI::AbandonedCheckout` but create them with `ShopifyAPI ::Checkout`. 

We should do this if we have a good reason for sepeating the abandonded checkouts from regular checkouts -- they look and behave like the same resource to me.

dschwar:
> My point is that by adding the ShopifyAPI::AbandonedCheckout we are creating an inconsistency. If we do this clients will fetch abandonded checkouts with the ShopifyAPI::AbandonedCheckout but create them with ShopifyAPI ::Checkout.
>
> We should do this if we have a good reason for sepeating the abandonded checkouts from regular checkouts -- they look and behave like the same resource to me.

I feel like there is an inconsistency regardless of what is done, since the [AbandonedCheckout API#index](https://help.shopify.com/en/api/reference/orders/abandoned_checkouts#index) action returns a Checkout.
Should we remain consistent with the AbandonedCheckout API, meaning that ShopifyAPI::AbandonedCheckout supports the count and list options, but not create/update/etc? Leaving those to ShopifyAPI::Checkout?

Hammadk:
Ok cool. I think we can resolve this by setting up clear expectations about the two resources and their usecases:
* `ShopifyAPI::AbandonedCheckout`: Find and count existing abandoned checkouts (`#index` + `#count` ).
* `ShopifyAPI::Checkout`:  `CRUD` all checkouts. `#index` action excludes abandoned checkouts

manumilou:
@nwtn Can you review this? I'd like to have your take on this. Any advice on the rollout strategy ?

lib/shopify_api/resources/checkout.rb
```diff
@@ -8,5 +8,21 @@ class Checkout < Base
     def complete
       post(:complete)
     end
+
+    def ready?
+      return false unless persisted?
+
+      reload
```
manumilou: Why do you need to reload here?

lib/shopify_api/resources/checkout.rb
```diff
@@ -8,5 +8,21 @@ class Checkout < Base
...
+      status_code = ShopifyAPI::Base.connection.response.code.to_i
+      [200, 201].include?(status_code) ? true : false
```
manumilou: This one-liner would work too:
```
["200", "201"].include?(ShopifyAPI::Base.connection.response.code)
```

lib/shopify_api/resources/checkout.rb
jamiemtdwyer: since the Checkout API must be polled to know if the checkout is ready (e.g. to retrieve shipping rates or before completing payment), we should reload the resource from the server to see what the status returned in the response is.

do you think it's better to have the client explicitly call `reload`?

jamiemtdwyer: Actually, I think we _need_ to reload here since we are reading the status code from the response object. We don't actually store the status code anywhere on the checkout instance.

manumilou:
For some reason, there is one test failure using `activeressource 3.0.0`. Investigating.

tylerball:
> I'm a bit worried about downstream consumers that might be using a floating dependency

Me too, is there any way for us to detect a client using `Checkout` while expecting an `AbandonedCheckout` and explicitly throw and error or something? Otherwise we're a `bundle update` away from someone totally messing up their app.

manumilou:
@tylerball This is a legit concern. But, if you look at this [Splunk query](https://logs.shopify.io/en-US/app/search/search?sid=1537884337.29968_0C7CF512-2D40-4AC5-8FDE-E56474880ED3), you'll see that almost no one is running the latest version. If there were clients out there using floating dependency in their Gemfile, I suppose we would have more clients updated to the latest version.

For more details on impact, I gathered metrics from Splunk on how the gem is used: https://docs.google.com/document/d/1px4EAysjmne7HfFR4fsACBZs03r7pXD79dBOURfY-9g/edit?usp=sharing (Section Annexe, at the end of the doc). I expect the disruptions to be minimal.

manumilou:
We should be able to ship this PR in the coming days. Still waiting for additional feedbacks on the [impact study](https://docs.google.com/document/d/1px4EAysjmne7HfFR4fsACBZs03r7pXD79dBOURfY-9g/edit?usp=sharing) here on Annexe ( cc @jordanliddle )

PR #467, opened 2018-09-19
Add a state param to create_permission_url 

### LINKED ISSUE #466: create_permission_url does not accept state parameter
The readme mentions that we should be providing and checking the `state` parameter:

>   * ``state`` – Optional – A randomly selected value provided by your application, which is unique for each authorization request. During the OAuth callback phase, your application must check that this value matches the one you provided during authorization. [This mechanism is important for the security of your application](https://tools.ietf.org/html/rfc6819#section-3.6).

>    * Ensure the provided ``state`` is the same one that your application provided to Shopify during Step 3.

But surprisingly the `create_permission_url` method does not accept a state parameter.

drewish:
Fixes #466

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -73,9 +73,10 @@ module ShopifyAPI
       self.extra = extra
     end
 
-    def create_permission_url(scope, redirect_uri = nil)
-      params = {:client_id => api_key, :scope => scope.join(',')}
+    def create_permission_url(scope, redirect_uri = nil, state: nil)
+      params = { client_id: api_key, scope: scope.join(',') }
       params[:redirect_uri] = redirect_uri if redirect_uri
+      params[:state] = state if state
       "#{site}/oauth/authorize?#{parameterize(params)}"
     end
 
--- a/test/session_test.rb
+++ b/test/session_test.rb
@@ -74,7 +74,7 @@ class SessionTest < Test::Unit::TestCase
   end
 
   test "create_permission_url returns correct url with single scope no redirect uri" do
-    ShopifyAPI::Session.setup(:api_key => "My_test_key", :secret => "My test secret")
+    ShopifyAPI::Session.setup(api_key: "My_test_key", secret: "My test secret")
     session = ShopifyAPI::Session.new('http://localhost.myshopify.com')
     scope = ["write_products"]
     permission_url = session.create_permission_url(scope)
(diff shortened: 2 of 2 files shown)


lib/shopify_api/session.rb
```diff
@@ -73,9 +73,10 @@ def initialize(url, token = nil, extra = {})
       self.extra = extra
     end
 
-    def create_permission_url(scope, redirect_uri = nil)
-      params = {:client_id => api_key, :scope => scope.join(',')}
+    def create_permission_url(scope, redirect_uri = nil, state: nil)
```
drewish: Moving to named parameters seemed better long term but makes for sort of a wonky interface right now.

lib/shopify_api/session.rb
```diff
@@ -73,9 +73,10 @@ def initialize(url, token = nil, extra = {})
...
+      params = {client_id: api_key, scope: scope.join(',')}
```
drewish: Seems like we're only using ruby 2+ now so we can switch off the old style hashes right?

lib/shopify_api/session.rb
0ptimusrhyme: What would we think about instead consolidating into an `options` hash? CC @jamiemtdwyer 

```rb
def create_permission_url(scope, options = {})
# ...
end

create_permission_url(["read_orders"], state: "foobar", redirect_uri: "baz.myshopify.com")
```

jamiemtdwyer: Consolidating into an options hash isn't a bad idea. I like it because it's future proof, it allows extending this in the future to support generating online access tokens without changing the public API again. That being said, I think I would keep `redirect_uri` and actually remove the defaulting to `nil`, since redirect_uri is required now.

I don't much like the idea of mixing keyword and positional arguments.

drewish: @jamiemtdwyer Just so I'm clear before I try to go get all those tests changed around: You're okay with a breaking to the interface that makes the `redirect_uri` a required parameter? So we'll have a method like:
```rb
    def create_permission_url(scope, redirect_uri, state: nil)
      params = { client_id: api_key, scope: scope.join(','), redirect_uri: redirect_uri }
      params[:state] = state if state
      "#{site}/oauth/authorize?#{parameterize(params)}"
    end
```

drewish: Do I need to make changes to the readme or were you planning on handling that?

0ptimusrhyme: Changes to the README where necessary would be super appreciated as part of this PR

jdjkelly:
@drewish just ran into the lack of this feature - and not using omniauth in my codebase if I don't need to

looks like you need to sign the CLA? since this is so old, I'm happy to open a new pull request inspired by it if you've moved on - lmk

drewish:
Yeah not using this gem any more so I’ll close it and let someone else give it a clean shot.

jdjkelly:
@drewish awesome thanks! I'll pick the idea up

PR #469, opened 2018-09-21
Add Discount Code batch creation job

g-gagnon:
Batch discount code creation job is was not supported in the gem, so added it. Since the paths and json structure are different from the other resources, I'm not certain about my naming convention.

Tested in a project with success.


lib/shopify_api/resources/discount_code_batch.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def price_rule_id
+      @prefix_options[:price_rule_id]
+    end
+
+    def discount_code_job
+      @discount_codes ||= begin
+        if id
+          path = "/admin/price_rules/#{price_rule_id}/batch/#{id}/discount_codes.json"
+          ShopifyAPI::DiscountCode.find :all, from: path
```
g-gagnon: Should this return a discount code object, or a struct would be wiser? https://help.shopify.com/en/api/reference/discounts/discountcode#batch_discount_codes_index

lib/shopify_api/resources/discount_code_batch.rb
```diff
@@ -0,0 +1,24 @@
...
+        end
+      end
+    end
+
+    def encode(options = {})
```
g-gagnon: Need to override this method, since the json format is not standard to the other resource endpoint.
https://help.shopify.com/en/api/reference/discounts/discountcode#batch_create

PR #470, opened 2018-09-22
Rails 5.2.1 session issues fixes

yjaspar:
Fixes to use of ActionParams rather than just hash (undefined method map for ActionController::Parameters).
The `encoded_params_for_signature` called map on ActionController::Parameters which was being an issue. I believe that the original PR was the correct one [#352](https://github.com/Shopify/shopify_api/pull/352)

ie. We can also enforce the use of `to_unsafe_hash` on params in `validate_signature` since the method is supposed to only be called with ActionController::Params but I don't know what it would actually mean in terms of retro-compatibility (prior to strong params) other. The safest way to do it is probably this one.

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -51,6 +51,7 @@ module ShopifyAPI
       end
 
       def validate_signature(params)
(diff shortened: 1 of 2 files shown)


yjaspar:
@rafaelfranca the fix I made this week wasn't working, I had to go back and pick up the previous PR and fix.

PR #472, opened 2018-09-27
Add support for new products endpoint

dominiquesr:
We're introducing a new feature for merchants, to provide the ability to sort a subset of products at the top of the collection as detailed [here](https://help.shopify.com/en/manual/products/collections/automated-collections/auto-sort). 

In support of these changes, we are making modifications to the SmartCollection API. Specifically, the behaviour of the **order** endpoint will change as detailed in the [updated API docs](https://help.shopify.com/en/api/reference/products/smartcollection). Additionally, we are adding a new products endpoint, to provide the ability to retrieve products in the collection, filtered by manually sorted or automatically sorted.

This PR adds support to the Shopify API gem for the new products endpoint. Since ShopifyAPI already has SmartCollection.products defined to so a search for products on collections, this will be a two step process:
1 - Add support for existing SmartCollection.products to call the new products endpoint. Prior to the new manual sort feature being on all shops, we will only use the endpoint when only_sorted is provided as a filter.
2 - Once the new manual sort feature is available on all shops, change products to always use the products endpoint on the controller. This will change the sort order of the products, but I think that will be ok. Also for the test/example we can add the new column for manually sorted to the results to provide a better example.

**WHAT TO FOCUS ON**

- Are there any risks with reusing the products method? Should we give a different name to our new endpoint? We know that the sort order will change from find, to our new endpoint. Could this be a risk to API clients using this gem?
- I still need to decide if I should bump the version now, or only after the entire change. Opinions?

For more context, see https://github.com/Shopify/core/issues/1509

### DIFF
--- a/lib/shopify_api/resources/smart_collection.rb
+++ b/lib/shopify_api/resources/smart_collection.rb
@@ -3,8 +3,12 @@ module ShopifyAPI
     include Events
     include Metafields
 
-    def products
-      Product.find(:all, :params => {:collection_id => self.id})
+    def products(options = {})
+      if options.key?(:only_sorted)
+        Product.find(:all, from: "/admin/smart_collections/#{id}/products.json", params: options)
+      else
+        Product.find(:all, params: { collection_id: id })
+      end
     end
 
     def order(options={})
--- a/test/fixtures/smart_collection_products.json
+++ b/test/fixtures/smart_collection_products.json
@@ -1,0 +1,155 @@
+{
+  "products": [
+    {
+      "product_type": "Cult Products",
+      "handle": "ipod-nano",
+      "created_at": "2011-10-20T14:05:13-04:00",
+      "body_html": "<p>It's the small iPod with one very big idea: Video. Now the world's most popular music player, available in 4GB and 8GB models, lets you enjoy TV shows, movies, video podcasts, and more. The larger, brighter display means amazing picture quality. In six eye-catching colors, iPod nano is stunning all around. And with models starting at just $149, little speaks volumes.</p>",
+      "title": "IPod Nano - 8GB",
+      "template_suffix": null,
+      "updated_at": "2011-10-20T14:05:13-04:00",
+      "id": 632910392,
+      "tags": "Emotive, Flash Memory, MP3, Music",
+      "images": [
+        {
+          "position": 1,
+          "created_at": "2011-10-20T14:05:13-04:00",
+          "product_id": 632910392,
+          "updated_at": "2011-10-20T14:05:13-04:00",
+          "src": "http://static.shopify.com/s/files/1/6909/3384/products/ipod-nano.png?0",
+          "id": 850703190
+        }
+      ],
+      "variants": [
+        {
+          "position": 1,
+          "price": "199.00",
+          "product_id": 632910392,
+          "created_at": "2011-10-20T14:05:13-04:00",
+          "requires_shipping": true,
+          "title": "Pink",
+          "inventory_quantity": 10,
+          "compare_at_price": null,
+          "inventory_policy": "continue",
+          "updated_at": "2011-10-20T14:05:13-04:00",
+          "inventory_management": "shopify",
+          "id": 808950810,
+          "taxable": true,
+          "grams": 200,
+          "sku": "IPOD2008PINK",
+          "option1": "Pink",
+          "fulfillment_service": "manual",
+          "option2": null,
+          "option3": null
+        },
+        {
+          "position": 2,
+          "price": "199.00",
+          "product_id": 632910392,
+          "created_at": "2011-10-20T14:05:13-04:00",
+          "requires_shipping": true,
+          "title": "Red",
+          "inventory_quantity": 20,
+          "compare_at_price": null,
+          "inventory_policy": "continue",
+          "updated_at": "2011-10-20T14:05:13-04:00",
+          "inventory_management": "shopify",
+          "id": 49148385,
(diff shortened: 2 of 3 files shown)


lib/shopify_api/resources/smart_collection.rb
```diff
@@ -3,13 +3,17 @@ class SmartCollection < Base
     include Events
     include Metafields
 
-    def products
-      Product.find(:all, :params => {:collection_id => self.id})
+    def products(options={})
+      if options.has_key?(:only_sorted)
+        get(:products, options).map { |p| Product.new(p) }
```
dominiquesr: There must be a more effective way to do this

test/fixtures/smart_collection_products.json
```diff
@@ -0,0 +1,153 @@
...
+      "vendor": "Apple",
+      "published_at": "2007-12-31T19:00:00-05:00",
+      "options": [
+        {
+          "name": "Title"
+        }
+      ]
+    },
+    {
+      "product_type": "Cult Products",
```
ignacio-chiazzo: are u gonna add `manually_sorted` to the products?

test/smart_collection_test.rb
```diff
@@ -7,4 +7,20 @@ class SmartCollectionTest < Test::Unit::TestCase
...
+
+  test "Smart Collection get products gets all products in a smart collection" do
+    fake "smart_collections/1063001432", :method => :get, :status => 200, :body => load_fixture('smart_collection')
+    smart_collection = ShopifyAPI::SmartCollection.find(1063001432)
+
+    fake "products.json?collection_id=1063001432", :method => :get, :status => 200, :body => load_fixture('smart_collection_products'), :extension => false
+	assert_equal [632910392, 921728736], smart_collection.products.map(&:id)
+  end
+
+  test "Smart Collection get products with only_sorted=only_manual gets only manually sorted products" do
```
ignacio-chiazzo: add a test for `only_automatic`

lib/shopify_api/resources/smart_collection.rb
jamiemtdwyer: `ShopifyAPI::Product.find(:all, from: "/admin/smart_collections/#{id}/products.json)` I think

jamiemtdwyer: but, I think this will cause issues if there are more than 50 products. We may need to parameterize `limit`, `page`, and/or `since_id`

jamiemtdwyer:
Out of curiosity, will this be supported in the [CollectionListings](https://help.shopify.com/en/api/reference/sales_channels/collectionlisting) controller as well?

lib/shopify_api/resources/smart_collection.rb
jamiemtdwyer: `Product.find()` won't use the products controller if you specify the `from:` parameter

dominiquesr: So Jamie, this does go to the heart of the question I raised in the PR description -- do people who use the existing products method in this Gem expect to be able to add all of the same filters that can be added to the Product API when a collection_id is specified? Limit will work but page and offset will not as this endpoint will use a different pagination method. the only way I can think to work around that is to create a different name for this new endpoint, which isn't products. Or to make a really hacky version of our new endpoint that redirects to the profuct controller if we receive params we don't support.  

I suppose we don't ever have to switch the Gem to use the new endpoint for just a straight call to products, but then if they want to receive all products sorted accurately according to the storefront order which has manualyl sorted products first they'll need to use `only_sorted:all` in the params to force it. I suppose its an ok workaround, but not ideal.

@DrewMartin @ignacio-chiazzo thoughts?

dominiquesr: > `Product.find()` won't use the products controller if you specify the `from:` parameter

Right, I should have read your response more closely, thanks!

jamiemtdwyer: > do people who use the existing products method in this Gem expect to be able to add all of the same filters that can be added to the Product API when a collection_id is specified? 

I think it's definitely a reasonable expectation, but either way we'll need to provide some interface in the gem to paginate in the case where there might be > 250 products in a collection.

Hows does pagination work in the new endpoint?

dominiquesr: I'll send you the design doc so you can see. We're also creating a migration page to explain it because its a new concept for the REST API.

test/fixtures/smart_collection_products.json
dominiquesr: I decided to add it here afterall

dominiquesr:
Ready for 👀 again. 
- Loading products more consistently using Product.find
- Added "manually_sorted" to the products for my new fake fixtures
- Addressed 👮

test/smart_collection_test.rb
```diff
@@ -7,4 +7,29 @@ class SmartCollectionTest < Test::Unit::TestCase
     smart_collection = ShopifyAPI::SmartCollection.create(:title => "Macbooks", :rules => rules)
     assert_equal 1063001432, smart_collection.id
   end
+
+  test "Smart Collection get products gets all products in a smart collection" do
+    fake "smart_collections/1063001432", method: :get, status: 200, body: load_fixture('smart_collection')
+    smart_collection = ShopifyAPI::SmartCollection.find(1063001432)
```
ignacio-chiazzo: Could you do `ShopifyAPI::SmartCollection.` instead of faking the endpoint data?

PR #473, opened 2018-10-01
Updated missing entries in CHANGELOG

jamiemtdwyer:
Adds more detail to the CHANGELOG for release 5.0.0

Also, add a changelog entry for v4.13.0 which was never updated.


jamiemtdwyer:
Added your suggestion 👍

PR #476, opened 2018-10-12
Add support for Publication APIs

jamiemtdwyer:
Add support for the Publication, ProductPublication, and CollectionPublication resources.


clement911:
Is that a new API? Different from product listings?

PR #478, opened 2018-10-17
Adds state parameter to create_permission_url

### LINKED ISSUE #466: create_permission_url does not accept state parameter

0ptimusrhyme:
Closes #466 

Title says it all. Because we'd now be looking at multiple optional arguments I decided to convert the single optional redirect_uri argument into a an `options` hash:

```rb
session = ShopifyAPI::Session.new("snowdevil.myshopify.com", "access_token")
scope = ["write_products"]
redirect_uri = "snowdevil.myshopify.com/redirect_uri"
state = "foobar"
session.create_permission_url(scope, :redirect_uri => redirect_uri, :state => state)
```

Tests 🍏

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,10 +98,11 @@ ShopifyAPI uses ActiveResource to communicate with the REST web service. ActiveR
    permission_url = session.create_permission_url(scope)
    ```
 
-   or if you want a custom redirect_uri:
+   or if you want a custom redirect_uri and/or state:
 
    ```ruby
-   permission_url = session.create_permission_url(scope, "https://my_redirect_uri.com")
+   state = SecureRandom.hex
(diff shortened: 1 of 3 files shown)


0ptimusrhyme:
Whoops, dupe. Closing.

PR #479, opened 2018-10-17
Remove support for non-standard protocols and ports.

clayton-shopify:
The Shopify API is only available via HTTPS on port 443. I suspect the `protocol` and `port` options in `ShopifyAPI::Session` were added to support local testing, but even our local development environment now uses HTTPS on port 443. I couldn't find any projects that are setting a custom `protocol` or `port` so I think we can safely remove these options.

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -126,17 +126,18 @@ module ShopifyAPI
     end
 
     private
-      def parameterize(params)
-        URI.escape(params.collect{|k,v| "#{k}=#{v}"}.join('&'))
-      end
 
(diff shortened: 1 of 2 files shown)


manumilou:
I agree on the protocol, we should enforce `https`. But it's usually nice to be able to specify a non-standard port to run an app on a local env, in case the default one is already taken.

clayton-shopify:
> it's usually nice to be able to specify a non-standard port to run an app on a local env

Is it even possible to run the Shopify API on a port other than 443 in local dev?

jamiemtdwyer:
🐑

PR #498, opened 2018-10-25
Log warning when Shopify indicates deprecated API call was performed

chrisbutcher:
Warn via the standard `ActiveResource::Base.logger` when ShopifyAPI makes a deprecated API call.

### DIFF
--- a/lib/active_resource/detailed_log_subscriber.rb
+++ b/lib/active_resource/detailed_log_subscriber.rb
@@ -1,6 +1,17 @@
 module ActiveResource
   class DetailedLogSubscriber < ActiveSupport::LogSubscriber
     def request(event)
+      log_request_response_details(event)
+      warn_on_deprecated_header(event)
+    end
+
+    def logger
+      ActiveResource::Base.logger
+    end
+
+    private
+
+    def log_request_response_details(event)
       data = event.payload[:data]
       headers = data.extract_options!
       request_body = data.first
@@ -10,8 +21,20 @@ module ActiveResource
       info "Response:\n#{event.payload[:response].body}"
     end
 
-    def logger
-      ActiveResource::Base.logger
+    def warn_on_deprecated_header(event)
+      payload = event.payload
+
+      payload[:response].each do |header_name, header_value|
+        case header_name.downcase
+        when 'x-shopify-api-deprecated-reason'
+          warning_message = <<~MSG
+            [DEPRECATED] ShopifyAPI made a call to #{payload[:method].upcase} #{payload[:path]}, and this call made
+            use of a deprecated endpoint, behaviour, or parameter. See #{header_value} for more details.
+          MSG
+
+          warn warning_message
+        end
+      end
     end
   end
 end
--- a/shopify_api.gemspec
+++ b/shopify_api.gemspec
@@ -34,4 +34,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency("minitest", ">= 4.0")
   s.add_development_dependency("rake")
   s.add_development_dependency("timecop")
+  s.add_development_dependency("pry")
+  s.add_development_dependency("pry-byebug")
 end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -2,6 +2,7 @@ require 'rubygems'
 require 'minitest/autorun'
 require 'fakeweb'
 require 'mocha/setup'
+require 'pry'
 
 $LOAD_PATH.unshift(File.dirname(__FILE__))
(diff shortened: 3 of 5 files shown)


lib/active_resource/detailed_log_subscriber.rb
```diff
@@ -10,8 +21,20 @@ def request(event)
...
 
-    def logger
-      ActiveResource::Base.logger
+    def warn_on_deprecated_header(event)
+      payload = event.payload
+
+      payload[:response].each do |header_name, header_value|
+        case header_name.downcase
+        when 'x-shopify-api-deprecated-reason'
+          warning_message = <<~MSG
```
eapache: no need for a variable here?

shopify_api.gemspec
```diff
@@ -34,4 +34,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency("minitest", ">= 4.0")
   s.add_development_dependency("rake")
   s.add_development_dependency("timecop")
+  s.add_development_dependency("pry")
+  s.add_development_dependency("pry-byebug")
+  s.add_development_dependency("byebug")
```
jtgrenz: I used to just add and remove this all the time. I wasn't sure if I should make a separate PR or sneak it into whatever I was working on.

chrisbutcher:
Thanks. Updated!

jtgrenz:
I'm trying to tophat this locally with a copy of Shopify_cli and I'm not getting a depreciation warning  out of the box. The nice thing about the OAuth warning is that if Im just hacking away with the cli, it spits out that warning right away. Anyway we can also log this to stdout all the time as well? Or should we  update the cli?

chrisbutcher:
Ah interesting. I think the `warn` being called by the `OAuth.revoke` is actually Kernel#warn, which probably behaves differently than the `ActiveSupport::LogSubscriber#warn` being called in DetailedLogSubscriber in this PR.

We _could_ warn via both the Kernel and ActiveResource::Base.logger warn methods?

jtgrenz:
I think we should. I just initialized the ActiveResource Logger in the CLI and its just too much detail. Good for a server log, not for a quick test hack.

lib/active_resource/detailed_log_subscriber.rb
```diff
@@ -10,8 +21,18 @@ def request(event)
       info "Response:\n#{event.payload[:response].body}"
     end
 
-    def logger
-      ActiveResource::Base.logger
+    def warn_on_deprecated_header(event)
```
jtgrenz: maybe move this method to lib/shopify_api/connection under  `ShopifyAPI::RequestNotification.warn_on_deprecated_header` and then call `.warn_on_deprecated_header` in `.notify_about_request` 

then call again as part of `ActiveResource::DetailedLogSubscriber.log_request_response_details`?

test/test_helper.rb
```diff
@@ -2,6 +2,7 @@
 require 'minitest/autorun'
 require 'fakeweb'
 require 'mocha/setup'
+require 'pry'
```
jamiemtdwyer: ✂️

chrisbutcher: 😢 I like pry though!

Is this really not needed? Without it, dropping `binding.pry` into code/tests fails with an undefined method error.

jtgrenz: agreed. Might make sense as its own PR mind you. Is there anyway we can autoload it in development so we don't need to require it and remove it? I often like to use it outside the test_helper

jamiemtdwyer: ah, fair enough. didn't realize your intention was to keep the dependency in, I thought this was just leftover from your debugging. I'm fine to keep it 👍

chrisbutcher: @jtgrenz I think if we loaded it in 'development', we'd be requiring it within client code, which we definitely don't want to do. If you're working with shopify_api in a project, that project can require 'pry' in whatever environments it wants.

lib/active_resource/detailed_log_subscriber.rb
chrisbutcher: If I understand correctly, I think that would add coupling via direct method calls, and defeat the purpose of the pub/sub model currently being used, where `ShopifyAPI::Connection#request` ends up notifying subscribers of `request.active_resource_detailed` about the payload (req/res) it encountered.

eapache:
Makes sense to me.

chrisbutcher:
Looks like I need one more review from @Shopify/app-partner-dev-tools-education

chrisbutcher:
Re-ping @Shopify/app-partner-dev-tools-education 🙏 - I just need one more review from your team to ship this, please and thanks!

chrisbutcher:
np and thanks Tyler! <3

PR #499, opened 2018-10-26
Restrict X-Shopify-Checkout-Version header to ShopifyAPI::Checkout

peterjm:
As far as I can tell, all subclasses of `ShopifyAPI::Base` [share the same `headers` object](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/resources/base.rb#L32-L40), unlike `ActiveResource::Base` which [creates a copy in each subclass](https://github.com/rails/activeresource/blob/0ae3010f505b5e6063b7acbcd22cc221995e7873/lib/active_resource/base.rb#L661-L668).

This is intentional, because in ShopifyAPI it's a common pattern to change the access token header to connect to a different shop, and we wouldn't want copies of an access token stored in every subclass.

This means that setting the checkout version header like this ended up setting it for every resource, not just the Checkout resource. Unfortunately, some resources, like the AbandonedCheckout resource, rely on that header not being present.

This PR changes the Checkout resource to inject the header at runtime, rather that store it permanently at class-load time.

### DIFF
--- a/lib/shopify_api/resources/checkout.rb
+++ b/lib/shopify_api/resources/checkout.rb
@@ -3,7 +3,10 @@
 module ShopifyAPI
   class Checkout < Base
     self.primary_key = :token
-    headers['X-Shopify-Checkout-Version'] = '2016-09-06'
+
+    def self.headers
+      super.merge('X-Shopify-Checkout-Version' => '2016-09-06')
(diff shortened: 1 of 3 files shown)


peterjm:
re-ping @manumilou for review

Issue #403, opened 2017-12-21
Discount Code search


diwu86:
Hi, according to Shppify we have discount code search endpoint now . https://help.shopify.com/api/reference/discountcode#lookup.
Can we support this endpoint? This is very useful. 
Currently, in order to search discount code, we have to have a price_rule_id as the prefix. In the discount code search/lookup endpoint, we don't need price_rule_id anymore. I am not sure how to support this using current discount code model.

TheInterwebbers:
+1

Leteyski:
+1

machadolab:
This is the one missing method keeping me from being able to use this gem for all my API needs.

dan003400:
+1

stujo:
Is there a way to call this API at all through the gem?

dan003400:
I created a patch to the gem adding this ability. Add it to the gem by including the following method to an initializer.

```ruby
ShopifyAPI::DiscountCode.class_eval do
  def self.lookup(code)
    find(:one, from: '/admin/discount_codes/lookup.json?code=' + code)
  rescue ActiveResource::Redirection => e
    r = find(:one, from: e.response['Location'])
    unless r.nil?
      r.attributes[:price_rule_id] = %r{price_rules/([0-9]+)}.match(e.response['Location'])[1].to_i
      r.prefix_options[:price_rule_id] = %r{price_rules/([0-9]+)}.match(e.response['Location'])[1].to_i
    end
    r
  end
end
```

ryana:
@dan003400 my hero!

PR #500, opened 2018-10-28
Add discount code lookup

davidlawson:
This is a potential fix for #403, adding a lookup method to DiscountCode.

Usage:
```
code = ShopifyAPI::DiscountCode.lookup('...')
```


jamiemtdwyer:
It's definitely not pretty... but this is probably the only way to solve this in this library.

lib/shopify_api/resources/discount_code.rb
```diff
@@ -5,5 +5,15 @@ class DiscountCode < Base
     def price_rule_id
       @prefix_options[:price_rule_id]
     end
+
+    def self.lookup(code)
+      find(:one, from: '/admin/discount_codes/lookup.json?code=' + code)
+    rescue ActiveResource::Redirection => ex
+      r = find(:one, from: ex.response['Location'])
```
jamiemtdwyer: if we store `ex.response['Location']` to a variable, there will be a little less repetition here

tnastine:
I created a price rule/discount code ("1234") to a development shop and have been trying to get this to work simply by triggering the call on the Home Controller index function. Trying to display the price_rule_id that it finds for the specific code to the Index View.


My API call from Home Controller:

`@price_rule_id = ShopifyAPI::DiscountCode.lookup("1234")`

This is what the view is displaying:

"<ShopifyAPI::DiscountCode:0x007d19686a32f0>"


I'm relatively new to Ruby -- am I missing something?

lib/shopify_api/resources/discount_code.rb
davidlawson: good idea, made the change.

davidlawson:
@tnastine try 
```
c = ShopifyAPI::DiscountCode.lookup("1234")
puts c.attributes
```
to see the attributes, then you can do e.g. `c.id` to get the ID

tnastine:
And of course I figure it out two minutes later. Sorry for the clutter.

I didn't include `.prefix[:price_rule_id]` on my variable in the view. Working now 100%.

nwtn:
Looks like this dropped off everyone’s radar, but it seems like it’s worth adding. @davidlawson are you able to add a test for this method?

Nkadze:
really useful pull, is it planned to merge on master?

dan003400:
Why is this not being merged in?

PR #514, opened 2018-11-21
Add delivery confirmation endpoint to Ping resources.

Tieshia:
Create DeliveryConfirmation resource in Ping to send delivery confirmation details to PingAPI with corresponding tests.


test/ping/message_test.rb
```diff
@@ -0,0 +1,46 @@
...
+      conversation_id: '123',
+    )
+
+    message.save
+    assert_equal "d0c7a2e6-8084-4e79-8483-e4a1352b81f7", message.id
+  end
+
+  def test_confirm
+    # delivery_confirmation_fixture = load_fixture('ping/delivery_confirmation')
+    # delivery_confirmation_details = ActiveSupport::JSON.decode(delivery_confirmation_fixture)
```
pkuruppu: We should remove the comments here

pkuruppu:
Can you tophat this locally against PingAPI?

lib/shopify_api/resources/ping/message.rb
```diff
@@ -1,9 +1,23 @@
 # frozen_string_literal: true
 
+require 'byebug'
```
pkuruppu: Should remove this from the lib

lib/shopify_api/resources/ping/message.rb
```diff
@@ -1,9 +1,23 @@
...
+
 module ShopifyAPI
   module Ping
     class Message < Base
       self.prefix = "/admin/api/ping-api/v1/conversations/:conversation_id/"
+
+      def successful_delivery(options)
+        delivery_details = ShopifyAPI::Ping::DeliveryConfirmation.new(options)
+        delivery_details.save
```
pkuruppu: Does `delivery_details.save` not return a boolean already?

lib/shopify_api/resources/ping/message.rb
```diff
@@ -1,9 +1,23 @@
 # frozen_string_literal: true
 
+require 'byebug'
+
 module ShopifyAPI
   module Ping
     class Message < Base
       self.prefix = "/admin/api/ping-api/v1/conversations/:conversation_id/"
+
+      def successful_delivery(options)
```
ChanChar: how does this work if you can't query for a message?

lib/shopify_api/resources/ping/message.rb
```diff
@@ -1,9 +1,23 @@
...
+        delivery_details = ShopifyAPI::Ping::DeliveryConfirmation.new(options)
+        delivery_details.save
+        "true"
+      end
+
+      def failed_delivery(options)
```
ChanChar: the content of this method seems similar to the above. shouldn't this method build the body for a failed message confirmation and the above method have the body for a successful message?

test/ping/message_test.rb
```diff
@@ -0,0 +1,73 @@
...
+      conversation_id: '123',
+    )
+
+    message.save
+
+    fake "api/ping-api/v1/conversations/123/messages/#{message.id}/delivery_confirmation",
+      method: :post,
+      body: ''
+
+    delivery_confirmation = message.successful_delivery(conversation_id: '123', message_id: message.id)
```
ChanChar: This only works one way. Integration -> Ping since the integration has a reference to the message object but will not work when all you have is the message_id and conversation_id.

lib/shopify_api/resources/ping/message.rb
Tieshia: Looks like it does, just updated it accordingly.

lib/shopify_api/resources/ping/message.rb
Tieshia: While you can't query for a message, explicitly passing in `conversation_id` and `message_id` sets them as prefix values for the `delivery_confirmation` endpoint.

lib/shopify_api/resources/ping/message.rb
Tieshia: Then are we looking to create the entire `delivery_confirmation_details` hash within each of these methods? I thought we were just passing whether the message was successfully delivered, and if not, the string containing the error message. And then move the responsibility of processing the response to the `ping-api` side.

test/ping/message_test.rb
Tieshia: That is correct, will there be use cases where we'll need this to work in the opposite direction since the endpoint is specifically for messages that have been processed by an integration?

Tieshia:
> Can you tophat this locally against PingAPI?

It doesn't look like PingAPI actually uses the shopify_api gem?

lib/shopify_api/resources/ping/conversation.rb
```diff
@@ -13,6 +13,33 @@ def send_message(message_attrs)
...
         message
       end
+
+      def successful_delivery(message_id:, delivery_timestamp:)
+        delivery_details = ShopifyAPI::Ping::DeliveryConfirmation.new(
+          conversation_id: id,
+          message_id: message_id,
+          delivery_confirmation_details: {
+            delivered: true,
+            delivery_timestamp: delivery_timestamp,
```
ChanChar: as discussed, we don't need to change the internal API.

test/ping/conversation_test.rb
```diff
@@ -36,4 +36,33 @@ def test_send_message
...
+
+    conversation = ShopifyAPI::Ping::Conversation.new(id: '123')
+    delivery_confirmation = conversation.failed_delivery(
+      message_id: '111',
+      delivery_timestamp: Time.now.to_s,
+      details: "Integration failed to deliver message."
+    )
+
+    assert_equal("false", delivery_confirmation.delivered)
+  end
```
sealabcore: Should there also be a test that asserts if `message_id`, `details`, and `delivery_timestamp` are set?

test/ping/conversation_test.rb
```diff
@@ -36,4 +36,33 @@ def test_send_message
...
+      method: :post,
+      body: load_fixture('ping/successful_delivery_confirmation'))
+
+    conversation = ShopifyAPI::Ping::Conversation.new(id: '123')
+    delivery_confirmation = conversation.successful_delivery(
+      message_id: '111',
+      delivery_timestamp: Time.now.to_s
+    )
+
+    assert_equal("true", delivery_confirmation.delivered)
```
sealabcore: Should we assert that `message_id` and `delivery_timestamp` are set?

test/ping/conversation_test.rb
Tieshia: `message_id` is actually used as a variable in the endpoint for the delivery confirmation resource

test/ping/conversation_test.rb
Tieshia: `message_id` is actually used as a variable in the endpoint for the delivery confirmation resource

Tieshia:
@Shopify/app-partner-dev-tools-education Can I get a review on this please?

markiiikram:
@tylerball please review. thanks!

PR #516, opened 2018-12-10
Packaging for release v5.2.1

chrisbutcher:
Also ignoring a few local folders that appear when running `bundle install`.


chrisbutcher:
@tylerball Do you mind also reviewing this gem release prep PR?

PR #520, opened 2018-12-20
Add optional currency parameter to Order#capture

jamiemtdwyer:
Adds the optional `currency` argument to Order#capture, as per the [API reference](https://help.shopify.com/en/api/reference/orders/transaction).

Apps that are part of the multi-currency beta need to send this parameter to the Transactions API.

cc @kieranmasterton

### DIFF
--- a/test/order_test.rb
+++ b/test/order_test.rb
@@ -56,4 +56,20 @@ class OrderTest < Test::Unit::TestCase
     order.cancel(email: false, restock: true)
     assert_request_body({'email' => false, 'restock' => true}.to_json)
   end
+
+  test "capture an order with currency param" do
+    fake 'orders/450789469', body: load_fixture('order')
+    order = ShopifyAPI::Order.find(450789469)
+
+    fake 'orders/450789469/transactions', method: :post, status: 201, body: load_fixture('transaction')
+    order.capture(100.00, currency: 'CAD')
+
+    assert_request_body({
+      transaction: {
+        amount: 100.00,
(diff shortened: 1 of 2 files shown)


rebeccajfriedman:
Should we add a test for this?

jamiemtdwyer:
Changed `currency` to a keyword argument and added a test.

test/order_test.rb
```diff
@@ -56,4 +56,12 @@ class OrderTest < Test::Unit::TestCase
     order.cancel(email: false, restock: true)
     assert_request_body({'email' => false, 'restock' => true}.to_json)
   end
+
+  test "capture an order with currency param" do
+    fake 'orders/450789469', body: load_fixture('order')
+    fake 'orders/450789469/transactions', method: :post, status: 201, body: load_fixture('transaction')
+    order = ShopifyAPI::Order.find(450789469)
+    order.capture(100.00, currency: 'CAD')
```
elsom25: nit: can we add newlines to breakup setup/execution/assertions?

i.e.:

```rb
fake('orders/450789469', body: load_fixture('order'))
fake('orders/450789469/transactions', method: :post, status: 201, body: load_fixture('transaction'))
order = ShopifyAPI::Order.find(450789469)

order.capture(100.00, currency: 'CAD')

assert_request_body({ 
  'transaction' => { 
    'amount' => 100.00,
    'kind' => 'capture',
    'currency' => 'CAD'
  } 
}.to_json)
```

Issue #511, opened 2018-11-15
GraphQL client memory usage


flavio-b:
Hello there, 
I have found a potential problem with working with the GraphQL client library. In a simple test in the console, I found each instance of the client takes up to 100mb of memory. 

In case you need to instantiate a new client on every request that needs it, with the specific session of a shop, just 7-10 concurrent requests on a route that instantiates the client would be enough to overwhelm a 1GB server. 

Do you have any tips on working with this scenario? Any inputs on how to handle this type of situation on a typical Rails app would be appreciated! I'd stick with REST API if I could, but barcode/SKU query is only possible with GraphQL at the moment...

Jay-Plumb:
@flavio-b what was the reason for closing this issue? I am observing around ~`60mb` of memory when I initialize the client via `ShopifyAPI::GraphQL.new` 

> I am monkey patching `ShopifyAPI::GraphQL.new` as I have the shopify graphql schema stored locally to help speed up request response times which may explain why my memory usage is ~half to what you have documented above

flavio-b:
@Jay-Plumb , I figured there's probably not much that can be done in the underlying gem (graphql-client). I wasn't able to use it within my Rails app, even while loading the schema from a local file; It was using too much memory, and was tricky to work inside controller actions, with the query needing to be stored in a constant.

I ended up writing a wrapper around [faraday](https://github.com/lostisland/faraday), to use for a few cases. I think graphql-client provides other benefits for more advanced usage, but I didn't need that right now. 

Did you manage another solution?

Jay-Plumb:
@flavio-b In terms of requiring the query to be stored as a constant, it is possible to dynamically assign these during runtime. 

In terms of consuming too much memory, I do not have a good solution for this and I have the same concerns that you outlined initially:

> In case you need to instantiate a new client on every request that needs it, with the specific session of a shop, just 7-10 concurrent requests on a route that instantiates the client would be enough to overwhelm a 1GB server.

Is it possible to re-open this issue? I assume other developers would be facing the same problem.

flavio-b:
No problem, I'm reopening the issue. Hopefully there will be a solution. I still think it might be more feasible for Shopify to create a lighter version of the GraphQL gem.
If you want, I can share the basic wrapper I wrote. It's very crude but works for what I need.

Jay-Plumb:
@flavio-b Thanks - I am curious to see the wrapper that you wrote so please to share :)

flavio-b:
[...]

PR #525, opened 2019-01-28
[WIP] GraphQL client memoization

swalkinshaw:
Background: https://github.com/Shopify/shopify_api/issues/511

This is an attempt at memoizing the `::GraphQL::Client` instance so it's only created once on app boot instead of on every request (this will also avoid an introspection request each time).

@chrisbutcher and I worked on this a while ago. Wanted to open it up before I forgot about it again.

cc @Shopify/api-patterns-team

### DIFF
--- a/lib/shopify_api/resources/graphql.rb
+++ b/lib/shopify_api/resources/graphql.rb
@@ -23,7 +23,7 @@ module ShopifyAPI
         self.class.graphql_client = ::GraphQL::Client.new(schema: schema, execute: @http_client)
       end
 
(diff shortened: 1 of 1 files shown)


nwtn:
@swalkinshaw is this still WIP? Other than a rebase, is there outstanding work to be done, or is it ready for review?

swalkinshaw:
Ideally it has some tests for it. Not sure if @ShayneP is taking this over completely?

swalkinshaw:
See https://github.com/Shopify/shopify_api/pull/672

PR #535, opened 2019-03-11
Packaging for release 6.0.0

clayton-shopify:
Changeset: https://github.com/Shopify/shopify_api/compare/v5.2.4...3de3a6a9fd01d062e0bb6ae662e98e9e9f0e41a0

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+== Version 6.0.0
+
+* Removed undocumented `protocol` and `port` options from `ShopifyAPI::Session`.
+
 == Version 5.2.4
 
 * Added `currency` parameter to `ShopifyAPI::Order#capture`. This parameter is required for apps that belong to the
--- a/lib/shopify_api/version.rb
+++ b/lib/shopify_api/version.rb
@@ -1,3 +1,3 @@
 module ShopifyAPI
-  VERSION = "5.2.4"
+  VERSION = "6.0.0"
 end


nwtn:
At quick glance #479 looks like a breaking change. Should this be a major version instead?

clayton-shopify:
I brought that up with @jamiemtdwyer back when I opened #479 but he didn't think it was worth a major version bump. Although unlikely, there are a couple scenarios where things could potentially break:

* tests that use a simulated version of Shopify
* apps that interface with a proxy server instead of connecting to Shopify directly

I'm not against changing the version number to 6.0.0 to address the small chance of breakage.

nwtn:
OK. Given that this is open/public and our partners could be using it differently than we are internally, I’d prefer to go with 6.0.0.

clayton-shopify:
OK, I've changed it to 6.0.0.

jamiemtdwyer:
I still think that a major version bump is overkill. Consider that partner developers don't have access to a local instance of Shopify, so I'm not sure how they could access the Shopify API on any other port.

I'll defer to @nwtn though.

nwtn:
> I still think that a major version bump is overkill. Consider that partner developers don't have access to a local instance of Shopify, so I'm not sure how they could access the Shopify API on any other port.

It’s the proxying thing that I’m more concerned about. I agree it’s a pretty small chance that it’ll break, but I don’t think there’s any real harm in doing it as a major release, and it at least signals that folks should exercise some caution when adopting.

PR #537, opened 2019-03-18
Add test support for ActiveResource 5.x 

alexaitken:
Updating test suite to support ActiveResource 5.x

Updated our CI run to require passing tests for AR 5.0 and 5.1  anything on master can still fail and not fail the build.

The Countable module correctly handles the difference result structure for custom methods that are in the gem.

The other test failure was only a variation in results in the Shopify threadsafe Active Resource extension.

### DIFF
--- a/lib/shopify_api/resources/marketing_event.rb
+++ b/lib/shopify_api/resources/marketing_event.rb
@@ -1,5 +1,7 @@
 module ShopifyAPI
   class MarketingEvent < Base
+    include Countable
+
     def add_engagements(engagements)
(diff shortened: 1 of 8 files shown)


lib/shopify_api/resources/marketing_event.rb
```diff
@@ -1,5 +1,7 @@
 module ShopifyAPI
   class MarketingEvent < Base
+    include Countable
```
jamiemtdwyer: 🎉

PR #539, opened 2019-03-19
Consolidate slash admin url prefix

alexaitken:
Bring all the definitions of the `/admin/` rest url prefix into 1 place.

This change is breaking for any Resource that has set a prefix.
`self.prefix = '/admin/orders/:order_id/fulfillments/:fulfillment_id/'`
  

The change will be to not include `/admin/` at the beginning of the prefix and switch from `prefix=` to `resource_prefix=` .
Example of the change
`self.prefix = '/admin/orders/:order_id/fulfillments/:fulfillment_id/'` 
`self.resource_prefix = "orders/:order_id/fulfillments/:fulfillment_id/"`

### DIFF
--- a/lib/shopify_api/resources/base.rb
+++ b/lib/shopify_api/resources/base.rb
@@ -4,6 +4,7 @@ module ShopifyAPI
   class Base < ActiveResource::Base
     class InvalidSessionError < StandardError; end
     extend Countable
+
     self.timeout = 90
     self.include_root_in_json = false
     self.headers['User-Agent'] = ["ShopifyAPI/#{ShopifyAPI::VERSION}",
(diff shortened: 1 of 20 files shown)


chrisbutcher:
Should we error out if someone tries to assign a resource prefix a path starting with `/admin`,
i.e. `self.resource_prefix = '/admin/...`?

alexaitken:
Good point, I think it would be better add the error + depreciation notice on a call to prefix= and not add anything to resource_prefix.  This will help people discover the breaking changes.

lib/shopify_api/session.rb
```diff
@@ -132,12 +132,17 @@ def parameterize(params)
...
+      uri = URI.parse(construct_oauth_url('access_token'))
       https = Net::HTTP.new(uri.host, uri.port)
       https.use_ssl = true
       request = Net::HTTP::Post.new(uri.request_uri)
       request.set_form_data('client_id' => api_key, 'client_secret' => secret, 'code' => code)
       https.request(request)
     end
+
+    def construct_oauth_url(path, query_params = {})
+      query_string = "?#{parameterize(query_params)}" unless query_params.empty?
```
tylerball: 👏

lib/shopify_api/resources/base.rb
```diff
@@ -63,12 +66,61 @@ def clear_session
...
+        ''
+      end
+
+      def resource_prefix(_options = {})
+        ''
+      end
+
+      # Sets the \prefix for a resource's nested URL (e.g., <tt>prefix/collectionname/1.json</tt>).
+      # Default value is <tt>site.path</tt>.
+      def resource_prefix=(value)
```
tylerball: does this method need additional testing or is it covered by existing ones?

alexaitken: it is covered by existing tests,  all the urls would be broken with out it.

PR #541, opened 2019-03-19
Update call limit header

swalkinshaw:
`http_x_shopify_shop_api_call_limit` is the legacy header so this updates to use the proper one.

edit: this is *not* a breaking change since we respond with both headers. This is better future compatibility.

### DIFF
--- a/lib/shopify_api/limits.rb
+++ b/lib/shopify_api/limits.rb
@@ -8,13 +8,14 @@ module ShopifyAPI
 
     module ClassMethods
 
-      # Takes form num_requests_executed/max_requests
-      # Eg: 101/3000
+      # Takes form <call count>/<bucket size>
+      # See https://help.shopify.com/en/api/getting-started/api-call-limit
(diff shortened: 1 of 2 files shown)


grcooper:
I don't have a ton of context so just curious:
* Has this been broken for a while, or was it not causing any issues? It seems like https://github.com/Shopify/shopify_api/pull/541/files#diff-b65f00621bd7ff510b316d5345e1eb22L70 would have been broken until this fix goes in?
* Is there a way to 🎩 it/should we?

swalkinshaw:
Nothing's broken, this is just using the old style header.

See https://github.com/Shopify/shopify/blob/373f3a1f82bb10051bc889e992a4b94ddd617922/components/access_and_auth/app/controllers/concerns/authorization_protection.rb#L125-L126

swalkinshaw:
We always send both headers so nothing was broken, but we should be using this newer one. I'll be updating the docs which reference it too.

PR #544, opened 2019-03-20
[BREAKING][WIP] Simplify call limits and update naming

swalkinshaw:
The `Limits` module used to contain code for a "global" limit which has been removed for a while. The code is overly complicated since there's only a single "shop" throttle now. It's also using terminology that's inconsistent with our docs and other communications.

Note that this is a breaking change so we'll have to wait until we want to cut a new major release.

The naming is WIP here so looking for feedback on that but it should be consistent with our docs.

### DIFF
--- a/lib/shopify_api/limits.rb
+++ b/lib/shopify_api/limits.rb
@@ -10,48 +10,39 @@ module ShopifyAPI
       # Takes form <call count>/<bucket size>
       # See https://help.shopify.com/en/api/getting-started/api-call-limit
       # Eg: 2/40
-      CREDIT_LIMIT_HEADER_PARAM = {
-        shop: 'X-Shopify-Shop-Api-Call-Limit',
-      }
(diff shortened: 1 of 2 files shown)


lib/shopify_api/limits.rb
```diff
@@ -10,48 +10,39 @@ module ClassMethods
...
 
       ##
       # How many API calls have I made?
-      # @param {Symbol} scope [:shop]
       # @return {Integer}
       #
-      def credit_used(scope=:shop)
-        api_credit_limit_param(scope).shift.to_i
+      def call_count
+        call_limit_status[0].to_i
```
grcooper: Nit, while re-working this should we coerce `call_limit_status` into an object rather than calling indices?

swalkinshaw: We could... I'm on the fence since it will only ever be two values

PR #545, opened 2019-03-22
Change init_prefix_explicit to use self.resource_prefix.

alexaitken:
Missed 1 use of `prefix=` that should have been migrated to `resource_prefix=` to avoid depreciation warnings.


tiffanytse:
:shipit:

PR #548, opened 2019-03-29
Add api versions for url prefix control

alexaitken:
Add api version to control the url prefix of both the graphql schema and rest resources.

Coercion support from strings and symbols of all the defined api versions.

The version has been added to the ShopifyAPI::Base as a thread safe attribute when that is available.
It is also added to the session and defaulted to `no_version` to match current behaviour.

Session usage looks like:
`ShopifyAPI::Session.new("https://this-is-my-test-shop.myshopify.com", "test_token", :no_version)`
`ShopifyAPI::Session.new("https://this-is-my-test-shop.myshopify.com", "test_token", 'unstable')`


lib/shopify_api/session.rb
```diff
@@ -65,8 +66,9 @@ def encoded_params_for_signature(params)
       end
     end
 
-    def initialize(url, token = nil, extra = {})
+    def initialize(url, token = nil, api_version = :no_version, extra = {})
```
chrisbutcher: Could `api_version` be a kwarg? Same q for `temp`.

lib/shopify_api/api_version.rb
```diff
@@ -0,0 +1,87 @@
...
+      def initialize
+        @version_name = "no_version"
+      end
+
+      def construct_api_path(path)
+        "#{API_PREFIX}#{path}"
+      end
+
+      def construct_graphql_path
+        '/admin/api/graphql.json'
```
chrisbutcher: Re-use `API_PREFIX`?

lib/shopify_api/resources/base.rb
```diff
@@ -57,21 +71,31 @@ def activate_session(session)
...
+        self.api_version = nil
         self.headers.delete('X-Shopify-Access-Token')
       end
 
-      def api_prefix
-        API_PREFIX
+      def api_version
+        if _api_version_defined?
+          _api_version
+        elsif superclass != Object && superclass.site
```
chrisbutcher: What's going on here?

lib/shopify_api/api_version.rb
chrisbutcher: Or just use `construct_api_path`?

lib/shopify_api/session.rb
alexaitken: compatibility is my biggest concern with that.

lib/shopify_api/api_version.rb
alexaitken: i'll move this into a constant,  i don't think sharing the API prefix is worth it in this case.

lib/shopify_api/resources/base.rb
alexaitken: pulled from older versions of active resource and how the to correctly pull the value from the parent class and not have the value changed.
https://github.com/rails/activeresource/blob/v4.0.0/lib/active_resource/base.rb#L418-L439

lib/shopify_api/session.rb
chrisbutcher: But couldn't a kwarg just have a default, too?

alexaitken: could you provide the full signature you are picturing?

chrisbutcher: (See Slack)

chrisbutcher: Converged on "maybe it is better to just go to keyword args", i.e. `def initialize(url:, token:, api_version:, ...)`

chrisbutcher: The method signature will change in the next PR, in which `api_version` becomes required.

PR #549, opened 2019-04-01
Remove default for version on session [Breaking change]

alexaitken:
Version will now be required when creating a session and using the `temp` method. 

At the same time the session methods have bee updated to use keyword arguments.  4 positional arguments is a bit too hard to remember.

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -10,8 +10,9 @@ module ShopifyAPI
     cattr_accessor :api_key, :secret, :myshopify_domain
     self.myshopify_domain = 'myshopify.com'
 
-    attr_accessor :url, :token, :name, :extra
+    attr_accessor :domain, :token, :name, :extra
     attr_reader :api_version
+    alias_method :url, :domain
 
     class << self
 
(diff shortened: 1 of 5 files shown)


lib/shopify_api/session.rb
```diff
@@ -67,7 +67,7 @@ def encoded_params_for_signature(params)
       end
     end
 
-    def initialize(url, token = nil, api_version = :no_version, extra = {})
+    def initialize(url:, token:, api_version:, extra: {})
```
chrisbutcher: Should we take this opportunity to also rename `url` to `domain`, to be consistent with `temp`, but also to choose a name that honours the URL/URI specs? https://doepud.co.uk/blog/anatomy-of-a-url

Another name to consider might be `shop_domain`?

alexaitken: I agree I put it up like this so we could make a call on what this should look like.

alexaitken: alternative would be to call it `site` since that is the name used for this value in ActiveResource.

chrisbutcher: Depends on whether that leaks any of the abstraction?

alexaitken: review everything closer.  `domain` is the best name since we will produce a full url for `site` already.

PR #550, opened 2019-04-02
Add release version and helper methods for switching versions

alexaitken:
## What has been added
Release version type. 

Versions are marked as stable or not stable, Release versions are the only stable versions.

Versions can now be ordered, this allows us to pull the `latest_stable_version` for use in the shopify_app gem config generation.

Helper method for cleanly changing versions.
```
ShopifyAPI::Session.temp(domain: 'shop1.myshopify.com', token: token, api_version: :unstable) do
  Shop.current # calls https://shop1.myshopify.com/admin/api/unstable/shop.json
  ShopifyAPI::Session.with_version(:no_version) do     
    Shop.current # calls https://shop1.myshopify.com/admin/shop.json
  end
end
```

### DIFF
--- a/lib/shopify_api/api_version.rb
+++ b/lib/shopify_api/api_version.rb
@@ -30,6 +30,10 @@ module ShopifyAPI
       define_version(Unstable.new)
     end
 
+    def self.latest_stable_version
+      @versions.values.select(&:stable?).sort.last
+    end
+
     def to_s
       @version_name
     end
@@ -49,6 +53,10 @@ module ShopifyAPI
 
     def <=>(other)
       numeric_version <=> other.numeric_version
+    end
+
+    def stable?
+      false
     end
 
     def construct_api_path(_path)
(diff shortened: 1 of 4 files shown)


lib/shopify_api/api_version.rb
```diff
@@ -84,5 +66,68 @@ def construct_api_path(_path)
...
+
+    class Release < ApiVersion
+      FORMAT = /^\d{4}-\d{2}$/.freeze
+      API_PREFIX = '/admin/api/'
+
+      def initialize(version_number)
+        raise InvalidVersion, version_number unless version_number.match(FORMAT)
+        @version_name = version_number
+        @url = "#{API_PREFIX}#{version_number}/"
+        @numeric_version = version_number.tr('-','').ljust(6, '0').to_i
```
chrisbutcher: legit warning

chrisbutcher: Also re: `version_number.tr('-','').ljust(6, '0').to_i` I think you said this would still function as expected if version handles included days of the month, i.e. `2019-03-15`. I think:
* a) it would be better to avoid jumping levels of abstraction (from version names, urls, to cryptic primitive string, integer manipulation) and put this logic into a well-named private method. and
* b) add a test to verify the `ljust` works as expected, perhaps asserting `2019-03-15` > `2019-03` > `2019-02-15`

PR #554, opened 2019-04-08
Adjust ar and ruby versions supported

alexaitken:
This drops support of ActiveResource versions that are not threadsafe. `< 4.1`
  - version 4.1 is compatible with rails 4.x
  - version 3.2 is part of rails 3.2 and it reached end of life for support in 2016.

Update ruby versions we run against match what versions of ruby have support today.
Run against:
  - 2.4.0
  - 2.5.0
  - 2.6.1

Instead of:
  - 2.3.1   # support has recently ended for ruby 2.3.x
  - 2.4.0-preview1


alexaitken:
I took out the change related to the `headers` method.  I don't think it is needed but I will need to understand why the implementation we have is different than the implementation in `ActiveResource::Base` is.

.travis.yml
```diff
@@ -2,31 +2,22 @@ language: ruby
...
-      gemfile: Gemfile_ar50
-    - rvm: 2.4.0-preview1
-      gemfile: Gemfile_ar51
-    - rvm: 2.4.0-preview1
+    - rvm: 2.4.0
       gemfile: Gemfile_ar_master
-    - rvm: 2.3.1
+    - rvm: 2.5.0
+      gemfile: Gemfile_ar_master
+    - rvm: 2.6.1
```
nwtn: wanna do 2.6.2 instead, since it’s out? https://www.ruby-lang.org/en/news/2019/03/13/ruby-2-6-2-released/

alexaitken: updated to newest version of ruby on each minor version

PR #563, opened 2019-04-10
fix set base api version

JohnnyKei:
# Issue
https://github.com/Shopify/shopify_api#getting-started
When I set like below, the error occurred.
`ShopifyAPI::Base.api_version = '2019-04'`
```
/shopify_api-7.0.0/lib/shopify_api/resources/base.rb:84:in `prefix': undefined method `construct_api_path' for "2019-04":String (NoMethodError)
```

# How
To Create ApiVersion like Session do.

### DIFF
--- a/lib/shopify_api/resources/base.rb
+++ b/lib/shopify_api/resources/base.rb
@@ -76,8 +76,8 @@ module ShopifyAPI
         end
       end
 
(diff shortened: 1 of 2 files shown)


nwtn:
awesome, thank you!

PR #569, opened 2019-04-16
Setting Docker for development

### LINKED ISSUE #565: Dockerize development version
I'd like to work on a Docker/Docker-compose script to make faster and easier to contribute with  Shopify API.

alaxalves:
Fixes #565 
@tylerball There's was no need for a Dockerfile since when don't have many external dependencies aside from those containing in Ruby's docker image. 
To test this out you should have [Docker](https://docs.docker.com/install/) and [Docker-Compose](https://docs.docker.com/compose/install/) installed. I have detailed some stuff in README.md file.


docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.5'
+services:
+  api:
+    container_name: shopify_api
+    image: ruby:2.4.6
```
alaxalves: Is it the correct ruby version?

tiffanytse: Last update: https://github.com/Shopify/shopify_api/commit/57c24ed888e1acff0c0a1aa0057a5a9754cadc7b

Rakefile
```diff
@@ -37,3 +37,10 @@ Rake::RDocTask.new do |rdoc|
   rdoc.rdoc_files.include('README*')
   rdoc.rdoc_files.include('lib/**/*.rb')
 end
+
+task :docker do
+  cmd = "docker-compose up -d && docker exec -i -t shopify_api bash"
+  unless system(cmd, err: File::NULL)
+    abort("Something went wrong, do you have Docker and Docker-Compose installed?")
```
tylerball: I think there could be other reasons for this failing (lack of internet connection, some kind of internal docker error) so let's not swallow the error and just print it. It's less clean but it will help with debugging.

README.md
```diff
@@ -374,6 +374,17 @@ bundle install
 bundle exec rake test
 ```
 
+or if you'd rather use docker just run:
+```bash
+docker run -it --name shopify_api -v $PWD:/shopify_api -w="/shopify_api" ruby:2.4.6 bundle install
```
tylerball: I think its wise to quote the volume arg so it's more portable to different shells: `"$PWD:/shopify_api"`

Rakefile
alaxalves: @tylerball What do you suggest here?

tylerball: If you just use `exec` instead of `system` it will raise any errors up to the user and you don't need to rescue anything

alaxalves: Sweet, I didn't know that. Thnx

Rakefile
```diff
@@ -37,3 +37,8 @@ Rake::RDocTask.new do |rdoc|
   rdoc.rdoc_files.include('README*')
   rdoc.rdoc_files.include('lib/**/*.rb')
 end
+
+task :docker do
+  cmd = "docker-compose up -d && docker exec -i -t shopify_api bash"
+  exec(cmd, err: File::NULL)
```
alaxalves: @tylerball Done as suggested ;)

PR #571, opened 2019-04-17
move version definitions to a separate file

### LINKED ISSUE #570: "You can find released versions here" points to incorrect line
In readme. I actually don't know where it should point to. You'd probably want to have it point to an exact line on an exact commit (not on `master` like now). You'd probably also want to extract the version numbers to a constant where it's easier to find it. I still don't know what API version numbers are supported nor when exactly new API version is released (5pm? 5pm in which timezone? On what day?)

tylerball:
This moves `define_known_versions` to a separate file for easier reference from the README. Closes #570


README.md
```diff
@@ -38,7 +38,7 @@ ShopifyAPI::Session.temp(domain: domain, token: token, api_version: api_version)
 end
 ```
 
-The `api_version` attribute can take the string or symbol name of any known version and correctly coerce it to a `ShopifyAPI::ApiVersion`.  You can find the currently defined versions [here](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/api_version.rb#L58), follow these [instructions](#adding-aditional-api-versions) to add additional version definitions if needed.
+The `api_version` attribute can take the string or symbol name of any known version and correctly coerce it to a `ShopifyAPI::ApiVersion`.  You can find the currently defined versions [here](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/defined_versions.rb), follow these [instructions](#adding-aditional-api-versions) to add additional version definitions if needed.
```
0ptimusrhyme: Should the defined versions link instead point to your commit like this? https://github.com/Shopify/shopify_api/blob/b9f431a9f593c3ef77c16f490d1399a48923526e/lib/shopify_api/defined_versions.rb

re: #570

nwtn: i just noticed there’s a typo in the anchor for `instructions`, which means there’s probably an error in the heading itself (`aditional` --> `additional`). can you fix that now too?

tylerball: I don't think we have to, by moving it to a file that only contains the version definitions we don't need to constantly update the link to a specific commit as the versions change.

tylerball: Done!

nwtn: the link here needs to change now :)

PR #577, opened 2019-05-08
[WIP] Use newly created abandoned checkouts route from core

manumilou:
The `admin/abandoned_checkouts` route is now available on core.
Hence, there is no need to specify the `element_name`, as we can use the default one.

Depends on https://github.com/Shopify/shopify/pull/198172 .

### DIFF
--- a/lib/shopify_api/resources/abandoned_checkout.rb
+++ b/lib/shopify_api/resources/abandoned_checkout.rb
@@ -2,6 +2,5 @@
 
 module ShopifyAPI
   class AbandonedCheckout < Base
-    self.element_name = "checkout"
   end
 end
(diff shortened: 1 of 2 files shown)


nwtn:
@manumilou is this tied to a particular version of the admin API? I noticed in our [docs for abandoned checkouts](https://help.shopify.com/en/api/reference/orders/abandoned-checkouts) the path still shows as `checkouts` rather than `abandoned_checkouts`.

manumilou:
@nwtn I jumped the gun a little bit when creating this PR. The new path is available but not used. I would like the new path to be bundled with the `2019-07` version. We should wait for it before shipping this.

andyw8:
@manumilou are you still interested in shipping this?

PR #593, opened 2019-07-02
Define the upcoming version of July

Edouard-chin:


pstare:
We'll need a gem bump too?

PR #594, opened 2019-07-04
Implements relative cursor pagination.

garethson:
Implements a convenient API for using our new [relative cursor pagination](https://help.shopify.com/en/api/guides/paginated-rest-results), which has been partially released as of API version 2019-07.

Rather than providing an automatic way to iterate through all records in the query, we're adding the following methods to any model collection:
- `next_page?`
- `previous_page?`
- `fetch_next_page`
- `fetch_previous_page`

This allows you to perform the navigation however you see fit. Here is an example of using the API to iterate and process a list of all pending orders:

```ruby
  orders = ShopifyAPI::Order.find(:all, params: { financial_status: 'pending' })
  process_orders(orders)
  while orders.next_page?
    orders = orders.fetch_next_page
    process_orders(orders)
  end
```

Currently making all endpoints available on the unstable API, and we can merge this when all endpoints have relative cursor pagination available in Shopify.

Tested:
```
products = ShopifyAPI::Product.where(limit: 3)
=> [1933186826296, 1933188005944, 1933188563000]

products.previous_page?
=> false

products.next_page?
=> true

products = products.fetch_next_page
=> [1933189120056, 1933200949304, 1970112430136]

products.previous_page?
=> true

products.next_page?
=> true

products = products.fetch_previous_page
=> [1933186826296, 1933188005944, 1933188563000]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,6 +365,36 @@ ActiveResource is threadsafe as of version 4.1 (which works with Rails 4.x and a
 
 If you were previously using Shopify's [activeresource fork](https://github.com/shopify/activeresource) then you should remove it and use ActiveResource 4.1.
 
+## Pagination
+
+Pagination can occur in one of two ways.
+
+Page based pagination
+```ruby
+page = 1
+products = ShopifyAPI::Product.find(:all, params: { limit: 50, page: page })
+process_products(products)
+while(products.count == 50)
+  page += 1
+  products = ShopifyAPI::Product.find(:all, params: { limit: 50, page: page })
+  process_products(products)
+end
+```
+
+Page based pagination will be deprecated in the `2019-10` API version, in favor of the second method of pagination:
+
+[Relative cursor based pagination](https://help.shopify.com/en/api/guides/paginated-rest-results)
+```ruby
+products = ShopifyAPI::Product.find(:all, params: { limit: 50 })
+process_products(products)
+while products.next_page?
+  products = products.fetch_next_page
+  process_products(products)
+end
+```
+
+Relative cursor pagination is currently available for all endpoints using the `unstable` API version.
+
 ## Using Development Version
 
 Download the source code and run:
--- a/lib/active_resource/collection_ext.rb
+++ b/lib/active_resource/collection_ext.rb
@@ -1,0 +1,7 @@
+require 'shopify_api/collection_pagination'
+
+module ActiveResource
+  class Collection
+    prepend ShopifyAPI::CollectionPagination
+  end
+end
--- a/lib/shopify_api/collection_pagination.rb
+++ b/lib/shopify_api/collection_pagination.rb
@@ -1,0 +1,54 @@
+module ShopifyAPI
+  module CollectionPagination
+
+    def initialize(args)
+      @previous_url_params = extract_url_params(pagination_link_headers.previous_link)
+      @next_url_params = extract_url_params(pagination_link_headers.next_link)
+      super(args)
+    end
+
+    def next_page?
+      ensure_available
+      @next_url_params.present?
+    end
+
+    def previous_page?
+      ensure_available
+      @previous_url_params.present?
+    end
+
+    def fetch_next_page
+      fetch_page(@next_url_params)
+    end
+
+    def fetch_previous_page
+      fetch_page(@previous_url_params)
+    end
+
+    private
+
+    AVAILABLE_IN_VERSION = ShopifyAPI::ApiVersion::Unstable.new
+
+    def fetch_page(url_params)
+      ensure_available
+      return [] unless url_params.present?
+
+      resource_class.where(url_params)
+    end
+
+    def extract_url_params(link_header)
+      return nil unless link_header.present?
+      Rack::Utils.parse_nested_query(link_header.url.query)
+    end
+
+    def pagination_link_headers
+      @pagination_link_headers ||= ShopifyAPI::PaginationLinkHeaders.new(
+        ShopifyAPI::Base.connection.response["Link"]
+      )
+    end
+
+    def ensure_available
+      raise NotImplementedError unless ShopifyAPI::Base.api_version >= AVAILABLE_IN_VERSION
+    end
+  end
+end
--- a/lib/shopify_api/pagination_link_headers.rb
+++ b/lib/shopify_api/pagination_link_headers.rb
@@ -1,0 +1,33 @@
+module ShopifyAPI
+  class InvalidPaginationLinksError < StandardError; end
+
+  class PaginationLinkHeaders
+    LinkHeader = Struct.new(:url, :rel)
+    attr_reader :previous_link, :next_link
+
+    def initialize(link_header)
+      links = parse_link_header(link_header)
+      @previous_link = links.find { |link| link.rel == :previous }
+      @next_link = links.find { |link| link.rel == :next }
+
+      self
+    end
+
+    private
+
+    def parse_link_header(link_header)
+      return [] unless link_header.present?
+      links = link_header.split(',')
+      links.map do |link|
+        parts = link.split('; ')
+        raise ShopifyAPI::InvalidPaginationLinksError.new("Invalid link header: url and rel expected") unless parts.length == 2
+
+        url = parts[0][/<(.*)>/, 1]
+        rel = parts[1][/rel="(.*)"/, 1]&.to_sym
+
+        url = URI.parse(url)
+        LinkHeader.new(url, rel)
+      end
+    end
+  end
+end
--- a/test/pagination_test.rb
+++ b/test/pagination_test.rb
@@ -1,0 +1,183 @@
+require 'test_helper'
+
+class PaginationTest < Test::Unit::TestCase
+  def setup
+    super
+
+    @version = ShopifyAPI::ApiVersion::Unstable.new
+    ShopifyAPI::Base.api_version = @version.to_s
+    @next_page_info = "eyJkaXJlY3Rpb24iOiJuZXh0IiwibGFzdF9pZCI6NDQwMDg5NDIzLCJsYXN0X3ZhbHVlIjoiNDQwMDg5NDIzIn0%3D"
+    @previous_page_info = "eyJsYXN0X2lkIjoxMDg4MjgzMDksImxhc3RfdmFsdWUiOiIxMDg4MjgzMDkiLCJkaXJlY3Rpb24iOiJuZXh0In0%3D"
+
+    @next_link_header = "<https://this-is-my-test-shop.myshopify.com/admin/api/unstable/orders.json?page_info=#{@next_page_info}>; rel=\"next\""
+    @previous_link_header = "<https://this-is-my-test-shop.myshopify.com/admin/api/unstable/orders.json?page_info=#{@previous_page_info}>; rel=\"previous\""
+  end
+
+  test "navigates using next and previous link headers with no original params" do
+    link_header ="#{@previous_link_header}, #{@next_link_header}"
+
+    fake 'orders', :method => :get, :status => 200, api_version: @version, :body => load_fixture('orders'), :link => link_header
+    orders = ShopifyAPI::Order.all
+
+    fake(
+      'orders',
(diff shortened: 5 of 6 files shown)


lib/shopify_api/collection_pagination.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    private
+
+    AVAILABLE_IN_VERSION = ShopifyAPI::ApiVersion::Unstable.new
+
+    def fetch_page(page_info)
+      return [] unless page_info
+
+      resource_class.where(original_params.merge(page_info: next_page_info))
```
garethson: We could do `where(page_info: next_page_info)`but we're supporting old versions of ActiveResource, and where [was added only in later versions](https://github.com/rails/activeresource/pull/153).

lib/shopify_api/collection_pagination.rb
```diff
@@ -0,0 +1,52 @@
...
+      next_page_info.present?
+    end
+
+    def previous_page?
+      previous_page_info.present?
+    end
+
+    def fetch_next_page
+      fetch_page(next_page_info)
+    end
```
garethson: Thought about it, but decided to not memoize the next/previous page. If you're calling fetch, just have it fetch.

lib/shopify_api/collection_pagination.rb
```diff
@@ -0,0 +1,52 @@
...
+      raise NotImplementedError unless ShopifyAPI::Base.api_version >= AVAILABLE_IN_VERSION
+
+      return nil unless link_header.present?
+
+      CGI.parse(link_header.url.query)["page_info"][0]
+    end
+
+    def pagination_link_headers
+      @pagination_link_headers ||= ShopifyAPI::PaginationLinkHeaders.new(
+        ShopifyAPI::Base.connection.response["Link"]
```
kmcphillips: I don't think I understand why we get a response off the base connection here.

Does that mean this gem is not threadsafe?

lib/shopify_api/pagination_link_headers.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def parse_link_header(link_header)
+      return [] unless link_header.present?
+      links = link_header.split(',')
+      links.map do |link|
+        parts = link.split('; ')
+        raise ShopifyAPI::InvalidPaginationLinksError.new("Invalid link header: url and rel expected") unless parts.length == 2
+
+        url = parts[0][/<(.*)>/, 1]
+        rel = parts[1][/rel="(.*)"/, 1]&.to_sym
```
kmcphillips: This is fine as long as we don't support below ruby 2.2

lib/shopify_api/pagination_link_headers.rb
```diff
@@ -0,0 +1,33 @@
...
+
+        url = URI.parse(url)
```
kmcphillips: This will raise if `parts[0][/<(.*)>/, 1]` returns nothing. Which, I guess that pattern will always match something?

kmcphillips: Come to think of it, what's that pattern for?

lib/shopify_api/collection_pagination.rb
garethson: ActiveResource appears to be threadsafe, and [sets the connection class to be threadsafe](https://github.com/rails/activeresource/blob/64cedae3019f022a301aad301dc3b7984bb3a153/lib/active_resource/base.rb#L324). 

We have a few places where we reference the base connection for response attributes. [Here](https://github.com/Shopify/shopify_api/blob/6600fae8afa69d29680a3f0413311c6d00724644/lib/shopify_api/resources/checkout.rb#L19) and [here](https://github.com/Shopify/shopify_api/blob/6600fae8afa69d29680a3f0413311c6d00724644/lib/shopify_api/limits.rb#L61).

This was added in AR 4.1, and we only support >= 4.1 

Of note, is that [headers](https://github.com/Shopify/shopify_api/blob/6600fae8afa69d29680a3f0413311c6d00724644/lib/shopify_api/resources/base.rb#L31) is explicitly handled as threadsafe in this gem, but that was done [before the threadsafe stuff was added](https://github.com/Shopify/shopify_api/pull/95). I believe this can be removed, which I'll look into after this.

lib/shopify_api/pagination_link_headers.rb
garethson: `rel` is used down below.

lib/shopify_api/collection_pagination.rb
```diff
@@ -0,0 +1,52 @@
...
+    def next_page_info
+      @next_page_info ||= extract_page_info(pagination_link_headers.next_link)
+    end
+
+    def extract_page_info(link_header)
+      raise NotImplementedError unless ShopifyAPI::Base.api_version >= AVAILABLE_IN_VERSION
+
+      return nil unless link_header.present?
+
+      CGI.parse(link_header.url.query)["page_info"][0]
```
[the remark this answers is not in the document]
garethson: > s.required_ruby_version = ">= 2.4"

DrewMartin:
Should we update the readme?

lib/shopify_api/collection_pagination.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    private
+
+    AVAILABLE_IN_VERSION = ShopifyAPI::ApiVersion::Unstable.new
+
+    def fetch_page(page_info)
+      return [] unless page_info
+
+      resource_class.where(original_params.merge(page_info: page_info))
```
DrewMartin: If your first request had the param `order: 'title asc'` then this would merge page_info to that? Because that would make a bad URL. If we tried to request a page against Shopify with the original params + page_info it will fail if there are filters or sort orders in there.

We're probably better off just parsing out all the params out of the next/previous url and using them directly

garethson: Discussed offline, I'm going to take that approach, parse the next/previous page and use those for params.

test/pagination_test.rb
```diff
@@ -9,14 +9,12 @@ def setup
     @next_page_info = "eyJkaXJlY3Rpb24iOiJuZXh0IiwibGFzdF9pZCI6NDQwMDg5NDIzLCJsYXN0X3ZhbHVlIjoiNDQwMDg5NDIzIn0%3D"
     @previous_page_info = "eyJsYXN0X2lkIjoxMDg4MjgzMDksImxhc3RfdmFsdWUiOiIxMDg4MjgzMDkiLCJkaXJlY3Rpb24iOiJuZXh0In0%3D"
 
-    @next_link_header = "<https://this-is-my-test-shop.myshopify.com/admin/api/unstable/orders.json?page_info=#{@next_page_info}>; rel=\"next\""
-    @previous_link_header = "<https://this-is-my-test-shop.myshopify.com/admin/api/unstable/orders.json?page_info=#{@previous_page_info}>; rel=\"previous\""
+    @next_link_header = "<https://this-is-my-test-shop.myshopify.com/admin/api/unstable/orders.json?page_info=#{@next_page_info}>; rel=\"next\">"
```
DrewMartin: I don't think this is right. The chevrons enclose the URL, not the whole tuple

DrewMartin:
```suggestion
    @next_link_header = "<https://this-is-my-test-shop.myshopify.com/admin/api/unstable/orders.json?page_info=#{@next_page_info}>; rel=\"next\""
```

test/pagination_test.rb
```diff
@@ -0,0 +1,145 @@
...
+  def setup
+    super
+
+    @version = ShopifyAPI::ApiVersion::Unstable.new
+    ShopifyAPI::Base.api_version = @version.to_s
+    @next_page_info = "eyJkaXJlY3Rpb24iOiJuZXh0IiwibGFzdF9pZCI6NDQwMDg5NDIzLCJsYXN0X3ZhbHVlIjoiNDQwMDg5NDIzIn0%3D"
+    @previous_page_info = "eyJsYXN0X2lkIjoxMDg4MjgzMDksImxhc3RfdmFsdWUiOiIxMDg4MjgzMDkiLCJkaXJlY3Rpb24iOiJuZXh0In0%3D"
+
+    @next_link_header = "<https://this-is-my-test-shop.myshopify.com/admin/api/unstable/orders.json?page_info=#{@next_page_info}>; rel=\"next\">"
+    @previous_link_header = "<https://this-is-my-test-shop.myshopify.com/admin/api/unstable/orders.json?page_info=#{@previous_page_info}>; rel=\"previous\">"
```
DrewMartin:
```suggestion
    @previous_link_header = "<https://this-is-my-test-shop.myshopify.com/admin/api/unstable/orders.json?page_info=#{@previous_page_info}>; rel=\"previous\""
```

test/pagination_test.rb
```diff
@@ -0,0 +1,145 @@
...
+      body: load_fixture('orders').gsub("450789469", "1122334455")
+    )
+
+    previous_page = orders.fetch_previous_page
+    assert_equal 1122334455, previous_page.first.id
+  end
+
+  test "uses all passed in querystring parameters" do
+    params = "page_info=#{@next_page_info}&limit=50&fields=#{CGI.escape('id,created_at')}"
+    @next_link_header = "<https://this-is-my-test-shop.myshopify.com/admin/api/unstable/orders.json?#{params}>; rel=\"next\">"
```
DrewMartin:
```suggestion
    @next_link_header = "<https://this-is-my-test-shop.myshopify.com/admin/api/unstable/orders.json?#{params}>; rel=\"next\""
```

test/pagination_test.rb
garethson: @DrewMartin Fixed up for both.

garethson:
I've updated this to, instead of using `original_params`, simply passing in all the params returned via the next/previous link headers.

Tested this [here](https://gist.github.com/garethson/771d219cb13354563a9b8b07255568ed), with products with `order`, as well as products with `fields,limit,order`.

Tested with order (which is not yet converted).

Finally: Tested with events with `filter,limit,order`.

README.md
```diff
@@ -365,6 +365,36 @@ ActiveResource is threadsafe as of version 4.1 (which works with Rails 4.x and a
...
+  page += 1
+  products = ShopifyAPI::Product.find(:all, params: { limit: 50, page: page })
+  process_products(products)
+end
+```
+
+Page based pagination will be deprecated in the `2019-10` API version, in favor of the second method of pagination:
+
+[Relative cursor based pagination](https://help.shopify.com/en/api/guides/paginated-rest-results)
+```
```
DrewMartin:
```suggestion
```ruby
```

lib/shopify_api/pagination_link_headers.rb
ignacio-chiazzo: can you add validations for `url` + `rel`?

lib/active_resource/collection_ext.rb
```diff
@@ -0,0 +1,7 @@
+require 'shopify_api/collection_pagination'
+
+module ActiveResource
+  class Collection
+    prepend ShopifyAPI::CollectionPagination
+  end
+end
```
ignacio-chiazzo: as discussed IRL do we have to solve the problem of querying the next/prev link after doing another query?, e.g.

```
page1 = Produdcts.find(:all)
page2 = Produdcts.find(:all)
page2.fetch_next_page
page1.next_page
page1.previous_page
```

garethson: Updated to parse the headers on collection initialization, vs lazily evaluating them. Also added a test for that case. The collection objects don't have access to the actual response, so we have to rely on `ShopifyAPI::Base.connection.response["Link"]`, which is always for the last made request.

garethson:
@tanema Thanks. Regarding disabling the page param, the API currently raises a fairly descriptive error with a link to the docs if a page param is passed in for a version of the API that isn't supported.

Having a client side error raise would be helpful and would probably catch errors earlier for developers, rather than seeing it in production. I can look into adding this after this is merged.

dominiquesr:
❤️ It's happening! This is going to have a huge impact on adoption!

isaacbowen:
🙏

PR #600, opened 2019-08-08
Add ApiVersion coercion_modes and fetch_known_versions from Shopify

jtgrenz:
This PR  allows us to dynamically define api_versions based on the response from  https://app.shopify.com/services/apis.json

This removes the need to manually updated the gem every time a new version is released or set to end of life. 

When the gem is required, we define a set of known versions from the unauthenticated `services/apis.json` endpoint and everything behaves more or less like it did before. If an unknown version is attempted to be set (ie as of today, `2020-04` or `taco-42`) an UnknownVersion error is raised with all known versions as message.

If for some reason the known_version_set cannot be initialized at start, then we default to assuming the version is valid, and rely on the server to handle the request and return an error if necessary. Developers can retry defining the version set by calling `ShopifyAPI::ApiVersion.define_known_versions` at anytime. Any version that was fetched from the server will have `persisted?` return true, and any on the fly version instances would return false.

ApiVersion attributes were solidified internally at shopify I think after version support was added to the gem, so I also deprecated and aliased methods like `#name` and `#stable?` to `#handle` and `#supported?`


lib/shopify_api/api_version.rb
```diff
@@ -1,113 +1,111 @@
...
+      def coerce_to_version(version_or_handle)
+        return version_or_handle if version_or_handle.is_a?(ApiVersion)
+        if @versions.nil?
+          warn "[API VERSION WARNING] Known API Version set is empty. Initializing unvalidated version from handle."
+          return ApiVersion.new(handle: version_or_handle)
+        end
+
+        @versions[version_or_handle.to_s].tap do |api_version|
+          unless api_version
+            raise ApiVersion::UnknownVersion, "UnknownVersion API version specified, `#{version_or_handle}`. \n Versions available: #{@versions.keys}"
```
eapache: I think this should also just be a warning, since otherwise code specifying `2019-04` will not boot after that version reaches EOL?

lib/shopify_api/api_version.rb
```diff
@@ -1,113 +1,111 @@
...
+    UNSTABLE_AS_DATE = Time.utc(3000, 1, 1)
+    API_PREFIX = '/admin/api/'
+
+    class << self
+      attr_reader :versions
+
+      def coerce_to_version(version_or_handle)
+        return version_or_handle if version_or_handle.is_a?(ApiVersion)
+        if @versions.nil?
+          warn "[API VERSION WARNING] Known API Version set is empty. Initializing unvalidated version from handle."
```
eapache: I think we can come up with a warning message that will fit better in the context of what an app developer cares about?

lib/shopify_api/meta.rb
```diff
@@ -0,0 +1,16 @@
+
+# frozen_string_literal: true
+module ShopifyAPI
+  class Meta < ActiveResource::Base
```
eapache: What is the default ActiveResource timeout? This endpoint is very fast normally, and we probably don't want to delay app startup by more than a second or two before just moving on.

README.md
```diff
@@ -344,19 +344,11 @@ result = client.query(SHOP_NAME_QUERY)
...
-ShopifyAPI::ApiVersion.define_version(ShopifyAPI::ApiVersion::Release.new('2022-03'))
-```
-Once you have done that you can now set this version in a Sesssion like this:
+When shopify_api.rb is loaded, it calls `ShopifyAPI::ApiVersion.define_known_versions` which fetches all known public api versions from https://app.shopify.com/services/apis.json. If for some reason this endpoint is unreachable, you will receive a warning, `[API VERSION WARNING] Could not fetch Admin API versions.`. Whenever you try to set the api version for a request when the know api version set is empty, you will receive another warning, `[API VERSION WARNING] Known API Version set is empty. Initializing unvalidated version from handle.` This means the gem is unable to verify before hand if the version your are attempting to use is valid. A version object will still be instantiated, but it will default its attributes to `{ handle: 'whatever you set', supported: false, latest_supported: false, display_name: 'whatever you set' }`. If the version handle was invalid, the server will return an error, whereas had the known api version set been defined, the gem would prevent you making the call in the first place.
```
eapache: It's a good idea to explain this process in the README, but this feels like a bit too much detail to me?

tanema: Yeah this much detail seems like it is bound to break in the near future.

lib/shopify_api/meta.rb
tanema: Or at least not delay startup without telling the developer what we are waiting on?

lib/shopify_api/api_version.rb
```diff
@@ -1,113 +1,111 @@
...
+        )
+        @versions.values.find(&:latest_supported?)
+      end
     end
 
-    def self.latest_stable_version
-      @versions.values.select(&:stable?).sort.last
+    def initialize(attributes = {}, persisted = false)
+      super
+      @attributes['handle'] = @attributes['handle'].to_s
```
tanema: could handle default to unstable?

lib/shopify_api/api_version.rb
```diff
@@ -1,113 +1,111 @@
...
-    end
+      def clear_defined_versions
+        @versions = nil
+      end
 
-    def self.clear_defined_versions
-      @versions = {}
+      def latest_stable_version
+        warn(
+          '[DEPRECATED] ShopifyAPI::ApiVersion.latest_stable_version is deprecated and will be removed in a future version.'
```
tanema: Is there a recommended way to do this same thing after this method is removed?  Is this just discouraged going forward?

lib/shopify_api/api_version.rb
jtgrenz: well, we talked about changing the endpoint to return EOL versions as well. The current version of the gem does raise an error for unknown versions.

lib/shopify_api/api_version.rb
jtgrenz: I think so. I was drawing a blank though. Any suggestions?

lib/shopify_api/api_version.rb
jtgrenz: I'm not sure we want to be in a situation where the default is unstable. unstable should never be used in production and I cant think of a reason you'd want to call `ApiVersion.new`  vs `ApiVersion.new(handle: :unstable)` if you wanted unstable

README.md
jtgrenz: I unfortunately have inherited my mothers natural ability to turn any simple story or explanation into something closer to war and peace in length ^_^"

lib/shopify_api/api_version.rb
eapache: I'm not sure we want the endpoint to return all EOL versions, since that would make it grow without limit.

lib/shopify_api/api_version.rb
eapache: "Unable to validate API Version"?

lib/shopify_api/api_version.rb
jtgrenz: I think we wanted to discourage anyone trying to set the version to the latest version automatically since it's likely to break apps. Bumping to the latest version should hopefully be intentional. 

It can still be computed however since each ApiVersion has a `latest_supported` attribute, its just a little more work.

`ShopifyAPI::Meta.admin_versions.find(&:latest_supported?)` works at the expense of another api call, but we could also cache the versions in `Meta.admin_versions`

lib/shopify_api/api_version.rb
jtgrenz: true, but at 4 versions a year, it'll take 12.5 years to each 50 versions, which is a pretty long time.

jtgrenz:
updated per @alexaitken 's comments on slack. ApiVersion now supports two coercion modes. The default will just accept whatever handle you give it and leave it to the server to deal with any invalid versions, probably what you would want in production since you should have everything tested already. 

The second stricter mode will not allow you to set any version that is not pre-defined. Setting `ApiVersion.coercion_mode = :predefined_only` will require you to fetch the known versions from Shopify with `ApiVersion.define_known_versions`. Theres no more remote request when requiring the gem and any developers who want to pre-define versions on boot caa do so in an initializer. 

Little more flexibility for different developers.

jtgrenz:
Just pushed an update and rebased against master.

I made `ShopifyAPI::ApiVersion` a PORO so its more light weight like Alex suggested. This now means that `ShopifyAPI::Meta.admin_versions` no longer returns an array of `ShopifyAPI::ApiVersion`s but instead falls back to using ActiveResource magic and returns an array of `ShopifyAPI::Meta::Version`s instead. 

When using `ShopifyAPI::ApiVersion.define_known_versions`, we still grab the array of versions from `Meta` but we then build and cache new lighter weight  `ApiVersion` objects from the fetched `ShopifyAPI::Meta::Version.attributes` hash.

lib/shopify_api/api_version.rb
```diff
@@ -3,111 +3,143 @@ module ShopifyAPI
...
+          else
+            @versions[handle] = ApiVersion.new(handle: handle)
+          end
+        end
       end
-    end
 
-    def self.define_version(version)
-      @versions ||= {}
+      def define_known_versions
```
eapache: I might call this `fetch_known_versions` to be explicit that it involves a remote lookup

lib/shopify_api/api_version.rb
```diff
@@ -3,111 +3,143 @@ module ShopifyAPI
...
+        @versions = Meta.admin_versions.map { |version| [version.handle, ApiVersion.new(version.attributes, true)] }.to_h
+      end
 
-      @versions[version.name] = version
-    end
+      def clear_defined_versions
```
eapache: `clear_known_versions`?

lib/shopify_api/meta.rb
eapache: We should still probably set an explicit timeout here

lib/shopify_api/api_version.rb
```diff
@@ -3,111 +3,143 @@ module ShopifyAPI
...
+          [handle, @versions[handle]]
+        end.compact.to_h
+      end
     end
 
-    def self.latest_stable_version
-      @versions.values.select(&:stable?).sort.last
+    attr_reader :handle, :display_name, :supported, :latest_supported, :verified
+
+    def initialize(attributes = {}, verified = false)
```
eapache: Is there any reason to give these two default values? They're only constructed in two places and I'd think explicit would be better for both.

lib/shopify_api/api_version.rb
jtgrenz: Yeah I think I had that at first and at some point changed it back to minimize the number of changes or something? 🤔 It sorta made sense at the time, but I think you're right that fetch is a better choice.

lib/shopify_api/api_version.rb
jtgrenz: same concern as define/fetch_known_versions

lib/shopify_api/api_version.rb
jtgrenz: ehh, theres no real reason to have attributes defaulted beyond it being a a hold over from converting from active resource. If it defaults to `{}` it will raise an error. Verified I do think should have a default though since it allows us to define a version with `ApiVersion.new('2019-10')` and automatically know its not verified. When we call `define_known_versions` we explicitly set verified to true because its from the server.

lib/shopify_api/api_version.rb
alexaitken: latest_supported_version is used by the ShopifyApp gem https://github.com/Shopify/shopify_app/blob/0900ddcb2f268f6efe3e184739ee83eeab3a0020/lib/generators/shopify_app/install/install_generator.rb#L12 to run the generators.   

It can now use `ShopifyAPI::Meta.admin_versions.find(&:latest_supported?)` to find that info.

lib/shopify_api/meta.rb
```diff
@@ -5,6 +5,7 @@ class Meta < ActiveResource::Base
     self.site = "https://app.shopify.com/services/"
     self.element_name = 'api'
     self.primary_key = :handle
+    self.timeout = 90
```
eapache: This is too high; if the api doesn't respond in a second or so something bigger has gone wrong, and waiting another 89s is likely just wasting time. I'd probably go with 5s to be very conservative.

jtgrenz: sure. I just grabbed this from ShopifyAPI::Base, but I suppose a longer timeout makes sense there given the api. This one endpoint should be pretty zippy always

README.md
```diff
@@ -38,7 +38,15 @@ ShopifyAPI::Session.temp(domain: domain, token: token, api_version: api_version)
 end
 ```
 
-The `api_version` attribute can take the string or symbol name of any known version and correctly coerce it to a `ShopifyAPI::ApiVersion`.  You can find the currently defined versions [here](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/defined_versions.rb), follow these [instructions](#adding-additional-api-versions) to add additional version definitions if needed.
+The `api_version` attribute takes a version handle (ie `'2019-07'` or `:unstable`) sets an instance of `ShopifyAPI::ApiVersion` matching the handle.
```
eapache: missing word?

README.md
```diff
@@ -38,7 +38,15 @@ ShopifyAPI::Session.temp(domain: domain, token: token, api_version: api_version)
...
+By default, any handle will naïvely create a new `ApiVersion` if the version is not in the known versions returned by `ShopifyAPI::ApiVersion.versions`. To ensure only known and active versions can be set, call
+
+```ruby
+ShopifyAPI::ApiVersion.version_lookup_mode = :raise_on_unknown
+ShopifyAPI::ApiVersion.fetch_known_versions
+```
+
+Known versions are fetched and cached from https://app.shopify.com/services/apis.json. Trying to use a version outside this set will raise an error. To switch back to naïve lookup and create a version if its not found, call `ShopifyAPI::ApiVersion.version_lookup_mode = :define_on_unknown` (this is the default mode).
```
eapache: We should maybe mention that this set currently-supported APIs only

jtgrenz:
Just pushed an update that cleans up a few things. I suppose this is a good time to summarize the whole change again.

### New
- Adds `ShopifyAPI::Meta`. This Meta class inherits from `ActiveResource::Base` and not `ShopifyAPI::Base`. It returns read only information about our Apis and their versions. It fetches data from `https://app.shopify.com/services/apis.json` and does not use authentication.
- Adds `apis.json` and `api_versions.json` test fixtures.
- Adds `MetaTestCase` to test that Meta.admin_versions returns an array of admin versions.

### Changed

#### ShopifyAPI::ApiVersion
**Class methods**
- replaces `self.` class methods with `class << self` style methods.
- adds `.versions` method to return hash of known versions in `{ handle: version.attributes }` format.
- adds `.version_lookup_mode` and `.version_lookup_mode=` which changes how we look up versions from the know versions list. Valid options are `:raise_on_unknown` and `:define_on_unknown`
- adds `.find_version` which returns a version from the known versions set or instantiates a new one and adds it to the set (based on `version_lookup_mode`)
- adds `.fetch_known_versions` which looks up version info from server via `ShopifyAPI::Meta` and adds them to the known version set.
- adds `.add_to_known_versions`  which adds a version to the known version set
- adds private `.sanitize_known_versions` which removes any version from the known set that is not verified.
- adds private `.unknown_version_error_message` which returns an error message for an unknown version when version_lookup_mode is :raise_on_unknown.
- Adds `ApiVersion.initialize`. Accepts an attributes hash argument. `handle:` is required, all other attributes default to false.

**Instance Methods**
- redefines `#construct_api_path` and `#construct_graphql_path` to use the version handle in the path.
- adds `#handle_as_date` method to convert a version handle into a date for comparison. If `handle == :unstable`, the date is set to the year 3000 so its always in the future.
- adds `#unstable?` which is true if `handle == 'unstable'`
- adds accessors for all attributes.
- adds `#==` equality method. Returns true if the version handles between two objects match.

#### Misc
 
- deletes `ShopifyAPI::DefinedVersions` 
- removes `ShopifyAPI::ApiVersion.define_known_versions` from `shopify_api.rb`
- updates calls to deprecated methods to use new methods.
- updates readme.

### Deprecated

#### ApiVersion
- `.coerce_to_version` deprecated. use `.find_version`
- `.define_known_versions` deprecated. Use `.fetch_known_versions`
- `.clear_defined_versions` deprecated. Use. `.clear_known_versions`
- `.latest_stable_version` deprecated. Use `ShopifyAPI::Meta.admin_versions.find(&:latest_supported)`
- `#name` deprecated. Use `#handle`
- `#stable?` deprecated. Use `#supported?`.

README.md
jtgrenz: well it includes release candidate as well which is unsupported.

jtgrenz:
sweet. I'll fix these merge conflicts and 🚢

jtgrenz:
fixed merge conflicts and rebased after merging #605. I added a new commit to add more methods to ApiVersion::NullVersion so it raises a more useful error if you try to call any of the new methods added to ApiVersion (like handle, display_name, verified?, etc).

jtgrenz:
Forgot about this on friday. Rebased again and it should be good to merge as soon as CI is done.

@tylerball did we want to release this as version 8.0.0? Is there anything else we need to include in a major version bump?

PR #606, opened 2019-08-15
Coercers with remote version fetch

jtgrenz:
Just to make this a little easier to reason about. I branched off #600 and incorporated #603 to get a better sense of how they would look together. This PR isn't ready to go and I didn't port over any tests yet.  I think I'm still leaning on the side of this might be a little over kill, but I wanted to see it all together anyway.

@alexaitken thoughts?


lib/shopify_api/api_version.rb
```diff
@@ -3,111 +3,143 @@ module ShopifyAPI
...
-      @versions.fetch(version_or_name.to_s) do
-        raise UnknownVersion, "#{version_or_name} is not in the defined version set: #{@versions.keys.join(', ')}"
+      def coercer
+        @coercer ||= VersionCoercers::GenerateVersion.new
       end
-    end
 
-    def self.define_version(version)
-      @versions ||= {}
+      def coercion_mode=(mode)
```
alexaitken: assignment of the coercer replaced the symbols it also allows for extension by implementing any coercer you want.

jtgrenz: ah yes, sorry. That got lost in translation last night. I'm not sure what else someone would want to do though beyond always creating a new version and raising an error if a version is not in the known set. If we had another use case I think I'd be more open to using a coercer class like this. 

Do you have an idea of any future needs that would require this extra flexibility and abstraction from the ApiVersion class?

PR #609, opened 2019-08-27
Add support Collection#products endpoint

siddhantbajaj:
We are introducing a new endpoint which allow merchants to query for products for a particular collection without knowing its type.This PR adds support for this new endpoint in the shopify api gem.

**NOTE:**  We currently have this change behind `unstable`. I will create another PR to replace `unstable` with specific version these changes will go in.
### What should reviewers focus on?

1. Try using the local version of the gem.
2. `ShopifyAPI::Collection.find(2).products` this should return the products of the collection.

Note: This PR is safe to rollback as the new endpoint is behind a version change which will probably be introduced in `2019/10` version.


siddhantbajaj:
Currently blocked by the ^^ PR, but I can use some code reviews till then. :-)

AlexAxeman:
I think you also need to add an api-test. Will this be a new entry in the API docs? It's neither a method in SmartCollections, neither in CustomCollections so you might need to add it.

siddhantbajaj:
Yes, I already have a PR up for it - https://github.com/Shopify/shopify/pull/213687. Just added you for the review there as well.

lib/shopify_api/resources/collection.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module ShopifyAPI
+  class Collection < Base
+    include Events
+    include Metafields
+
+    def products(options = {})
+      Product.find(:all, from: "#{self.class.prefix}collections/#{id}/products.json", params: options)
```
tanema: This should probably be behind it's own version check. There is no need to make a request if we can already tell that they are using the incorrect API version.

siddhantbajaj: 👍

test/collection_test.rb
```diff
@@ -0,0 +1,58 @@
...
+    fake(
+        'products',
+        url: 'https://shop2.myshopify.com/admin/api/2019-07/collections/1/products.json',
+        method: :get,
+        status: 200,
+        body: load_fixture('collection_products'),
+        extension: false
+    )
+
+    assert_raises NotImplementedError do
```
AlexAxeman: Weird.. I would have expected this to be a 404.

test/collection_test.rb
```diff
@@ -0,0 +1,58 @@
...
+        url: 'https://shop2.myshopify.com/admin/api/2019-07/collections/1.json',
+        method: :get,
+        status: 200,
+        body: load_fixture('collection'),
+        extension: false
+    )
+
+    collection = ShopifyAPI::Collection.find(1)
+
+    fake(
```
AlexAxeman: Does this need to be here if the call is going to fail?

siddhantbajaj: Yes, we can remove this. 👍

test/collection_test.rb
siddhantbajaj: Don't really have a strong opinion here, just something we are following in other parts of the project as well, where version is incorrect.

lib/shopify_api/resources/collection.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module ShopifyAPI
+  class Collection < Base
+    include Events
```
ignacio-chiazzo: do we need to include `Events`?

siddhantbajaj: I think yes, we do need to keep support for `Events`. We will also have to create PR in `Shopify/shopify` to add support for `Collection.events`.

PR #622, opened 2019-09-13
[WIP] Adding FulfillmentOrder, FulfillmentRequest, CancellationRequest, AssignedFulfillmentOrder resources/filters

karmakaze:
Issue [#211513](https://github.com/Shopify/shopify/issues/211513) make shopify_api FO compatible

Add `/orders/:order_id/fulfillment_orders` resource

Reviewers please carefully check the `move` endpoint test in particular for form of request query/body and response.

### DIFF
--- a/test/fixtures/fulfillment_order.json
+++ b/test/fixtures/fulfillment_order.json
@@ -1,0 +1,41 @@
+{
+   "fulfillment_order": {
+      "id": 519788021,
+      "shop_id": 690933842,
+      "order_id": 450789469,
+      "fulfillment_service_handle": "shipwire",
+      "status": "open",
+      "requires_shipping": true,
+      "is_not_deleted": true,
+      "deleted_at": null,
+      "created_at": "2016-07-12T11:23:42-04:00",
+      "updated_at": "2016-07-12T11:23:42-04:00",
+      "happened_at": "2016-07-12T11:23:42-04:00",
+      "assigned_location_id": 905684977,
+      "request_status": "unsubmitted",
+      "delivery_category": null,
+      "fulfillment_order_line_items": [
+         {
+            "id": 519788021,
+            "shop_id": 690933842,
+            "fulfillment_order_id": 519788021,
+            "line_item_id": 466157049,
+            "variant_id": 43729076,
+            "inventory_item_id": 808950810,
+            "inventory_commitment_id": 273471403,
+            "quantity": 1,
+            "sku": "draft-151",
+            "vendor": "Snowdevil",
+            "variant_title": "151cm",
+            "product_title": "Draft",
+            "pre_tax_price": 123.45,
+            "currency": "CAD",
+            "properties": null,
+            "is_not_deleted": true,
+            "deleted_at": null,
+            "created_at": "2016-07-12T11:23:54-04:00",
+            "updated_at": "2016-07-12T11:23:54-04:00"
+         }
+      ]
+   }
+}
--- a/test/fulfillment_order_test.rb
+++ b/test/fulfillment_order_test.rb
@@ -1,0 +1,87 @@
+require 'test_helper'
+
+class FulFillmentOrderTest < Test::Unit::TestCase
+  def setup
+    super
+    fake "orders/450789469/fulfillment_orders/519788021", method: :get,
+         body: load_fixture('fulfillment_order')
+  end
+
+  context "FulfillmentOrder" do
+    context "#find" do
+      should "be able to find fulfillment order" do
+        fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021, params: { order_id: 450789469 })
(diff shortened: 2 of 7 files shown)


test/fixtures/fulfillment_order.json
```diff
@@ -0,0 +1,41 @@
...
+      "delivery_category": null,
+      "fulfillment_order_line_items": [
+         {
+            "id": 519788021,
+            "shop_id": 690933842,
+            "fulfillment_order_id": 519788021,
+            "line_item_id": 466157049,
+            "variant_id":  null,
+            "inventory_item_id":  null,
+            "inventory_commitment_id":  null,
```
justinkrol: I think it would make sense for these (`variant_id`, `inventory_item_id`, `inventory_commitment_id`) to be set. In the real system, we shouldn't have `null` values here.

karmakaze: Ok, I'll find and add some values.

test/fixtures/fulfillment_order.json
```diff
@@ -0,0 +1,41 @@
...
+      "delivery_category": null,
+      "fulfillment_order_line_items": [
+         {
+            "id": 519788021,
+            "shop_id": 690933842,
+            "fulfillment_order_id": 519788021,
+            "line_item_id": 466157049,
+            "variant_id": 43729076,
+            "inventory_item_id": 808950810,
+            "inventory_commitment_id": null,
```
justinkrol: Still missing an `inventory_commitment_id` here. Since these fixtures seem to be completely isolated, I think the foreign key id values can be anything.

karmakaze: I thought that `inventory_commitment_id` were nullable and should be filled in when it made sense to do so rather than always having a value to clear out when not needed. Is that not the case?

test/fulfillment_order_test.rb
```diff
@@ -0,0 +1,80 @@
...
+        new_location_id = 5
+
+        moved = ActiveSupport::JSON.decode(load_fixture('fulfillment_order'))
+        moved['fulfillment_order']['assigned_location_id'] = new_location_id
+        fake "orders/450789469/fulfillment_orders/519788021/move", :method => :post, :body => ActiveSupport::JSON.encode(moved)
+
+        assert_equal 'open', fulfillment_order.status
+        assert fulfillment_order.move(new_location_id: new_location_id)
+        assert_equal 'open', fulfillment_order.status
+        assert_equal new_location_id, fulfillment_order.assigned_location_id
```
justinkrol: I think this assertion should actually check that `fulfillment_order` is closed.
When a FO is moved, we close the original and open a new one at the new location.

test/fixtures/fulfillment_order.json
justinkrol: You're right that a fulfillment order line item may not always have a corresponding commitment (like when a variant is deleted, maybe digital products in the future), but that's more of an edge case.

test/fulfillment_order_test.rb
karmakaze: I don't know how to map the response to the resource object that makes the request. The response from fulfillment_orders_controller has three keys
```
{
original_fulfillment_order:
moved_fulfillment_order:
remaining_fulfillment_order:
}
```
What should the state of 'resource` be after `resource.move`?

karmakaze:
Adding `fulfillment_request`/`cancellation_request` actions to `FulfillmentOrder` resource.
Tried adding methods for filtering FOs by `assigned_status`/`location_ids` but it didn't like having its resource path altered from /orders/#/fulfillment_orders to /orders/#/assigned_fulfillment_orders.

test/fulfillment_order_test.rb
```diff
@@ -0,0 +1,218 @@
...
+        )
+
+        assert_equal 1, fulfillment_orders.count
+        fulfillment_order = fulfillment_orders.first
+        assert_equal 519788021, fulfillment_order.id
+        assert_equal 450789469, fulfillment_order.order_id
+      end
+    end
+
+    # context "#with_assigned_status" do
```
caueguerra: what is up with this commented out code?

stevelounsbury:
Hey @karmakaze, are you still working on this one?

test/fulfillment_order_test.rb
karmakaze: I was experimenting with how to organize assigned_fulfillment_orders either as a resource type matching the implemented endpoints or as methods on a fulfillment_orders resource from the client's perspective.

This can be deleted put into an assigned_fulfillment_orders_test.

karmakaze:
@stevelounsbury Sorry I haven't had time to get back to Fulfillment Orders work, just finishing up the last of the Priority-1 Pickup In-store issues.

karmakaze:
This work has been completed on branches #633 #635 (#664) #637 (#665) #639 (#666).

PR #633, opened 2019-11-06
Add fulfillment order related read requests

karmakaze:
This PR adds the following read requests:
- GET orders/:order_id/fulfillment_orders
- GET assigned_fulfillment_orders
- GET fulfillment_orders/:fulfillment_order_id
- GET fulfillment_orders/:fulfillment_order_id/fulfillments

Subsequent PRs will add fulfillment order related write requests to relevant resources.

The approach I took was pretty straight forward with the following notable choices:
- the resources returned via `AssignedFulfillmentOrder` are converted to `FulfillmentOrder` type so that subsequent actions defined on FulfillmentOrder may be used
- the `fulfillments` method on FulfillmentOrder will get the related fulfillments

### DIFF
--- a/lib/shopify_api/resources/assigned_fulfillment_order.rb
+++ b/lib/shopify_api/resources/assigned_fulfillment_order.rb
@@ -1,0 +1,8 @@
+module ShopifyAPI
+  class AssignedFulfillmentOrder < Base
+    def self.all(options = {})
+      assigned_fulfillment_orders = super(options)
+      assigned_fulfillment_orders.map { |afo| FulfillmentOrder.new(afo.attributes) }
+    end
+  end
+end
--- a/lib/shopify_api/resources/assigned_fulfillment_order.rb
+++ b/lib/shopify_api/resources/assigned_fulfillment_order.rb
@@ -1,7 +1,7 @@
 module ShopifyAPI
   class AssignedFulfillmentOrder < Base
-    def self.all(options = {})
-      assigned_fulfillment_orders = super(options)
+    def self.find(scope, *args)
+      assigned_fulfillment_orders = super(scope, *args)
       assigned_fulfillment_orders.map { |afo| FulfillmentOrder.new(afo.attributes) }
     end
   end
--- a/lib/shopify_api/resources/fulfillment_order.rb
+++ b/lib/shopify_api/resources/fulfillment_order.rb
@@ -1,0 +1,16 @@
+module ShopifyAPI
+  class FulfillmentOrder < Base
+    def self.all(options = {})
+      order_id = options.dig(:params, :order_id)
+      raise ShopifyAPI::ValidationException, "'order_id' is required" if order_id.nil? || order_id == ''
+
+      order = ::ShopifyAPI::Order.new(id: order_id)
+      order.fulfillment_orders
+    end
+
+    def fulfillments(options = {})
+      fulfillment_hashes = get(:fulfillments, options)
+      fulfillment_hashes.map { |fulfillment_hash| Fulfillment.new(fulfillment_hash) }
+    end
+  end
+end
--- a/lib/shopify_api/resources/fulfillment_order.rb
+++ b/lib/shopify_api/resources/fulfillment_order.rb
@@ -1,11 +1,15 @@
 module ShopifyAPI
   class FulfillmentOrder < Base
-    def self.all(options = {})
-      order_id = options.dig(:params, :order_id)
-      raise ShopifyAPI::ValidationException, "'order_id' is required" if order_id.nil? || order_id == ''
+    def self.find(scope, *args)
+      if scope == :all
+        order_id = args.first&.dig(:params, :order_id)
+        raise ShopifyAPI::ValidationException, "'order_id' is required" if order_id.blank?
 
-      order = ::ShopifyAPI::Order.new(id: order_id)
-      order.fulfillment_orders
+        order = ::ShopifyAPI::Order.new(id: order_id)
+        order.fulfillment_orders(args.first[:params].except(:order_id))
+      else
+        super(scope, *args)
+      end
     end
 
     def fulfillments(options = {})
--- a/lib/shopify_api/resources/order.rb
+++ b/lib/shopify_api/resources/order.rb
@@ -30,6 +30,11 @@ module ShopifyAPI
       Transaction.create(capture_transaction)
     end
 
+    def fulfillment_orders(options = {})
+      fulfillment_order_hashes = get(:fulfillment_orders, options)
+      fulfillment_order_hashes.map { |fulfillment_order_hash| FulfillmentOrder.new(fulfillment_order_hash) }
+    end
+
     class ClientDetails < Base
     end
   end
--- a/test/assigned_fulfillment_order_test.rb
+++ b/test/assigned_fulfillment_order_test.rb
@@ -1,0 +1,41 @@
+require 'test_helper'
+
+class AssignedFulFillmentOrderTest < Test::Unit::TestCase
+  context "AssignedFulfillmentOrder" do
+    context "#all" do
+      should "list assigned fulfillment orders by assigned_status" do
+        fulfillment_order_fixture = load_fixture('assigned_fulfillment_orders')
+        fake 'assigned_fulfillment_orders.json?assigned_status=cancellation_requested', method: :get,
+             body: fulfillment_order_fixture, extension: false
+
+        assigned_fulfillment_orders = ShopifyAPI::AssignedFulfillmentOrder.all(
+            params: { assigned_status: 'cancellation_requested' }
+        )
+
+        assert_equal 2, assigned_fulfillment_orders.count
+        assigned_fulfillment_orders.each do |fulfillment_order|
+          assert fulfillment_order.is_a?(ShopifyAPI::FulfillmentOrder)
+          assert_equal 'in_progress', fulfillment_order.status
+          assert_equal 'cancellation_accepted', fulfillment_order.request_status
+        end
+      end
+
+      should "be able to list assigned fulfillment orders by location_ids" do
+        fulfillment_order_fixture = load_fixture('assigned_fulfillment_orders')
+        assigned_location_id = 905684977
+        fake "assigned_fulfillment_orders.json?location_ids%5B%5D=#{assigned_location_id}", method: :get,
+             body: fulfillment_order_fixture, extension: false
+
+        assigned_fulfillment_orders = ShopifyAPI::AssignedFulfillmentOrder.all(
+            params: { location_ids: [assigned_location_id] }
+        )
+
+        assert_equal 2, assigned_fulfillment_orders.count
+        assigned_fulfillment_orders.each do |fulfillment_order|
+          assert fulfillment_order.is_a?(ShopifyAPI::FulfillmentOrder)
+          assert_equal assigned_location_id, fulfillment_order.assigned_location_id
+        end
+      end
+    end
+  end
+end
--- a/test/order_test.rb
+++ b/test/order_test.rb
@@ -72,4 +72,18 @@ class OrderTest < Test::Unit::TestCase
       },
     }.to_json)
   end
+
+  test "fulfillment_orders should get fulfillment orders for an order" do
+    fake 'orders/450789469', body: load_fixture('order')
+    order = ShopifyAPI::Order.find(450789469)
+
+    fake 'orders/450789469/fulfillment_orders', method: :get, body: load_fixture('fulfillment_orders')
+    fulfillment_orders = order.fulfillment_orders
+
+    assert_equal [519788021, 519788022], fulfillment_orders.map(&:id).sort
+    fulfillment_orders.each do |fulfillment_order|
+      assert fulfillment_order.is_a?(ShopifyAPI::FulfillmentOrder)
+      assert_equal 450789469, fulfillment_order.order_id
+    end
+  end
 end
--- a/test/fixtures/assigned_fulfillment_orders.json
+++ b/test/fixtures/assigned_fulfillment_orders.json
@@ -1,0 +1,78 @@
+[
+   {
+      "id": 519788021,
+      "order_id": 450789469,
+      "fulfillment_service_handle": "shipwire",
+      "status": "in_progress",
+      "requires_shipping": true,
+      "is_not_deleted": true,
+      "deleted_at": null,
+      "created_at": "2016-07-12T11:23:42-04:00",
+      "updated_at": "2016-07-12T11:23:42-04:00",
+      "happened_at": "2016-07-12T11:23:42-04:00",
+      "assigned_location_id": 905684977,
+      "request_status": "cancellation_accepted",
+      "delivery_category": null,
+      "fulfillment_order_line_items": [
+         {
+            "id": 519788021,
+            "shop_id": 690933842,
+            "fulfillment_order_id": 519788021,
+            "line_item_id": 466157049,
+            "variant_id": 43729076,
+            "inventory_item_id": 808950810,
+            "inventory_commitment_id": 273471403,
+            "quantity": 1,
+            "sku": "IPOD2008GREEN",
+            "vendor": null,
+            "variant_title": "green",
+            "product_title": "IPod Nano - 8gb",
+            "pre_tax_price": 398.00,
+            "currency": "USD",
+            "properties": null,
+            "is_not_deleted": true,
+            "deleted_at": null,
+            "created_at": "2016-07-12T11:23:54-04:00",
+            "updated_at": "2016-07-12T11:23:54-04:00"
+         }
+      ]
+   },
+   {
+      "id": 519788022,
+      "order_id": 450789469,
+      "fulfillment_service_handle": "manual",
+      "status": "in_progress",
+      "requires_shipping": true,
+      "is_not_deleted": true,
+      "deleted_at": null,
+      "created_at": "2016-07-12T11:23:42-04:00",
+      "updated_at": "2016-07-12T11:23:42-04:00",
+      "happened_at": "2016-07-12T11:23:42-04:00",
+      "assigned_location_id": 905684977,
+      "request_status": "cancellation_accepted",
+      "delivery_category": null,
+      "fulfillment_order_line_items": [
+         {
+            "id": 519788021,
+            "shop_id": 690933842,
+            "fulfillment_order_id": 519788021,
+            "line_item_id": 466157049,
+            "variant_id": 43729076,
+            "inventory_item_id": 808950810,
+            "inventory_commitment_id": 273471403,
+            "quantity": 1,
+            "sku": "IPOD2008RED",
+            "vendor": null,
+            "variant_title": "red",
+            "product_title": "IPod Nano - 8gb",
+            "pre_tax_price": 398.00,
+            "currency": "USD",
+            "properties": null,
+            "is_not_deleted": true,
+            "deleted_at": null,
+            "created_at": "2016-07-12T11:23:54-04:00",
+            "updated_at": "2016-07-12T11:23:54-04:00"
+         }
+      ]
+   }
+]
--- a/test/fixtures/fulfillment_order.json
+++ b/test/fixtures/fulfillment_order.json
@@ -1,0 +1,38 @@
+{
+   "id": 519788021,
+   "order_id": 450789469,
+   "fulfillment_service_handle": "shipwire",
+   "status": "open",
+   "requires_shipping": true,
+   "is_not_deleted": true,
+   "deleted_at": null,
+   "created_at": "2016-07-12T11:23:42-04:00",
+   "updated_at": "2016-07-12T11:23:42-04:00",
+   "happened_at": "2016-07-12T11:23:42-04:00",
+   "assigned_location_id": 905684977,
+   "request_status": "unsubmitted",
+   "delivery_category": null,
+   "fulfillment_order_line_items": [
+      {
+         "id": 519788021,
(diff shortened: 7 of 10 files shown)


lib/shopify_api/resources/assigned_fulfillment_order.rb
```diff
@@ -0,0 +1,12 @@
+module ShopifyAPI
+  class AssignedFulfillmentOrder < Base
+    CANCELLATION_REQUESTED = 'cancellation_requested'
+    FULFILLMENT_REQUESTED = 'fulfillment_requested'
+    FULFILLMENT_ACCEPTED = 'fulfillment_accepted'
```
justinkrol: What are these constants for?

lib/shopify_api/resources/assigned_fulfillment_order.rb
```diff
@@ -0,0 +1,12 @@
...
+
+    def self.all(options = {})
+      assigned_fulfillment_orders = super(options)
+      assigned_fulfillment_orders.map { |afo| FulfillmentOrder.new(afo.as_json) }
```
justinkrol: If we do `AssignedFulfillmentOrder < FulfillmentOrder`, is this still necessary?

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
```diff
@@ -0,0 +1,4 @@
+module ShopifyAPI
+  class FulfillmentOrderFulfillment < Base
```
justinkrol: What does this class represent? Can we return `Fulfillment` objects in `FulfillmentOrder#fulfillments` instead?

test/assigned_fulfillment_order_test.rb
```diff
@@ -0,0 +1,41 @@
+require 'test_helper'
+
+class AssignedFulFillmentOrderTest < Test::Unit::TestCase
+  context "AssignedFulfillmentOrder" do
+    context "#all" do
+      should "be able to list assigned fulfillment orders by assigned_status" do
```
justinkrol:
```suggestion
      should "list assigned fulfillment orders by assigned_status" do
```

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
froot: Wondering if this is needed? I'm thinking that the fulfillments returned via `GET fulfillment_orders/:fulfillment_order_id/fulfillments` shouldn't be treated any differently than the existing fulfillment object.

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -0,0 +1,14 @@
+module ShopifyAPI
+  class FulfillmentOrder < Base
+    def self.all(options = {})
```
froot: With the `fulfillment_orders` method implemented on the order object, is this method necessary?

test/assigned_fulfillment_order_test.rb
```diff
@@ -0,0 +1,41 @@
...
+        fo_fixture = load_fixture('assigned_fulfillment_orders')
+        fake 'assigned_fulfillment_orders.json?assigned_status=cancellation_requested', method: :get,
```
justinkrol: I know it may not be feasible to always keep these things in sync, but it's somewhat confusing that the fixtures have `request_status: unsubmitted`, but will still be returned as part of this request, which passes in `assigned_status: 'cancellation_requested'`.

From here, it looks like the FOs returned should only be `in_progress/cancellation_accepted`
https://github.com/Shopify/shopify/blob/4ac84b757476157526b88347af3c970fe9e5cbcb/components/fulfillments/app/models/fulfillments/fulfillment_order_assignment_status.rb#L18-L22

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -0,0 +1,14 @@
...
+      order_id = options.dig(:params, :order_id)
```
justinkrol: Should this raise if `order_id` is not passed in `options`?

test/assigned_fulfillment_order_test.rb
justinkrol: nvm looks like this is the style for tests in this repo

test/order_test.rb
```diff
@@ -72,4 +72,19 @@ class OrderTest < Test::Unit::TestCase
...
   end
+
+  test "fulfillment_orders should get fulfillment orders for an order" do
+    fake 'orders/450789469', body: load_fixture('order')
+    order = ShopifyAPI::Order.find(450789469)
+
+    fake 'orders/450789469/fulfillment_orders', method: :get, body: load_fixture('fulfillment_orders')
+    fulfillment_orders = order.fulfillment_orders
+
+    assert_equal 2, fulfillment_orders.count
```
justinkrol: nit: this isn't necessary with the assertion on the `id`s below

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
karmakaze: It wasn't clear at the time whether it was ok to mix/match existing Fulfillment actions with those newly added, seemed like it might get confusing to remember which are legacy type requests vs new fulfillment order ones.

lib/shopify_api/resources/fulfillment_order.rb
karmakaze: That's right it can be queried on the order object. Any activeresource type will attempt to make queries, i.e. `FulfillmentOrder.all` will attempt to make a request even if this override method isn't included. Adding this just does the expected thing by routing to the order's implementation rather than failing without good diagnostics.

lib/shopify_api/resources/assigned_fulfillment_order.rb
karmakaze: They are the possible `assignment_status` values that can be used in `options` for selecting returned FulfillmentOrders.

lib/shopify_api/resources/assigned_fulfillment_order.rb
karmakaze: Seemed like that should work but it results in `uninitialized constant ShopifyAPI::FulfillmentOrder (NameError)`

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
karmakaze: Good question--being discussed in comment thread immediately below.

lib/shopify_api/resources/fulfillment_order.rb
karmakaze: Absolutely.

test/assigned_fulfillment_order_test.rb
karmakaze: Updated the `assigned_fulfillment_orders` fixture data to be logically consistent.

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
froot: Should we restrict apps from calling legacy requests on fulfillments (https://help.shopify.com/en/api/reference/shipping-and-fulfillment/fulfillment#show-2019-10) associated to fulfillment orders? I don't think we should since non opted-in FS should be able to query for fulfillments from a fulfillment order and still run legacy requests.

karmakaze: There are a few conflicting overlaps that have to get resolved if we collapse them into a single `Fulfullment` resource. Queries will either route to under orders or fulfillment_orders. There's also the `cancel` action that's defined for both legacy and FO fulfullment.

lib/shopify_api/resources/assigned_fulfillment_order.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    ALL_ASSIGNED_STATUSES = [
+      CANCELLATION_REQUESTED = 'cancellation_requested',
+      FULFILLMENT_REQUESTED = 'fulfillment_requested',
+      FULFILLMENT_ACCEPTED = 'fulfillment_accepted'
+    ].freeze
+
+    def self.all(options = {})
+      params = options[:params] || options['params'] || {}
+      assigned_status = params[:assigned_status] || params['assigned_status']
```
froot: Do we need to be matching on both string and symbol keys for the options hash?

lib/shopify_api/resources/assigned_fulfillment_order.rb
```diff
@@ -0,0 +1,12 @@
+module ShopifyAPI
+  class AssignedFulfillmentOrder < Base
+    CANCELLATION_REQUESTED = 'cancellation_requested'
+    FULFILLMENT_REQUESTED = 'fulfillment_requested'
+    FULFILLMENT_ACCEPTED = 'fulfillment_accepted'
+
+    def self.all(options = {})
+      assigned_fulfillment_orders = super(options)
```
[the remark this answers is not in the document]
glen-j-nkali: We could go either way. If we don't perform that check on this side there's not really any use for the constants described above.

From an API perspective I would argue that all checks that you can perform without making a request across the network should be performed locally. 

One could argue that doing so kind of forces you to keep these constant values in sync with Core but with versioning and the 1 year deprecation cycle you should have plenty of time to update them if anything changes on Core.

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
froot: We could collapse them together, and have to override a lot of the default `ActiveResource` behaviour to include both types of actions (legacy and new). It may not be necessary to do all of that extra work though, talked with Malaz and it should be fine to return a fulfillment resource that doesn't support legacy actions. Non opted-in FS shouldn't really even be querying fulfillments from fulfillment orders (even though we don't restrict them from doing so).

lib/shopify_api/resources/assigned_fulfillment_order.rb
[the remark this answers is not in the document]
froot: That's fair, I'm fine with both implementations 👍

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
karmakaze: If we know the long term desired end state then we can choose a path towards it. What is the expected adoption of FOs and support for legacy operations. Based on that, what would be clearest for consumers of the API and other supporting documentation, to have a single Fulfillment resource or the legacy and V2?

lib/shopify_api/resources/assigned_fulfillment_order.rb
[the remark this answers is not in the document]
karmakaze: My original thought was to provide the constants to prevent typos from custom client code. Putting the constants/checks here means that it should be kept in sync with server side changes. The rest of the shopify_api doesn't seem to define many constants, so removing them is more consistent.

lib/shopify_api/resources/assigned_fulfillment_order.rb
```diff
@@ -0,0 +1,24 @@
+module ShopifyAPI
+  class AssignedFulfillmentOrder < Base
+    CANCELLATION_REQUESTED = 'cancellation_requested'
+    FULFILLMENT_REQUESTED = 'fulfillment_requested'
+    FULFILLMENT_ACCEPTED = 'fulfillment_accepted'
+
+    ALL_ASSIGNED_STATUSES = [
+      CANCELLATION_REQUESTED = 'cancellation_requested',
```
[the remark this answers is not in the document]
karmakaze: Not at all. Will be removing these constants.

lib/shopify_api/resources/assigned_fulfillment_order.rb
karmakaze: There's no guarantee what users' custom code will pass in as a hash. The existing code is mostly :symbol with just a few stray 'keys', so we can just use :symbol here.

lib/shopify_api/resources/assigned_fulfillment_order.rb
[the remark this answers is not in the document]
glen-j-nkali: I think the `assigned Fulfillment orders` controller is probably the only one that accepts a custom type of filter param (assignment status) that's later converted into something else. I think that's the reason why you can't find another example on the `shopify api`.

I'm fine either way as I previously said but you should keep in mind that relying on Shopify Core to return you a `bad request` when you send in invalid `assignment status` means that you will have to parse that error message appropriately in order to return something meaningful to the user.

I ran a quick test and this is the error message Core will return :

`"Assignment status is not included in the list"`

This conveys what the issue is but isn't clear from an API standpoint IMO (what's the list). One way you could address this, could be through the documentation. Just something to keep in mind IMO.

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
froot: The long term desired end state is that all apps adopt FOs and will not be using the legacy operations. Consumers who use operations that return `FulfillmentV2` should be opted-in and thus should be using the new `V2` operations and are blocked from using legacy operations. It wouldn't be worth trying to merge the two together, just so in the future we undo all of that work. Documentation is not very clear on the distinction between the two though (both are called Fulfillment resources, one is just under the unstable version and the other latest version).

karmakaze: There's no technical problem with returning `Fulfillment` in the shopify_api from a v2 operation. Only confusion would be then trying to perform a legacy operation on it. It really comes down to how the two types of Fulfillment resources get documented. If we distinguish the operations as legacy / FO we can use the same `Fulfillment` resource in shopify_api. If we document two different types of Fulfillment resources then we can call the new one FulfillmentV2.

lib/shopify_api/resources/assigned_fulfillment_order.rb
karmakaze: Will be removing these constants and client-side checks. This will avoid duplication of logic to have to maintain sync.

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
karmakaze: We can either use the same `Fulfillment` object or a `FulfillmentV2` if we want to make a clear distinction between legacy and the newer fulfillment resource type. I'm starting to think using a single `Fulfillment` resource and documenting operations as legacy/new might be the way to go.

lib/shopify_api/resources/assigned_fulfillment_order.rb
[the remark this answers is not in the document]
karmakaze: That's not a great error message, but if the client specifies an `assigned_status` and got this Bad Request, it should be enough to point to documentation.

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
froot: Currently, the documentation exists like so:
* https://help.shopify.com/en/api/reference/shipping-and-fulfillment/fulfillment#complete-2019-10
* https://help.shopify.com/en/api/reference/shipping-and-fulfillment/unstable/fulfillmentunstable

karmakaze: Thanks this is very clear. We have two separate descriptions of distinct Fulfillment resources. We should go with `FulfillmentV2`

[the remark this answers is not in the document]
karmakaze: Naming as `FulfillmentV2` since it will most closely match the documentation having separate sections describing each form of fulfillment resource.

test/assigned_fulfillment_order_test.rb
```diff
@@ -0,0 +1,41 @@
...
+class AssignedFulFillmentOrderTest < Test::Unit::TestCase
+  context "AssignedFulfillmentOrder" do
+    context "#all" do
+      should "list assigned fulfillment orders by assigned_status" do
+        fo_fixture = load_fixture('assigned_fulfillment_orders')
+        fake 'assigned_fulfillment_orders.json?assigned_status=cancellation_requested', method: :get,
+             body: fo_fixture, extension: false
+
+        assigned_fulfillment_orders = ShopifyAPI::AssignedFulfillmentOrder.all(
+            params: { assigned_status: 'cancellation_requested' }
```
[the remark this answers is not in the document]
karmakaze: Constants and client-side checks now removed--relying on only server-side checks without duplication.

lib/shopify_api/resources/assigned_fulfillment_order.rb
```diff
@@ -0,0 +1,8 @@
+module ShopifyAPI
+  class AssignedFulfillmentOrder < Base
+    def self.all(options = {})
+      assigned_fulfillment_orders = super(options)
+      assigned_fulfillment_orders.map { |afo| FulfillmentOrder.new(afo.as_json) }
```
froot: Could we use `attributes` instead of `as_json`?

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -0,0 +1,16 @@
...
+      raise ShopifyAPI::ValidationException, "'order_id' is required" if order_id.nil? || order_id == ''
+
+      order = ::ShopifyAPI::Order.new(id: order_id)
+      order.fulfillment_orders
+    end
+
+    def fulfillments(options = {})
+      fo_fulfillments = get(:fulfillments, options)
```
froot: Anywhere you find yourself abbreviating fulfillment orders to fo, it's best to use the full phrase for clarity

lib/shopify_api/resources/assigned_fulfillment_order.rb
karmakaze: In this instance `attributes` works just fine.

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
froot: @karmakaze I'm working on documentation right now and realized that the `shipping-and-fulfillment/unstable/fulfillmentunstable` doc will be moved to under the `shipping-and-fulfillment/fulfillment` doc (under the RC version). The new `Fulfillment` resource under the 2020-01 version will have both the legacy operations and new operations that we are adding. With this, should we combine FulfillmentV2 and Fulfillment?

karmakaze: Yes, it makes sense to combine them to follow the documentation structure.

karmakaze:
Fixed the `find`/`all` problem found in the tests. There's two usages (1) direct `find` by id should use the `/fulfillment_orders/:id` endpoint, and `all` should require and order_id and use the `/orders/#/fulfillment_orders` endpoint.

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -0,0 +1,20 @@
+module ShopifyAPI
+  class FulfillmentOrder < Base
+    def self.find(scope, *args)
+      if scope == :all
+        order_id = args.first&.dig(:params, :order_id)
+        raise ShopifyAPI::ValidationException, "'order_id' is required" if order_id.blank?
+
+        order = ::ShopifyAPI::Order.new(id: order_id)
+        order.fulfillment_orders(args.first[:params].except(:order_id))
```
linzhao125: I don't think we need to support the case of `all` here. If a user wants to get all `fulfillment_orders` for an `order`, they can call `order.fulfillment_orders`. Support calling `.all` with a param `order_id` seems redundant IMO.

karmakaze: Given that FulfillmentOrder can be queried as a top-level resource with `find` it would be unexpected from an ActiveResource implementation to not support `all` in some meaningful way.

lib/shopify_api/resources/order.rb
```diff
@@ -30,6 +30,11 @@ def capture(amount = "", currency: nil)
       Transaction.create(capture_transaction)
     end
 
+    def fulfillment_orders(options = {})
```
tanema: Could this not be done with a `has_many`?

lib/shopify_api/resources/assigned_fulfillment_order.rb
```diff
@@ -0,0 +1,8 @@
+module ShopifyAPI
+  class AssignedFulfillmentOrder < Base
+    def self.find(scope, *args)
+      assigned_fulfillment_orders = super(scope, *args)
+      assigned_fulfillment_orders.map { |afo| FulfillmentOrder.new(afo.attributes) }
```
tanema: Wont this cause complications? You find assigned fulfilments and then if you update that FulfillmentOrder it will hit the fulfillments endpoint. Why are you returning FulfillmentOrders from AssignedFulfillmentOrder. (this may just be me not understanding the work that you are doing)

karmakaze: `AssignedFulfillmentOrders` are just `FulfillmentOrders` that just happen to be assigned. We didn't want to create two different ActiveResource types for subsequent actions common to both `AssignedFulfillmentOrders` and `FulfillmentOrders` so we convert the first to the second.

lib/shopify_api/resources/order.rb
karmakaze: That doesn't quite work for us because `FulfillmentOrder` is both a top-level resource (i.e. /fulfillment_orders/:id) as well as defined within an order (/orders/:order_id/fulfillments). We resolved this by creating the FulfillmentOrder ActiveResource as a top-level one (no :order_id prefix_option) then handling the order.fulfillment_orders with method implementation rather than association.

PR #635, opened 2019-11-08
Add move, cancel, close actions FulfillmentOrder

karmakaze:

POST fulfillment_orders/:fulfillment_order_id/move
POST fulfillment_orders/:fulfillment_order_id/cancel
POST fulfillment_orders/:fulfillment_order_id/close

GET requests are in PR #633
Subsequent PRs will add other fulfillment order related write requests to relevant resources.

The approach I took was to add the `move`, `cancel`, and `close` actions on the `FulfillmentOrder` resource.

Because there's already a legacy `Fulfillment` resource, I named the fulfillment orders fulfillment as `FulfillmentOrderFulfillment`.

Since all that's really needed is the `fulfillment_order_id`, if it's known you can do (rather than querying it first):
```
  closed_fulfillment_order = FulfillmentOrder.new(id: <the-fo-id>).close
```

### DIFF
--- a/lib/shopify_api/resources/fulfillment_order.rb
+++ b/lib/shopify_api/resources/fulfillment_order.rb
@@ -16,5 +16,48 @@ module ShopifyAPI
       fulfillment_hashes = get(:fulfillments, options)
       fulfillment_hashes.map { |fulfillment_hash| Fulfillment.new(fulfillment_hash) }
     end
+
+    def move(new_location_id:)
+      body = {
+        fulfillment_order: {
+          new_location_id: new_location_id
+        }
+      }
+      keyed_fulfillment_orders = keyed_fulfillment_orders_from_response(post(:move, {}, body.to_json))
+      load_keyed_fulfillment_order(keyed_fulfillment_orders, 'original_fulfillment_order')
+      keyed_fulfillment_orders
+    end
+
+    def cancel
+      keyed_fulfillment_orders = keyed_fulfillment_orders_from_response(post(:cancel, {}, only_id))
+      load_keyed_fulfillment_order(keyed_fulfillment_orders, 'fulfillment_order')
+      keyed_fulfillment_orders
+    end
+
+    def close(message: nil)
+      body = {
+        fulfillment_order: {
+          message: message
+        }
+      }
+      load_attributes_from_response(post(:close, {}, body.to_json))
+    end
+
+    private
+
+    def load_keyed_fulfillment_order(keyed_fulfillment_orders, key)
+      if keyed_fulfillment_orders[key]&.attributes
+        load(keyed_fulfillment_orders[key].attributes, false, true)
+      end
+    end
+
+    def keyed_fulfillment_orders_from_response(response)
+      return load_attributes_from_response(response) if response.code != '200'
+
+      keyed_fulfillment_orders = ActiveSupport::JSON.decode(response.body)
+      keyed_fulfillment_orders.transform_values do |fulfillment_order_attributes|
+        FulfillmentOrder.new(fulfillment_order_attributes) if fulfillment_order_attributes
+      end
+    end
   end
 end
--- a/test/fulfillment_order_test.rb
+++ b/test/fulfillment_order_test.rb
@@ -45,7 +45,6 @@ class FulFillmentOrderTest < Test::Unit::TestCase
     context "#fulfillments" do
       should "be able to list fulfillments for a fulfillment order" do
         fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021)
-
         fake "fulfillment_orders/#{fulfillment_order.id}/fulfillments", method: :get,
              body: load_fixture('fulfillments')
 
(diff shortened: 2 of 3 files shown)


froot:
A few questions:
* Given that `move` and `cancel` both return the original fulfillment order somewhere in the payload, shouldn’t we call `load_attributes_from_response` to populate the current object with the updated attributes?
• According to https://help.shopify.com/en/api/reference/shipping-and-fulfillment/unstable/fulfillmentorder#close should we enable the app the provide an option message in the body on close?

test/fulfillment_order_test.rb
```diff
@@ -49,5 +48,80 @@ def setup
...
+        moved['assigned_location_id'] = new_location_id
+        fulfillment_order.status = 'closed'
+
+        body = {
+            original_fulfillment_order: fulfillment_order,
+            moved_fulfillment_order: moved,
+            remaining_fulfillment_order: nil,
+        }
+        api_version = ShopifyAPI::ApiVersion.find_version('2019-01')
+        endpoint = "fulfillment_orders/519788021/move"
```
justinkrol: Looks like this should use `#{moved.id}`

test/fulfillment_order_test.rb
```diff
@@ -49,5 +48,80 @@ def setup
...
+        moved = ActiveSupport::JSON.decode(load_fixture('fulfillment_order'))
+        moved['assigned_location_id'] = new_location_id
+        fulfillment_order.status = 'closed'
+
+        body = {
+            original_fulfillment_order: fulfillment_order,
+            moved_fulfillment_order: moved,
+            remaining_fulfillment_order: nil,
+        }
+        api_version = ShopifyAPI::ApiVersion.find_version('2019-01')
```
justinkrol: Why `2019-01`?

test/fulfillment_order_test.rb
```diff
@@ -49,5 +48,80 @@ def setup
...
+        endpoint = "fulfillment_orders/519788021/move"
+        extension = ".json"
+        url = "https://this-is-my-test-shop.myshopify.com#{api_version.construct_api_path("#{endpoint}#{extension}")}"
+        url = url + "?fulfillment_order%5Bnew_location_id%5D=5"
+        fake endpoint, :method => :post, :url => url, :body => ActiveSupport::JSON.encode(body)
```
justinkrol: what does the `url` argument do?

test/fulfillment_order_test.rb
karmakaze: That's the current version for building the full URL. I've since found a simpler way of providing the fullpath/query_string using `extension: false` to avoid having to use a named version here. Will update.

test/fulfillment_order_test.rb
karmakaze: so the 'endpoint' is the short path form, and 'url' is the full url long form. As I mentioned there's a simpler way to achieve what's needed here.

test/fulfillment_order_test.rb
karmakaze: The URL paths for REST from routes:
```
    resources :fulfillment_orders, only: [:show] do
        member do
            post :cancel
            post :close
            post :move
        end
```
So the path ends up being .../fulfillment_orders/519788021/move.json

test/fulfillment_order_test.rb
```diff
@@ -49,5 +48,83 @@ def setup
...
+        original.status = 'closed'
+        moved = ActiveSupport::JSON.decode(load_fixture('fulfillment_order'))
+        moved['assigned_location_id'] = new_location_id
+
+        body = {
+            original_fulfillment_order: original,
+            moved_fulfillment_order: moved,
+            remaining_fulfillment_order: nil,
+        }
+        fake_query = "?fulfillment_order%5Bnew_location_id%5D=5"
```
justinkrol: Does ShopifyAPI add these url params when making the request to Shopify? For a POST, shouldn't this be in the payload?

karmakaze: I was wondering this exact thing when I first noticed it. I've been tophatting these changes with actual requests to the REST api and they seem to work. I suspect that the back-end is reading params from either the body or url. I can investigate further to make the tests work with POSTed bodies.

karmakaze: Now making requests with posted content rather than query params.

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -12,5 +12,41 @@ def fulfillments(options = {})
...
+    def close
+      load_attributes_from_response(post(:close, {}, only_id))
+    end
+
+    private
+
+    def load_keyed_attributes_from_response(response)
+      return load_attributes_from_response(response) if response.code != '200'
+
+      keyed_fulfillments = ActiveSupport::JSON.decode(response.body)
```
glen-j-nkali: nit: correct me if I'm mistaken but this variable is called `keyed_fulfillments` but when you apply the map on it below you get a `key` and `fo_attributes` should it be `keyed_fulfillment_orders` instead ?

Also I think the structure returned here would look like :

```
{
          original_fulfillment_order: 
          moved_fulfillment_order: 
          remaining_fulfillment_order:
}
```

correct ?

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -12,5 +12,41 @@ def fulfillments(options = {})
...
+    end
+
+    def close
+      load_attributes_from_response(post(:close, {}, only_id))
+    end
+
+    private
+
+    def load_keyed_attributes_from_response(response)
+      return load_attributes_from_response(response) if response.code != '200'
```
glen-j-nkali: could you please expand on what the structure returned in this case would look like ?

If I'm following things correctly, we will return here if the call either returned a `404` or a `422` with an error message, so are you just loading that error message ?

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -12,5 +12,41 @@ def fulfillments(options = {})
       fo_fulfillments = get(:fulfillments, options)
       fo_fulfillments.map { |fof| FulfillmentOrderFulfillment.new(fof.as_json) }
     end
+
+    def move(new_location_id:)
+      body = { fulfillment_order: { new_location_id: new_location_id } }.to_json
+      keyed_fos = load_keyed_attributes_from_response(post(:move, {}, body))
```
glen-j-nkali: I think the naming in this method could be simplified to improve readability.

you call the method `load_keyed_attributes_from_response` but if I'm following the logic correctly I think if there were actual `fulfillment order response` attributes for one of the following keys :
```
 original_fulfillment_order: 
 moved_fulfillment_order: 
 remaining_fulfillment_order:
```

you will load the actual `fulfillmentOrder` so the method returns `keyed_fulfillment_orders` and that's why the variable is named `keyed_fos` but the method name doesn't convey this information.

thoughts?

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -12,5 +12,41 @@ def fulfillments(options = {})
...
+      if keyed_fos&.fetch('original_fulfillment_order', nil)&.attributes
+        load(keyed_fos['original_fulfillment_order'].attributes, false, true)
```
glen-j-nkali: just for my personal comprehension here but what are you trying to achieve here ? Specifically why does this check only apply to the `original_fulfillment_order` ?

Are you using it to determine wether or not we got here with the `error message` payload ?

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -12,5 +12,41 @@ def fulfillments(options = {})
...
+      end
+      keyed_fos
+    end
+
+    def cancel
+      keyed_fos = load_keyed_attributes_from_response(post(:cancel, {}, only_id))
+      if keyed_fos&.fetch('fulfillment_order', nil)&.attributes
```
glen-j-nkali: same question as above about why specifically this key ?

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -12,5 +12,41 @@ def fulfillments(options = {})
...
+        load(keyed_fos['fulfillment_order'].attributes, false, true)
+      end
+      keyed_fos
+    end
+
+    def close
+      load_attributes_from_response(post(:close, {}, only_id))
```
glen-j-nkali: Is it safe to assume that this will just return the `fulfillment order response` received from Shopify/Shopify without doing any parsing (aka doing `FulfillmentOrder.new(fo_attributes)` )?

lib/shopify_api/resources/fulfillment_order.rb
karmakaze: Yes, good catch of my copy/pasta fail

lib/shopify_api/resources/fulfillment_order.rb
karmakaze: The idea here is that `load_keyed_attributes_from_response` reads multiple items in the response when OK (200) otherwise leverages the error handling in the commonly used `load_attributes_from_response`, which I'm guessing doesn't actually load anything just passes the error code / message back through in some way.

lib/shopify_api/resources/fulfillment_order.rb
karmakaze: 👍, `load_keyed_fulfillment_orders_from_response`?

lib/shopify_api/resources/fulfillment_order.rb
karmakaze: I was a bit uncertain about this one. The request is made using an instance of a FulfillmentOrder, e.g. `fulfillment_order.move(...)` where the server responds with multiple keyed FOs. I took the one named `original_fulfillment_order` to represent the mutation of the `fulfillment_order` that was the subject of the `.move` and so load the attributes of `original_fulfillment_order` into itself.

This is following the pattern where responses that return a single FO and calling `load_attributes_from_response` will load the calling object's attrributes from the only keyed value.

lib/shopify_api/resources/fulfillment_order.rb
karmakaze: We don't need to do `FulfillmentOrder.new(fo_attributes)` because the subject of the `.close` method is itself an instance of `FulfillmentOrder` and just updating its attributes follows the ActiveResource way of making network operations behave like ActiveRecord mutating methods. It's a bit weird but I'm learning to follow it.

lib/shopify_api/resources/fulfillment_order.rb
karmakaze: This key represents the mutation of the subject FO resource issuing the `.cancel`.

test/fulfillment_order_test.rb
```diff
@@ -54,5 +53,84 @@ def setup
...
+          fulfillment_order: cancelled,
+          replacement_fulfillment_order: fulfillment_order,
+        }
+        fake "fulfillment_orders/519788021/cancel", :method => :post, :body => ActiveSupport::JSON.encode(body)
+
+        response_fos = fulfillment_order.cancel
+
+        assert_equal 'cancelled', fulfillment_order.status
+        assert_equal 2, response_fos.count
+        fulfillment_order = response_fos['fulfillment_order']
```
froot: I would rename this to something other than `fulfillment_order`, maybe `response_fulfillment_order`.

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -12,5 +12,39 @@ def fulfillments(options = {})
...
+      keyed_fulfillment_orders
+    end
+
+    def cancel
+      keyed_fulfillment_orders = keyed_fulfillment_orders_from_response(post(:cancel, {}, only_id))
+      load_keyed_fulfillment_order(keyed_fulfillment_orders, 'fulfillment_order')
+      keyed_fulfillment_orders
+    end
+
+    def close
```
froot: Should be `def close(message: nil)` according to https://help.shopify.com/en/api/reference/shipping-and-fulfillment/unstable/fulfillmentorder#close

test/assigned_fulfillment_order_test.rb
```diff
@@ -14,7 +14,7 @@ class AssignedFulFillmentOrderTest < Test::Unit::TestCase
 
         assert_equal 2, assigned_fulfillment_orders.count
         assigned_fulfillment_orders.each do |fulfillment_order|
-          assert_equal 'ShopifyAPI::FulfillmentOrder', fulfillment_order.class.name
+          assert fulfillment_order.is_a?(ShopifyAPI::FulfillmentOrder)
```
froot: nit: should probably move this into the other PR dealing with assigned fulfillment order.

test/fulfillment_order_test.rb
```diff
@@ -57,5 +56,91 @@ def setup
...
+            message: "Test close message."
+          }
+        }
+        fake "fulfillment_orders/519788021/close", :method => :post,
+          :request_body => ActiveSupport::JSON.encode(request_body),
+          :body => ActiveSupport::JSON.encode(closed)
+
+        assert_equal 'open', fulfillment_order.status
+        assert fulfillment_order.close(message: "Test close message.")
+        assert_equal 'closed', fulfillment_order.status
```
linzhao125: I understand that the call to Shopify core is a stub. Just want to point out that in case of `close`, `fulfillment_order` status has to be `in_progress`. It becomes `incomplete` after the call.

PR #637, opened 2019-11-13
Fulfillment orders fulfillment cancellation request

karmakaze:

POST fulfillment_orders/:fulfillment_order_id/fulfillment_request
POST fulfillment_orders/:fulfillment_order_id/fulfillment_request/accept
POST fulfillment_orders/:fulfillment_order_id/fulfillment_request/reject

POST fulfillment_orders/:fulfillment_order_id/cancellation_request
POST fulfillment_orders/:fulfillment_order_id/cancellation_request/accept
POST fulfillment_orders/:fulfillment_order_id/cancellation_request/reject

GET requests are in PR #633
Actions move, cancel, close for fulfillment orders are in PR #635
Subsequent PRs will add other fulfillment order related POST requests to relevant resources.

Since all that's really needed is the `fulfillment_order_id`, if it's known you can do (rather than querying it first):
```
  FulfillmentOrder.new(id: <the-fo-id>).fulfillment_request(<params>)
```

Note to reviewer:

- when the endpoint returns a single resource, it is loaded into the current resource and `true` is returned if the request is ok
- when the endpoint returns multiple keyed resources (e.g. original_fulfillment_order, submitted_fulfillment_order, unsubmitted_fulfillment_order) the return value is a hash of the multiple resources keyed by the names. Is it conventional to reload the current resource with one of these as well?

### DIFF
--- a/lib/shopify_api/resources/fulfillment_order.rb
+++ b/lib/shopify_api/resources/fulfillment_order.rb
@@ -43,6 +43,63 @@ module ShopifyAPI
       load_attributes_from_response(post(:close, {}, body.to_json))
     end
 
+    def request_fulfillment(fulfillment_order_line_items: nil, message: nil)
+      body = {
+        fulfillment_request: {
+          fulfillment_order_line_items: fulfillment_order_line_items,
+          message: message
+        }
+      }
+      keyed_fulfillment_orders = keyed_fulfillment_orders_from_response(post(:fulfillment_request, {}, body.to_json))
+      load_keyed_fulfillment_order(keyed_fulfillment_orders, 'original_fulfillment_order')
+      keyed_fulfillment_orders
+    end
+
+    def accept_fulfillment_request(message: nil)
+      body = {
+        fulfillment_request: {
+          message: message
+        }
+      }
+      load_attributes_from_response(post('fulfillment_request/accept', {}, body.to_json))
+    end
+
+    def reject_fulfillment_request(message: nil)
+      body = {
+        fulfillment_request: {
+          message: message
+        }
+      }
+      load_attributes_from_response(post('fulfillment_request/reject', {}, body.to_json))
+    end
+
+    def request_cancellation(message: nil)
+      body = {
+        cancellation_request: {
+          message: message
+        }
+      }
+      load_attributes_from_response(post(:cancellation_request, {}, body.to_json))
+    end
+
+    def accept_cancellation_request(message: nil)
+      body = {
+        cancellation_request: {
+          message: message
+        }
+      }
+      load_attributes_from_response(post('cancellation_request/accept', {}, body.to_json))
+    end
+
+    def reject_cancellation_request(message: nil)
+      body = {
+        cancellation_request: {
+          message: message
+        }
+      }
+      load_attributes_from_response(post('cancellation_request/reject', {}, body.to_json))
+    end
+
     private
 
     def load_keyed_fulfillment_order(keyed_fulfillment_orders, key)
--- a/test/fulfillment_order_test.rb
+++ b/test/fulfillment_order_test.rb
@@ -142,5 +142,234 @@ class FulFillmentOrderTest < Test::Unit::TestCase
         assert_equal 'incomplete', fulfillment_order.status
       end
     end
+
+    context "#request_fulfillment" do
+      should "make a fulfillment request for a fulfillment order including unsubmitted" do
+        fake_original_fulfillment_order = ActiveSupport::JSON.decode(load_fixture('fulfillment_order'))
+        fake_original_fulfillment_order['status'] = 'closed'
+        fake_submitted_fulfillment_order = fake_original_fulfillment_order.clone
+        fake_submitted_fulfillment_order['id'] = 2
+        fake_submitted_fulfillment_order['status'] = 'open'
+        fake_submitted_fulfillment_order['request_status'] = 'submitted'
+        fake_unsubmitted_fulfillment_order = fake_original_fulfillment_order.clone
+        fake_unsubmitted_fulfillment_order['id'] = 3
+        fake_unsubmitted_fulfillment_order['status'] = 'open'
+        fake_unsubmitted_fulfillment_order['request_status'] = 'unsubmitted'
+        body = {
+          original_fulfillment_order: fake_original_fulfillment_order,
+          submitted_fulfillment_order: fake_submitted_fulfillment_order,
+          unsubmitted_fulfillment_order: fake_unsubmitted_fulfillment_order
+        }
+        request_body = {
+          fulfillment_request: {
+            fulfillment_order_line_items: [
+              { id: 1, quantity: 1 }
+            ],
+            message: 'Fulfill this FO, please.'
+          }
+        }
+        fake "fulfillment_orders/519788021/fulfillment_request", :method => :post,
+          :request_body => ActiveSupport::JSON.encode(request_body),
+          :body => ActiveSupport::JSON.encode(body)
+
+        fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021)
+        params = {
+          fulfillment_order_line_items: [{ id: 1, quantity: 1 }],
+          message: "Fulfill this FO, please."
+        }
+        response_fulfillment_orders = fulfillment_order.request_fulfillment(params)
+
+        assert_equal 'closed', fulfillment_order.status
+        assert_equal 3, response_fulfillment_orders.size
+
+        original_fulfillment_order = response_fulfillment_orders['original_fulfillment_order']
+        assert_equal 519788021, original_fulfillment_order.id
+        assert_equal 'closed', original_fulfillment_order.status
+
+        submitted_fulfillment_order = response_fulfillment_orders['submitted_fulfillment_order']
+        assert_equal 2, submitted_fulfillment_order.id
+        assert_equal 'open', submitted_fulfillment_order.status
+        assert_equal 'submitted', submitted_fulfillment_order.request_status
+
+        unsubmitted_fulfillment_order = response_fulfillment_orders['unsubmitted_fulfillment_order']
+        assert_equal 3, unsubmitted_fulfillment_order.id
+        assert_equal 'open', unsubmitted_fulfillment_order.status
+        assert_equal 'unsubmitted', unsubmitted_fulfillment_order.request_status
+      end
+
+      should "make a fulfillment request for a fulfillment order excluding unsubmitted" do
+        fake_original_fulfillment_order = ActiveSupport::JSON.decode(load_fixture('fulfillment_order'))
+        fake_original_fulfillment_order['status'] = 'closed'
+        fake_submitted_fulfillment_order = fake_original_fulfillment_order.clone
+        fake_submitted_fulfillment_order['id'] = 2
+        fake_submitted_fulfillment_order['status'] = 'open'
+        fake_submitted_fulfillment_order['request_status'] = 'submitted'
+        body = {
+          original_fulfillment_order: fake_original_fulfillment_order,
+          submitted_fulfillment_order: fake_submitted_fulfillment_order,
+          unsubmitted_fulfillment_order: nil,
+        }
+        request_body = {
+          fulfillment_request: {
+            fulfillment_order_line_items: [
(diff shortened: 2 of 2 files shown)


lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -34,13 +34,52 @@ def close
...
+      body = {
+        fulfillment_request: {
+          fulfillment_order_line_items: fulfillment_order_line_items,
+          message: message
+        }
+      }
+      keyed_fos = load_keyed_attributes_from_response(post(:fulfillment_request, {}, body.to_json))
+      if keyed_fos&.fetch('original_fulfillment_order', nil)&.attributes
+        load(keyed_fos['original_fulfillment_order'].attributes, false, true)
+      end
```
justinkrol: What does this do?

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -34,13 +34,52 @@ def close
...
+
+    def accept_fulfillment_request(params)
+      load_attributes_from_response(post('fulfillment_request/accept', {}, params.to_json))
+    end
+
+    def reject_fulfillment_request(params)
+      load_attributes_from_response(post('fulfillment_request/reject', {}, params.to_json))
+    end
+
+    def cancellation_request(message:)
```
justinkrol: I think this name should be `request_cancellation`

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -34,13 +34,52 @@ def close
       load_attributes_from_response(post(:close, {}, only_id))
     end
 
+    def fulfillment_request(fulfillment_order_line_items:, message:)
```
justinkrol: I think this name should be `request_fulfillment`

test/fulfillment_order_test.rb
```diff
@@ -132,5 +132,173 @@ def setup
...
+      should "make a fulfillment request for a fulfillment order" do
+        original_fulfillment_order = ActiveSupport::JSON.decode(load_fixture('fulfillment_order'))
+        submitted_fulfillment_order = original_fulfillment_order.clone
+        submitted_fulfillment_order['id'] = 2
+        submitted_fulfillment_order['status'] = 'open'
+        submitted_fulfillment_order['request_status'] = 'submitted'
+        unsubmitted_fulfillment_order = original_fulfillment_order.clone
+        unsubmitted_fulfillment_order['id'] = 3
+        unsubmitted_fulfillment_order['request_status'] = 'unsubmitted'
+        original_fulfillment_order['status'] = 'in_progress'
```
justinkrol: It looks like you're describing a split fulfillment request with this test. Ex.
FO for 3 hats
Request 1 hat, new FO created with 1 hat (open/submitted), remaining FO created with 2 hats (open/unsubmitted).

In this case, the original FO should be `closed`.

test/fulfillment_order_test.rb
```diff
@@ -132,5 +132,173 @@ def setup
...
+        fake "fulfillment_orders/519788021/fulfillment_request", :method => :post,
+          :request_body => ActiveSupport::JSON.encode(request_body),
+          :body => ActiveSupport::JSON.encode(body)
+
+        fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021)
+        params = {
+          fulfillment_order_line_items: [{ id: 1, quantity: 1 }],
+          message: "Fulfill this FO, please."
+        }
+        original_submitted_unsubmitted_fos = fulfillment_order.fulfillment_request(params)
```
justinkrol: This variable name is confusing, `result` or `response_fulfillment_orders` would be better.

test/fulfillment_order_test.rb
```diff
@@ -132,5 +132,173 @@ def setup
...
+        assert_equal 2, submitted_fo.id
+        assert_equal 'open', submitted_fo.status
+        assert_equal 'submitted', submitted_fo.request_status
+
+        unsubmitted_fo = original_submitted_unsubmitted_fos['unsubmitted_fulfillment_order']
+        assert_equal 3, unsubmitted_fo.id
+        assert_equal 'open', unsubmitted_fo.status
+        assert_equal 'unsubmitted', unsubmitted_fo.request_status
+      end
+    end
```
justinkrol: It's also possible for the response to be something like:
```
{
  original_fulfillment_order: {...},
  submitted_fulfillment_order: {...exact same as original...},
  unsubmitted_fulfillment_order: nil,
}
```
This case should also be tested.

test/fulfillment_order_test.rb
```diff
@@ -132,5 +132,173 @@ def setup
...
+
+    context "#accept_fulfillment_request" do
+      should "accept a fulfillment request for a fulfillment order" do
+        fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021)
+
+        request_body = {
+          'message' => "LGTM. Accept this FO fulfillment request",
+          'other' => "random"
```
justinkrol: What is this param for? Instead of passing through whatever params, maybe `FulfillmentOrder#accept_fulfillment_request` should explicitly accept `message` as an arg.

test/test_helper.rb
```diff
@@ -112,6 +112,11 @@ def fake(endpoint, options={})
         )
       end
 
+      def query_string(hash)
```
justinkrol: It doesn't look like this is used anywhere

lib/shopify_api/resources/fulfillment_order.rb
karmakaze: The caller code will look like `fulfillment_order.fulfillment_request(...)`. This code will use the attributes of the fulfillment order in the response body with hash key `original_fulfillment_order` to update the `fulfillment_order` that was used to call `.fulfillment_request`

test/fulfillment_order_test.rb
karmakaze: This is following what's being done on the back-end. The params appear to be open-ended (note the `request_data: params.to_s`):
```
    def accept
      fulfillment_order_id = params.require_type(fulfillment_order_id: :id)

      result = Fulfillments::RecordFulfillmentOrderFulfillAccept.call(
        fulfillment_order_id: fulfillment_order_id,
        shop_id: trusted_shop_id,
        api_client_id: current_api_client.id,
        fulfillment_context: Fulfillments::Context::FULFILLMENT_REQUESTS_CONTROLLER,
        message: params.fetch(:fulfillment_request, {}).permit(:message),
        request_key: request.uuid,
        request_data: params.to_s,
      )
```

test/fulfillment_order_test.rb
karmakaze: Added above.

test/fulfillment_order_test.rb
```diff
@@ -132,5 +132,197 @@ def setup
...
+
+    context "#request_cancellation" do
+      should "make a cancellation request for a fulfillment order" do
+        fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021)
+
+        cancelling = ActiveSupport::JSON.decode(load_fixture('fulfillment_order'))
+        cancelling['status'] = 'in_progress'
+        cancelling['request_status'] = 'cancellation_accepted'
+        fake "fulfillment_orders/519788021/cancellation_request", :method => :post,
+          :body => ActiveSupport::JSON.encode(cancelling)
```
froot: `ActiveSupport::JSON.encode({fulfillment_order: cancelling})` to stay consistent with what the API returns and ensuring that the deserialization works as expected.

test/fulfillment_order_test.rb
```diff
@@ -132,5 +132,197 @@ def setup
...
+        assert_equal 'cancellation_accepted', fulfillment_order.request_status
+      end
+    end
+
+    context "#accept_cancellation_request" do
+      should "accept a cancellation request for a fulfillment order" do
+        fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021)
+
+        message = 'Already in-progress. Reject this FO cancellation request'
+        request_body = { 'message' => message }
```
froot: nit: to keep consistent with what the API actually expects, this would be `{ 'fulfillment_request' => { 'message' => message } }` instead.

froot: Likewise for the test below.

test/fulfillment_order_test.rb
```diff
@@ -132,5 +132,197 @@ def setup
...
+        fake_response = { fulfillment_order: fulfillment_order.attributes.merge(status: 'closed') }
```
froot: nit: `cancelled` instead of `closed` is the expected status of a cancelled fulfillment order.

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -39,6 +39,64 @@ def close(message: nil)
...
+      body = {
+        cancellation_request: {
+          message: message
+        }
+      }
+      load_attributes_from_response(post('cancellation_request/accept', {}, body.to_json))
+    end
+
+    def reject_cancellation_request(message: nil)
+
```
froot: nit: remove unnecessary whitespace.

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -39,6 +39,64 @@ def close(message: nil)
       load_attributes_from_response(post(:close, {}, body.to_json))
     end
 
+    def request_fulfillment(fulfillment_order_line_items:, message: nil)
```
froot: > The fulfillment order line items to be requested for fulfillment. If left blank, all line items of the fulfillment order are requested for fulfillment.
https://help.shopify.com/en/api/reference/shipping-and-fulfillment/fulfillmentrequest#create-2020-01

It seems like `fulfillment_order_line_items` can also be left blank

test/fulfillment_order_test.rb
```diff
@@ -139,5 +139,223 @@ def setup
...
+
+        message = "Cancelling this please."
+        request_body = {
+          'cancellation_request' => {
+            'message' => message
+          }
+        }
+        cancelling = ActiveSupport::JSON.decode(load_fixture('fulfillment_order'))
+        cancelling['status'] = 'in_progress'
+        cancelling['request_status'] = 'cancellation_accepted'
```
froot: Nit: should be `cancellation_requested` instead of `cancellation_accepted`

test/fulfillment_order_test.rb
```diff
@@ -139,5 +139,223 @@ def setup
...
+      should "reject a cancellation request for a fulfillment order" do
+        fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021)
+
+        message = "Already in-progress. Reject this FO cancellation request"
+        request_body = {
+          'cancellation_request' => {
+            'message' => message
+          }
+        }
+        fake_response = { fulfillment_order: fulfillment_order.attributes.merge(status: 'in_progress') }
```
froot: Nit: the request status would also be changed to "cancellation_rejected" as well as the status to "in_progress"

test/fulfillment_order_test.rb
```diff
@@ -139,5 +139,223 @@ def setup
...
+      should "accept a cancellation request for a fulfillment order" do
+        fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021)
+
+        message = 'Already in-progress. Reject this FO cancellation request'
+        request_body = {
+          'cancellation_request' => {
+            'message' => message
+          }
+        }
+        fake_response = { fulfillment_order: fulfillment_order.attributes.merge(status: 'cancelled') }
```
froot: Nit: the request status would also be changed to "cancellation_accepted" as well as the status changed to "cancelled".

test/fulfillment_order_test.rb
```diff
@@ -139,5 +139,223 @@ def setup
...
+      should "accept a fulfillment request for a fulfillment order" do
+        fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021)
+
+        message = "LGTM. Accept this FO fulfillment request"
+        request_body = {
+          'fulfillment_request' => {
+            'message' => message
+          }
+        }
+        fake_response = { fulfillment_order: fulfillment_order.attributes.merge(status: 'in_progress') }
```
froot: nit: the request_status would be changed to `accepted` as well as the status change.

PR #639, opened 2019-11-14
Add save, update_tracking, and cancel actions to FulfillmentOrderFulfillment resource

karmakaze:

POST fulfillments (via FulfillmentOrderFulfillment.save)
POST fulfillments/:fulfillment_id/update_tracking
POST fulfillments/:fulfillment_id/cancel

GET requests are in PR #633
FulfillmentOrder actions move, cancel, close are in PR #635
FulfillmentOrder actions fulfillment_request (+ accept/reject), cancellation_request (+ accept/reject) are in PR #637
Subsequent PRs will add other fulfillment order related requests to relevant resources.

Since all that's really needed is the `fulfillment_id`, if it's known you can do (rather than querying it first):
```
  FulfillmentOrderFulfillment.new(id: <the-fulfillment-id>).update_tracking(<params>)
```

Notes to reviewer:

- there's already a `Fulfillment` resource which I haven't yet determined is the same/compatible with FO fulfillments so created a new resource FulfillmentOrderFulfillment
- the specific statuses used in the tests may not make logical sense in terms of fulfillment order/fulfillment lifecycle. the tests do verify proper request dispatching and serialization

### DIFF
--- a/lib/shopify_api/resources/fulfillment.rb
+++ b/lib/shopify_api/resources/fulfillment.rb
@@ -9,5 +9,39 @@ module ShopifyAPI
     def cancel; load_attributes_from_response(post(:cancel, {}, only_id)); end
     def complete; load_attributes_from_response(post(:complete, {}, only_id)); end
     def open; load_attributes_from_response(post(:open, {}, only_id)); end
+
+    def order_id=(order_id)
+      prefix_options[:order_id] = order_id
+    end
+
+    def load(attributes, remove_root = false, persisted = false)
+      order_id = attributes['order_id']
+      prefix_options[:order_id] = order_id if order_id
+      super(attributes, remove_root, persisted)
+    end
+
+    def save
+      if prefix_options[:order_id].present?
+        super
+      else
+        line_items = attributes['line_items_by_fulfillment_order'] || attributes[:line_items_by_fulfillment_order]
+        if line_items.blank?
+          raise ShopifyAPI::ValidationException,
+                "either 'line_items_by_fulfillment_order' or prefix_options[:order_id] is required"
+        end
+
+        fulfillmentV2 = FulfillmentV2.new(attributes)
+        result = fulfillmentV2.save
+        load(fulfillmentV2.attributes, false, true)
+        result
+      end
+    end
+
+    def update_tracking(tracking_info:, notify_customer:)
+      fulfillmentV2 = FulfillmentV2.new(attributes)
+      result = fulfillmentV2.update_tracking(tracking_info: tracking_info, notify_customer: notify_customer)
+      load(fulfillmentV2.attributes, false, true)
+      result
+    end
   end
 end
--- a/test/fulfillment_v2_test.rb
+++ b/test/fulfillment_v2_test.rb
@@ -1,0 +1,40 @@
+require 'test_helper'
+
+class FulfillmentV2Test < Test::Unit::TestCase
+  context "FulfillmentV2" do
+    context "#update_tracking" do
+      should "be able to update tracking info for a fulfillment" do
+        tracking_info = {
+          number: 'JSDHFHAG',
+          url: 'https://example.com/fulfillment_tracking/JSDHFHAG',
+          company: 'ACME co',
+        }
+        fake_fulfillment = ActiveSupport::JSON.decode(load_fixture('fulfillment'))['fulfillment']
+        fake_fulfillment['tracking_number'] = tracking_info[:number]
+        fake_fulfillment['tracking_numbers'] = [tracking_info[:number]]
+        fake_fulfillment['tracking_url'] = tracking_info[:url]
+        fake_fulfillment['tracking_urls'] = [tracking_info[:url]]
+        fake_fulfillment['tracking_company'] = tracking_info[:company]
+
+        request_body = {
+          fulfillment: {
+            tracking_info: tracking_info,
+            notify_customer: true
+          }
+        }
+        fake "fulfillments/#{fake_fulfillment['id']}/update_tracking", method: :post,
+          request_body: ActiveSupport::JSON.encode(request_body),
+          body: ActiveSupport::JSON.encode(fulfillment: fake_fulfillment)
+
+        fulfillment = ShopifyAPI::FulfillmentV2.new(id: fake_fulfillment['id'])
+        assert fulfillment.update_tracking(tracking_info: tracking_info, notify_customer: true)
+
+        assert_equal tracking_info[:number], fulfillment.tracking_number
+        assert_equal [tracking_info[:number]], fulfillment.tracking_numbers
+        assert_equal tracking_info[:url], fulfillment.tracking_url
+        assert_equal [tracking_info[:url]], fulfillment.tracking_urls
+        assert_equal tracking_info[:company], fulfillment.tracking_company
+      end
+    end
+  end
+end
--- a/lib/shopify_api/resources/fulfillment_v2.rb
+++ b/lib/shopify_api/resources/fulfillment_v2.rb
@@ -1,0 +1,15 @@
+module ShopifyAPI
+  class FulfillmentV2 < Base
+    self.element_name = 'fulfillment'
+
+    def update_tracking(tracking_info:, notify_customer:)
+      body = {
+        fulfillment: {
+          tracking_info: tracking_info,
+          notify_customer: notify_customer
+        }
+      }
+      load_attributes_from_response(post(:update_tracking, {}, body.to_json))
+    end
+  end
+end
--- a/test/fulfillment_test.rb
+++ b/test/fulfillment_test.rb
@@ -56,6 +56,127 @@ class FulFillmentTest < Test::Unit::TestCase
         assert_equal 450789469, fulfillment.order_id
       end
     end
-  end
 
+    context "#create" do
+      should "create a fulfillment with line_items_by_fulfillment_order" do
+        create_fulfillment_attributes = {
+          message: "The message for this FO fulfillment",
+          notify_customer: true,
+          tracking_info: {
+            number: "XSDFHYR23475",
+            url: "https://tracking.example.com/XSDFHYR23475",
+            company: "TFTC - the fulfillment/tracking company"
+          },
+          line_items_by_fulfillment_order: [
+            {
+              fulfillment_order_id: 3,
+              fulfillment_order_line_items: [{ id: 2, quantity: 1 }]
+            }
+          ]
+        }
+        request_body = { fulfillment: create_fulfillment_attributes }
+        response_body = { fulfillment: create_fulfillment_attributes.merge(id: 346743624) }
+        fake "fulfillments", :method => :post,
+          request_body: ActiveSupport::JSON.encode(request_body),
+          body: ActiveSupport::JSON.encode(response_body)
+
+        fulfillment = ShopifyAPI::Fulfillment.create(create_fulfillment_attributes)
+        assert fulfillment.is_a?(ShopifyAPI::Fulfillment)
+        assert fulfillment.persisted?
+        assert_equal 346743624, fulfillment.id
+      end
+    end
+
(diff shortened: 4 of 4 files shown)


test/fulfillment_order_fulfillment_test.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    context "#update_tracking" do
+      should "be able to update tracking info for a fulfillment" do
+        tracking_info = {
+          number: 'JSDHFHAG',
+          url: 'https://example.com/fulfillment_tracking/JSDHFHAG',
+          company: 'ACME co',
+        }
+        fake_f = ActiveSupport::JSON.decode(load_fixture('fulfillment').gsub(/1Z2345/, 'JSDHFHAG'))['fulfillment']
```
froot: We should probably assert that the fixture's tracking number, url, and company are not the ones that we are updating to initially.

test/fulfillment_order_fulfillment_test.rb
```diff
@@ -0,0 +1,84 @@
...
+      should "be able to cancel a fulfillment" do
+        fake_f = ActiveSupport::JSON.decode(load_fixture('fulfillment'))['fulfillment']
+        fake "fulfillments/#{fake_f['id']}/cancel", method: :post,
+           body: ActiveSupport::JSON.encode(fulfillment: fake_f)
+
+        fulfillment = ShopifyAPI::FulfillmentOrderFulfillment.new(id: fake_f['id'])
+        cancelled = fulfillment.cancel
+
+        assert_equal true, cancelled
+        assert_equal 'pending', fulfillment.status
```
froot: The status of the fulfillment shouldn't be `pending` after you cancel the fulfillment, it should be `cancelled`

lib/shopify_api/resources/fulfillment_order_fulfillment.rb
```diff
@@ -1,4 +1,19 @@
 module ShopifyAPI
   class FulfillmentOrderFulfillment < Base
+    self.element_name = 'fulfillment'
```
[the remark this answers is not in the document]
karmakaze: It's an override of the convention that a resource named `FulfillmentOrderFulfillment` (now `FulfillmentV2`) would have the REST path `/fulfillment_order_fulfillments` and we want the path `/fulfillments` to be used when a FulfillmentV2 resource is created.

test/fulfillment_order_fulfillment_test.rb
```diff
@@ -0,0 +1,84 @@
...
+            company: "TFTC - the fulfillment/tracking company"
+          },
+          line_items_by_fulfillment_order: [
+            {
+              fulfillment_order_id: 3,
+              fulfillment_order_line_items: [{ id: 2, quantity: 1 }]
+            }
+          ]
+        }
+        stub_request(:post, "https://this-is-my-test-shop.myshopify.com/admin/api/2019-01/fulfillments.json").with(
```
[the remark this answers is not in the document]
karmakaze: It was because `fake` took a response body but not a request body. I've since updated `fake` with a `request_body` option so it can now be used.

test/fulfillment_order_fulfillment_test.rb
karmakaze: There's only one request: `update_tracking` so we're only testing the serialization/deserialization of bodies and matching of the mocked request. I can update the test to be more explicit with the relevant fields.

froot: My thinking was along the lines of if the initial tracking info of the fulfillment is the same as the ones we are trying to update to (aka no change), then the update method could essentially do nothing (not make any network request), or the deserialization of the response messes up somewhere, and this test would still pass.

So by doing a quick assertion that the initial values indeed differ from the values that we are trying to change, we know for sure that 1. the stub was hit and 2. the deserialization did not fail.

test/fulfillment_v2_test.rb
```diff
@@ -0,0 +1,91 @@
...
+          }
+        }
+        fake "fulfillments/#{fake_f['id']}/update_tracking", method: :post,
+          request_body: ActiveSupport::JSON.encode(request_body),
+          body: ActiveSupport::JSON.encode(fulfillment: fake_f)
+
+        fulfillment = ShopifyAPI::FulfillmentV2.new(id: fake_f['id'])
+        updated = fulfillment.update_tracking(tracking_info: tracking_info, notify_customer: true)
+
+        assert_equal true, updated
```
froot: nit: I think we can do the same here `assert fulfillment.update_tracking(tracking_info: tracking_info, notify_customer: true)` since we don't use the variable `updated` anywhere else in the code.

test/fulfillment_v2_test.rb
```diff
@@ -0,0 +1,91 @@
...
+      should "be able to cancel a fulfillment" do
+        fake_f = ActiveSupport::JSON.decode(load_fixture('fulfillment'))['fulfillment']
+        fake_f['status'] = 'cancelled'
+        fake "fulfillments/#{fake_f['id']}/cancel", method: :post,
+           body: ActiveSupport::JSON.encode(fulfillment: fake_f)
+
+        fulfillment = ShopifyAPI::FulfillmentV2.new(id: fake_f['id'])
+        cancelled = fulfillment.cancel
+
+        assert_equal true, cancelled
```
froot: nit: I would do the same thing here. Something like `assert fulfillment.cancel`.

test/fulfillment_v2_test.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    context "#update_tracking" do
+      should "be able to update tracking info for a fulfillment" do
+        tracking_info = {
+          number: 'JSDHFHAG',
+          url: 'https://example.com/fulfillment_tracking/JSDHFHAG',
+          company: 'ACME co',
+        }
+        fake_f = ActiveSupport::JSON.decode(load_fixture('fulfillment'))['fulfillment']
```
froot: nit: I think a more verbose variable would increase the readability of these tests. Maybe `fake_fulfillment` or `resulting_fulfillment`?

test/fulfillment_v2_test.rb
```diff
@@ -0,0 +1,91 @@
+require 'test_helper'
+
+class FulfillmentV2Test < Test::Unit::TestCase
+  def setup
+    super
+    fake "fulfillment_orders/519788021", method: :get,
```
froot: I don't think your tests ever use this fake, removing it would be good to avoid confusion

test/fulfillment_v2_test.rb
```diff
@@ -0,0 +1,91 @@
...
+         body: load_fixture('fulfillment_order')
+  end
+
+  context "FulfillmentOrderFulfillment" do
```
froot: Probably should change this context to `FulfillmentV2` to avoid confusion

test/fulfillment_order_fulfillment_test.rb
karmakaze: I see what you mean. That's guaranteed by [initializing the requesting fulfillment with only the id](https://github.com/Shopify/shopify_api/pull/639/files#diff-dcdc368f3229b0031925b1881ac52bdfR58-R59)

lib/shopify_api/resources/fulfillment.rb
```diff
@@ -6,8 +6,32 @@ def order_id
...
+        result
+      else
+        load_attributes_from_response(post(:cancel, {}, only_id))
+      end
+    end
+
     def complete; load_attributes_from_response(post(:complete, {}, only_id)); end
     def open; load_attributes_from_response(post(:open, {}, only_id)); end
+
+    def create
```
froot: I'm unsure what's provided by `ActiveResource::Base` but is this intended to override `self.create` the class method instead of the instance method? What about `.save`? We still need to support both `POST /admin/api/2020-01/orders/#{order_id}/fulfillments.json` and `POST /admin/api/2020-01/fulfillments.json` no?

lib/shopify_api/resources/fulfillment.rb
```diff
@@ -6,8 +6,32 @@ def order_id
       @prefix_options[:order_id]
     end
 
-    def cancel; load_attributes_from_response(post(:cancel, {}, only_id)); end
+    def cancel(using_fulfillment_orders: false)
```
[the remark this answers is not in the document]
karmakaze: Yes, we had decided on this, everyone can use the existing endpoint.

lib/shopify_api/resources/fulfillment.rb
karmakaze: Yup, that makes sense, I forgot about `save` being used for create.

lib/shopify_api/resources/fulfillment.rb
```diff
@@ -9,5 +9,26 @@ def order_id
     def cancel; load_attributes_from_response(post(:cancel, {}, only_id)); end
     def complete; load_attributes_from_response(post(:complete, {}, only_id)); end
     def open; load_attributes_from_response(post(:open, {}, only_id)); end
+
+    def self.create(attributes = {})
```
froot: I think `self.create` is synonymous to calling `new` then `save`. https://api.rubyonrails.org/v3.2.6/classes/ActiveResource/Base.html#method-c-create so we should be good by just overriding save.

lib/shopify_api/resources/fulfillment.rb
```diff
@@ -9,5 +9,26 @@ def order_id
...
+      fulfillmentV2 = FulfillmentV2.create(attributes)
+      fulfillment = Fulfillment.new
+      fulfillment.load(fulfillmentV2.attributes, false, true)
+      fulfillment
+    end
+
+    def save
```
froot: We still need to support the ability to create fulfillments using the old endpoint, as well as the new fulfillment create:

Screen Shot 2019-12-18 at 11 56 37 AM

We can do so by inspecting the passed in params and deciding if the client intended to use the legacy/new endpoint.

lib/shopify_api/resources/fulfillment.rb
```diff
@@ -9,5 +9,41 @@ def order_id
     def cancel; load_attributes_from_response(post(:cancel, {}, only_id)); end
     def complete; load_attributes_from_response(post(:complete, {}, only_id)); end
     def open; load_attributes_from_response(post(:open, {}, only_id)); end
+
+    def order_id=(order_id)
+      binding.pry
+      attributes.delete('order_id') || attributes.delete(:order_id)
```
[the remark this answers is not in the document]
karmakaze: Normally the `attributes` contains string keys. When I was refactoring this, some versions ended up with symbol keys `:order_id`. Finally in the end, neither is needed as the `prefix_options` are set explicitly by client code and not inferred from `attributes` so removing this.

lib/shopify_api/resources/fulfillment.rb
```diff
@@ -9,5 +9,39 @@ def order_id
...
+        super
+      else
+        line_items = attributes['line_items_by_fulfillment_order'] || attributes[:line_items_by_fulfillment_order]
+        if line_items.blank?
+          raise ShopifyAPI::ValidationException,
+                "either 'line_items_by_fulfillment_order' or prefix_options[:order_id] is required"
+        end
+
+        fulfillmentV2 = FulfillmentV2.new(attributes)
+        result = fulfillmentV2.save
```
tanema: Is this a way of migrating fulfillments? Why are you doing this?

karmakaze: We have this situation because the backend has two REST paths for fulfillments, the existing pre-FulfillmentOrder one at `/order/#/fulfillments` and the new FulfillmentOrder based one at `/fulfillments`. To make them interoperable we use the `Fulfillment` resource which routes to the `FulfillmentV2` resource when we need to use the `/fulfillments` path for an operation. Whenever reading a response, we always convert FulfillmentV2 back to Fulfillment.

PR #641, opened 2019-11-20
Call Session#request_token with indifferent hash keys

raydefy:
My auth request is a hash where the keys are strings rather than symbols. The `Session#request_token` did not like. In other words, this failed
```
token = session.request_token({:code=>"e3c4", :hmac=>"5ad0", :shop=>"shop1", "timestamp" =>1574273231})
```
Using `HashWithIndifferentAccess` allowed me to have mixed key types (for better or worse). In the hash below, one key clobbers the other. This seems to be acceptable behaviour for the hmac check
``` 
>> h = {"key" => 1, key: 2}
=> {"key"=>1, :key=>2}
>> h = {"key" => 1, key: 2}.with_indifferent_access
=> {"key"=>2}
```

Here is a before/after of the projects tests.

Before code change
```
$ rake test
Run options: --seed 51380

# Running:

...........................................................................................................................................................................................................................................................E........................................................

Finished in 1.985446s, 155.1289 runs/s, 261.9059 assertions/s.

  1) Error:
SessionTest#test_Indifferent params access:
ShopifyAPI::ValidationException: Invalid timestamp: Possible malicious login
    /Users/rcapozzi/src/github.com/rcapozzi-shopify/shopify_api/lib/shopify_api/session.rb:104:in `request_token'
    /Users/rcapozzi/src/github.com/rcapozzi-shopify/shopify_api/test/session_test.rb:84:in `block in <class:SessionTest>'

308 runs, 520 assertions, 0 failures, 1 errors, 0 skips
rake aborted!
Command failed with status (1)
/usr/local/lib/ruby/gems/2.6.0/gems/rake-13.0.1/exe/rake:27:in `<top (required)>'
Tasks: TOP => test
(See full trace by running task with --trace)
```
After code change

```
rcapozzi@Raymonds-MacBook-Pro shopify_api
$ rake test
Run options: --seed 41280

# Running:

....................................................................................................................................................................................................................................................................................................................

Finished in 1.933314s, 159.3119 runs/s, 269.4855 assertions/s.

308 runs, 521 assertions, 0 failures, 0 errors, 0 skips
```

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -99,12 +99,16 @@ module ShopifyAPI
 
     def request_token(params)
       return token if token
+      params = (params.respond_to?(:to_unsafe_hash) ? params.to_unsafe_hash : params).with_indifferent_access
+      unless params[:timestamp].to_i > 24.hours.ago.utc.to_i
+        raise ShopifyAPI::ValidationException, "Invalid timestamp: Possible malicious login"
+      end
 
-      unless self.class.validate_signature(params) && params[:timestamp].to_i > 24.hours.ago.utc.to_i
+      unless self.class.validate_signature(params)
         raise ShopifyAPI::ValidationException, "Invalid Signature: Possible malicious login"
       end
 
-      response = access_token_request(params[:code])
+      response = access_token_request(params['code'])
       if response.code == "200"
         self.extra = JSON.parse(response.body)
         self.token = extra.delete('access_token')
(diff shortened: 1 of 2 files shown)


tanema:
I appreciate the need for this, and I really like that you have split out an exception for the timestamp being wrong but I wonder if you should use `symbolize_keys` instead of `with_indifferent_access`. Within the codepaths of the session, it is just expecting symbols and has no use for string keys.

rcapozzi-shopify:
Maybe the fix is to document that "This API requires symbols for any hash arg". Along those lines, add a API_STRICT_HASH_KEY_ARGS to the class if you need a phase in period. The "are keys strings or symbols" has lived too long. Let us commit to one and be done. You will note that the tests make lots of use of string keys.

PR #647, opened 2019-12-05
Enrich 4xx error message with error text from response.body

vbyno:
This PR makes 4xx error messages more informative.

Instead of having this:
```
ActiveResource::ResourceInvalid, "Failed.  Response code = 422.  Response message = Unprocessable Entity.
```
the message will contain the error from `response.body['error']`
```
ActiveResource::ResourceInvalid, "Failed.  Response code = 422.  Response message = Unprocessable Entity (Cannot cancel a paid and fulfilled order).
```
so that it would be easier to quickly identify what went wrong

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Enrich 4xx errors with error message from response body [#647](https://github.com/Shopify/shopify_api/pull/647)
+
 == Version 8.0.0
 
 * Api Version changes [#600](https://github.com/Shopify/shopify_api/pull/600)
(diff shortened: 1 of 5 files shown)


tanema:
One thing! Can you add a line to the changelog about your change.

vbyno:
@tanema , of course! I'm not sure if adding it to 8.0.0 section in Changelog is fine, since it is already released. Shall I create a new section (`8.0.1` ?) and bump the version, or just put the text without specifying the version?

tanema:
Yup, just the one line is good and we will add the version and the changelog title when we release the next version

PR #648, opened 2019-12-05
Support local graphql schema file

raydefy:
This patch improves the gems GraphQL efficiency two:
1. Reuse the scheme produced in previous calls
2. Keep a local copy of the schema's JSON (Optional)

NOTE: PR #525 takes a similar approach, but may offer safe multi-session support.

Problem
The current implementation of `ShopifyAPI::GraphQL.new` results in a download of the Shopify API GraphQL scheme for each call to `new`. The schema is 2.5MB. The data must be parsed into a GraphQL object. An example of the download/parse behaviour:
```ruby
client = ShopifyAPI::GraphQL.new
# wait 15 - 30 seconds as the Shopify server streams the JSON to the client, and client parses
client = ShopifyAPI::GraphQL.new
#  repeat everything from the top
```
The patch keeps the parsed schema as a class attribute. Therefore, the the network request is performed once.
```ruby
# calls server for JSON schema. Saves results into a class attribute
client = ShopifyAPI::GraphQL.session(shop.name, shop.token) 

# Uses the class attribute to avoid network request.
client = ShopifyAPI::GraphQL.session(shop.name, nil) # Uses class
```

### Store the schema on local filesystem
Further performance improvements can be made by storing a copy of the schema's JSON on the local filesystem. Set the  `ShopifyAPI::GraphQL.schema_file` class attribute to a file. If the file does not exist, it will be created when the schema downloads. If the file does exist, the class will load the file and parse its contents into the GraphQL schema.

Session 1: Download schema and keep a local copy (assumes file does not exist).
```ruby
ShopifyAPI::GraphQL.schema_file = "local_schema.json"
client = ShopifyAPI::GraphQL.session(shop.name, shop.token)
exit
```

Session 2: Use the local file. Avoid download.
```ruby
ShopifyAPI::GraphQL.schema_file = "local_schema.json"
client = ShopifyAPI::GraphQL.session(shop.name, shop.token)
```

## Limitations
There may be a need for a Mutex around two pieces. One is the schema download. The other is around the Session.active.


tanema:
Can you please add more to the PR description. Both the reason why you need this and how you implemented this.

swalkinshaw:
cc @Shopify/api-patterns-team

swalkinshaw:
Thanks for the great start on this @raydefy. Before reviewing the code in depth my main concern is around the API being exposed. Currently we recommend people created a new `ShopifyAPI::GraphQL` client *after* the usual session management:

```ruby
shopify_session = ShopifyAPI::Session.new(domain: "SHOP_NAME.myshopify.com", token: token, api_version: api_version, extra: extra)
ShopifyAPI::Base.activate_session(shopify_session)

client = ShopifyAPI::GraphQL.new

SHOP_NAME_QUERY = client.parse <<-'GRAPHQL'
  {
    shop {
      name
    }
  }
GRAPHQL

result = client.query(SHOP_NAME_QUERY)
result.data.shop.name
```

Right now with this PR it would be:

```ruby
client = ShopifyAPI::GraphQL.session(domain: "SHOP_NAME.myshopify.com", token: token)

SHOP_NAME_QUERY = client.parse <<-'GRAPHQL'
  {
    shop {
      name
    }
  }
GRAPHQL

result = client.query(SHOP_NAME_QUERY)
result.data.shop.name
```

This might be confusing introducing another thing named `session` which is a wrapper around the existing one; except it also has less functionality right now. Is it possible to keep the existing functionality by just instantiating a client via `new`? Or maybe introduce a new method like `client_from_session(shopify_session)`.

My other question is for the `schema_file` attribute. It's overloaded with two concerns right now:
1. local filename to *save* when the dynamic schema is fetched
2. local filename to *use* as the cached schema

Two separate attributes might be clearer. Something like:

```ruby
ShopifyAPI::GraphQL.save_schema_file = true
ShopifyAPI::GraphQL.schema_file_path = 'schema.json'
```

This provides an explicit opt-in/out.

swalkinshaw:
See https://github.com/Shopify/shopify_api/pull/525

PR #670, opened 2020-01-18
Added options hash to create_permission_url and makes redirect_uri required

jdjkelly:
Fixes https://github.com/Shopify/shopify_api/issues/466

A re-attempt at the solution first investigated in https://github.com/Shopify/shopify_api/pull/467

- Makes a **breaking change** in requiring `redirect_uri` to `create_permissions_url`, since this is actually required today
- Adds an options hash so that we don't break the interface again, which for now supports passing `state`
- Updates readme

cc @jamiemtdwyer @richter-alex let me know your thoughts


tanema:
If you could rebase with master, I can merge this in.

jdjkelly:
> If you could rebase with master, I can merge this in.

Done

Issue #624, opened 2019-09-30
Relative cursor based pagination across page loads


dnswus:
The old **Page based pagination** will be deprecated in the `2019-10` API version in favor of the **Relative cursor based pagination**. I'm writing a Rails app with this library and have problem navigating products in multiple pages using **Relative cursor based pagination**. The example given using the `fetch_next_page` call only works in the same page load. What is the best way to navigate different pages across multiple page loads?

ziaulrehman40:
How are we supposed to request say 10th page from the API? I am sorry, didn't quiet get how to use this new cursor approach.

ziaulrehman40:
I digged a little more, now i understand we can retrieve next_page_info(which seems like a hash of some sort) and previous_page_info, how are we supposed to use these hashes to query that specific page when lets say user clicks next page button.

Also my first comment stands that if we can query nth page on new api version?

Thanks.

ziaulrehman40:
Please close this, i was able to make it work to fetch next and previous page using `page_info` param. Only issue was i was including other filter params as well while providing the `page_info`, using `httplog` to inspect responses actually displayed error and i was able to fix it by removing those additional param, as they can be provided only on first request.

Though getting nth page is still a mystery. I don't think with latest pagination scheme, that would be possible. But i can think of some usecases where that would help reduce API calls.

patricklindsay:
For anyone coming here looking for a basic example of how to accomplish this;

View:
```
%nav
  %ul.menu.align-center
    %li
      - if @resource_collection.previous_page?
        = link_to "Previous Page", page_info: @resource_collection.previous_page_info
      - else
        %button.button.disabled.clear Previous Page
    %li
      - if @resource_collection.next_page?
        = link_to "Next Page", page_info: @resource_collection.next_page_info
      - else
        %button.button.disabled.clear Next Page
```

Controller:
```
@resource_collection = ShopifyAPI::Product.find(:all, {page_info: params[:page_info]})
```

asecondwill:
This does not seem to work.  On hitting next page, i get same page.  Any clues?

```
  <% if @products.previous_page? %>
        <li class='page-item'>
        <%= link_to "Previous Page", page_info: @products.previous_page_info, class: ' page-link' %>
        </li>
      <% else  %>
        <li class='page-item disabled'>
          <a class="page-link  ">Previous Page</a>
        </li>
      <% end %>

      <% if @products.next_page? %>
        <li class='page-item'>
[...]

PR #671, opened 2020-01-20
improve documentation for cursor based pagination

khier996:
Add a bit more explanation to cursor based pagination docs so that people are not confused like in issue [624](https://github.com/Shopify/shopify_api/issues/624)


khier996:
I signed CLA. how do I restart CLA check?

PR #672, opened 2020-01-20
Improved GraphQL client

swalkinshaw:
This is a complete rewrite of the GraphQL client functionality.

The previous implementation suffered from a few problems making it virtually unusable as detailed in
https://github.com/Shopify/shopify_api/issues/511. Here's a summary:

1. you couldn't specify a local schema file
2. due to the above, the client made a dynamic introspection request on initialization which was very slow
3. unbounded memory growth due to building new clients at runtime

This rewrite was focused on solving those problems first and foremost but we also had a few other goals:

* support API versioning
* provide better defaults and an improved developer experience
* ensure it's hard to do the wrong thing

The new GraphQL client *only* supports loading local schema files to ensure no introspection requests are made during app runtime. The goal is that clients are fully initialized at application boot time (and if you're using Rails this is handled automatically).

Workflow:
1. Set `ShopifyAPI::GraphQL.schema_location` to a directory path (or use the default in Rails of `db/shopify_graphql_schemas`).
2. Save a JSON version of Shopify's Admin schema locally (or use the `shopify_api:graphql:dump` Rake task) to the `schema_location` and name it after the API version: `2020-01.json`.
3. Access the client at `ShopifyAPI::GraphQL.client`
4. Execute queries via `client.query`

### Rails integration
All the new features are designed to work independently of Rails but there are a few additional integration points:

* `schema_location` is set to be Rails specific within `db/`
* the Rake task is loaded automatically with the `environment` task prerequisite
* clients are initialized during Rails initialization

## Migration
If anyone is using the current GraphQL client it's very easy to migrate to this new one.

Previously a client was initialized with `ShopifyAPI::GraphQL.new`:
```ruby
client = ShopifyAPI::GraphQL.new

SHOP_NAME_QUERY = client.parse <<-'GRAPHQL'
  {
    shop {
      name
    }
  }
GRAPHQL

result = client.query(SHOP_NAME_QUERY)
result.data.shop.name
```

Now you use `ShopifyAPI::GraphQL.client` which has already been initialized:
```ruby
SHOP_NAME_QUERY = ShopifyAPI::GraphQL.client.parse <<-'GRAPHQL'
  {
    shop {
      name
    }
  }
GRAPHQL

result = ShopifyAPI::GraphQL.client.query(SHOP_NAME_QUERY)
result.data.shop.name
```

### Todo
- [x] Update README


lib/shopify_api/graphql.rb
```diff
@@ -0,0 +1,80 @@
...
+        cached_client = @_client_cache[api_version]
+
+        if cached_client
+          cached_client
+        else
+          schema_file = schema_location.join("#{api_version}.json")
+
+          if !schema_file.exist?
+            raise InvalidClient, <<~MSG
+              Client for API version #{api_version} does not exist because no schema file exists at `#{schema_file}`.
```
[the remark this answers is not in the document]
swalkinshaw: Yep 🚀

lib/shopify_api/graphql/task.rake
```diff
@@ -0,0 +1,54 @@
...
+    prereqs = []
+    # add the Rails environment task as a prerequisite if loaded from a Rails app
+    prereqs << :environment if Rake::Task.task_defined?('environment')
+
+    desc 'Writes the Shopify Admin API GraphQL schema to a local file'
+    task dump: prereqs do
+      site_url = ENV['SITE_URL'] || ENV['site_url']
+      shop_domain = ENV['SHOP_DOMAIN'] || ENV['shop_domain']
+      api_version = ENV['API_VERSION'] || ENV['api_version']
+      access_token = ENV['ACCESS_TOKEN'] || ENV['access_token']
```
tanema: All of these environment variables are not specified in your error or where to get the values. Maybe this should be part of the README update as well.

swalkinshaw:
Updated with:

* README changes
* Full usage output for the rake task
* Documentation guide in `docs/graphql.md`

👀 appreciated

docs/graphql.md
```diff
@@ -0,0 +1,191 @@
...
+So the first step in making GraphQL queries is having a local JSON file of Shopify's Admin schema.
+This gem provides a `shopify_api:graphql:dump` Rake task to make it as easy as possible:
+
+```bash
+$ rake shopify_api:graphql:dump SHOP_URL="https://API_KEY:PASSWORD@SHOP_NAME.myshopify.com" API_VERSION=2020-01
+```
+
+If successful `db/shopify_graphql_schemas/2020-01.json` will be created.
+
+You can either use private app authentication or an Oauth access token. Run `rake shopify_api:graphql:dump`
```
RobertWSaunders: "Oauth" --> "OAuth"

lib/shopify_api/graphql/task.rake
```diff
@@ -0,0 +1,86 @@
...
+
+      shopify_session = ShopifyAPI::Session.new(domain: shop_domain, token: access_token, api_version: api_version)
+      ShopifyAPI::Base.activate_session(shopify_session)
+
+      if shop_url
+        ShopifyAPI::Base.site = shop_url
+      end
+
+      schema_location = ShopifyAPI::GraphQL.schema_location
+      FileUtils.mkdir_p(schema_location) unless Dir.exist?(schema_location)
```
alanhill: `ShopifyAPI::GraphQL.schema_location` is set here https://github.com/Shopify/shopify_api/pull/672/files#diff-49a342381a8e9c693123cb5d3ad9fea1R65 which is using https://github.com/Shopify/shopify_api/pull/672/files#diff-49a342381a8e9c693123cb5d3ad9fea1R7 `DEFAULT_SCHEMA_LOCATION_PATH` if the schema location is not specifically set, which results in the directory being created in the root of the app, opposed to the `db` folder. Not sure if intentional?

alanhill: I also noticed that when fetching the schema, if we get a 401 we overwrite the current schema with the error json response. Not sure if that's intentional?

swalkinshaw: No, good catch! We definitely shouldn't dump if it errors so I'll fix that.

swalkinshaw: Updated 👍

swalkinshaw:
@alanhill helped 🎩 this and got it all working upgrading a real app 🎉 

With his help we found some issues with the rake task so they've all been fixed now.

PR #676, opened 2020-01-28
Update graphql.md

chrisbutcher:
What do you think about just highlighting the two types of authentication the rake task supports in the doc?


katiedavis:
I'm going to close this as we merged [689](https://github.com/Shopify/shopify_api/pull/689) yesterday which covers the same thing.

PR #677, opened 2020-01-29
Take into account semantical 'errors' key in JSON body for 422 responses

vbyno:
According to the [documentation](https://help.shopify.com/en/api/getting-started/response-status-codes), there might be `errors` key containing an array of error messages. In this PR we check if the response body has `errors` key if we didn't found `error`

### DIFF
--- a/lib/shopify_api/message_enricher.rb
+++ b/lib/shopify_api/message_enricher.rb
@@ -5,7 +5,13 @@ module ShopifyAPI
 
       @_cached_message ||= begin
         detailed_error = begin
-          JSON.parse(body)['error'].to_s
+          parsed_body = JSON.parse(body)
+
+          if parsed_body['error']
+            parsed_body['error'].to_s
+          elsif parsed_body['errors']
+            Array(parsed_body['errors']).join('; ')
+          end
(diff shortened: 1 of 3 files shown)


lib/shopify_api/message_enricher.rb
```diff
@@ -1,11 +1,17 @@
 module ShopifyAPI
   class MessageEnricher < SimpleDelegator
     def message
-      return super unless (400...500).include?(code.to_i)
+      return super if code.to_i != 422
```
tanema: I would prefer to keep this within the whole range, 422 may be the only one we support currently by the docs but that doesnt mean we won't be adding it in the future.

vbyno:
> One thing we could do is call super if errors are undefined or the JSON is invalid to capture all cases.

@tanema , It looks to me that is already implemented here
https://github.com/Shopify/shopify_api/pull/677/files#diff-0c1174150b76ca5c6e7e2a5513559edfR15-R20

tanema:
I think it will be a little while before this is deployed and added to our other gems because we are currently propagating a new version right now

vbyno:
@tanema , no problem. Thanks!

PR #685, opened 2020-02-18
warning raised for version eol

toneymathews:
Closes https://github.com/Shopify/shopify/issues/233879

TL;DR  This pull request is raised to add a warning message case of API Version being used to make a call is expired or will expire within 30 days

**Background:**

Currently, a warning header is added to the responses of all admin and storefront APIs that are using a version that expires within 30 days or is expired already. This was done through https://github.com/Shopify/shopify/pull/232580

Now, we want to raise warning messages through the Shopify API gem as well if API calls are made using expired or EOL versions. 

**Solution**

The response headers of the API responses are checked for the presence of the version EOL warning header. If present, a warning message is raised.

### DIFF
--- a/lib/active_resource/detailed_log_subscriber.rb
+++ b/lib/active_resource/detailed_log_subscriber.rb
@@ -3,6 +3,7 @@ module ActiveResource
     def request(event)
       log_request_response_details(event)
       warn_on_deprecated_header(event)
+      warn_on_version_eol_warning_header(event)
     end
 
     def logger
@@ -30,6 +31,22 @@ module ActiveResource
           warning_message = <<-MSG
           [DEPRECATED] ShopifyAPI made a call to #{payload[:method].upcase} #{payload[:path]}, and this call made
           use of a deprecated endpoint, behaviour, or parameter. See #{header_value} for more details.
+          MSG
+
+          warn warning_message
+        end
+      end
+    end
+
+    def warn_on_version_eol_warning_header(event)
+      payload = event.payload
+
+      payload[:response].each do |header_name, header_value|
+        case header_name.downcase
+        when 'x-shopify-api-version-warning'
+          warning_message = <<-MSG
+          [Version EOL Warning] ShopifyAPI made a call to #{payload[:method].upcase} #{payload[:path]}, and this call made
+          use of an API version that is expired or will expire within 30 days. See #{header_value} for more details.
           MSG
 
           warn warning_message
--- a/lib/active_resource/detailed_log_subscriber.rb
+++ b/lib/active_resource/detailed_log_subscriber.rb
@@ -38,8 +38,8 @@ module ActiveResource
 
         when VERSION_EOL_WARNING_HEADER
           warning_message = <<-MSG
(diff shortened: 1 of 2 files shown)


lib/active_resource/detailed_log_subscriber.rb
```diff
@@ -36,6 +37,22 @@ def warn_on_deprecated_header(event)
         end
       end
     end
+
+    def warn_on_version_eol_warning_header(event)
+      payload = event.payload
+
+      payload[:response].each do |header_name, header_value|
+        case header_name.downcase
+        when 'x-shopify-api-version-warning'
```
chrisbutcher: We are now doing a fair amount of duplicated logic between `warn_on_version_eol_warning_header` and `warn_on_deprecated_header`. Can you see a way to iterate the response headers only once, and ensure that we detect both, one or neither response header if present?

chrisbutcher:
cc: @Shopify/api-patterns-team

lib/active_resource/detailed_log_subscriber.rb
```diff
@@ -21,17 +23,25 @@ def log_request_response_details(event)
...
           warning_message = <<-MSG
           [DEPRECATED] ShopifyAPI made a call to #{payload[:method].upcase} #{payload[:path]}, and this call made
           use of a deprecated endpoint, behaviour, or parameter. See #{header_value} for more details.
           MSG
 
+          warn warning_message
+
+        when VERSION_EOL_WARNING_HEADER
+          warning_message = <<-MSG
+          [Version EOL Warning] ShopifyAPI made a call to #{payload[:method].upcase} #{payload[:path]}, and this call made
```
swalkinshaw: "and this call uses an API version that is expired..."

swalkinshaw: We shouldn't use EOL without defining it and I'm not even sure we publicly use that term.

@chrisbutcher isn't it just "unsupported"? We just need to make sure this lines up with our public versioning terminology.

lib/active_resource/detailed_log_subscriber.rb
```diff
@@ -21,17 +23,25 @@ def log_request_response_details(event)
...
+          use of an API version that is expired or will expire within 30 days. See #{header_value} for more details.
```
grcooper: Nit: `made use of` -> `used`

lib/active_resource/detailed_log_subscriber.rb
chrisbutcher: Yes, the term we should use publicly is `unsupported`: https://shopify.dev/concepts/about-apis/versioning

test/detailed_log_subscriber_test.rb
```diff
@@ -114,4 +114,26 @@ def set_logger(logger)
       @logger.logged(:warn).first
     )
   end
+
+  test "warns when the server responds with a x-shopify-api-version-warning" do
```
chrisbutcher: add term `header`

lib/active_resource/detailed_log_subscriber.rb
chrisbutcher:
```suggestion
          use of an API version that is unsupported or will become unsupported within 30 days. See #{header_value} for more details.
```

PR #700, opened 2020-03-09
Require the gem in installation

### LINKED ISSUE #699: Readme file issues
Step 3 in your readme.md file doesn't mention that you need to add
require 'shopify_api' 
as the first line.

Also, you say we have to set ShopifyAPI::Base.site to the base URL but the readme file implies that it doesn't need to be in quotes.  Should it be in quotes?

ShayneP:
Resolves #699

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,12 @@ Or install via [gem](http://rubygems.org/)
 gem install shopify_api
 ```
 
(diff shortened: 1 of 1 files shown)


katiedavis:
Thanks Shayne!

PR #712, opened 2020-04-05
Fix `mocha/setup` deprecation warning

andyw8:
Prevents this warning when running the tests:

    Mocha deprecation warning at ... test/test_helper.rb:5:in `require':
    Require 'mocha/test_unit', 'mocha/minitest' or 'mocha/api' instead of 'mocha/setup'.

Requiring `mocha/setup` was deprecated in 1.10.0:

https://github.com/freerange/mocha/blob/master/RELEASE.md#1100


tanema:
Thanks for the fix!

PR #716, opened 2020-04-15
Bump version to 9.1.0

ragalie:


lib/shopify_api/version.rb
```diff
@@ -1,3 +1,3 @@
 module ShopifyAPI
-  VERSION = "9.0.4"
+  VERSION = "9.1.0"
```
emmack, line 2: What determines whether it goes `9.1.0` vs `9.0.5`? I'm not super familiar to versioning semantics.

ragalie: I've been following semvar (and assuming everyone else is too :p): https://semver.org/

This one's a little tricky. Theoretically someone could be relying on the default equality implementation, which is based on object identity. But I have no evidence that anyone's doing that, and it's hard to imagine it being useful. So I'm taking the position that this is _added_ functionality (and not backwards-incompatible).

Since this is added, backwards-compatible functionality, it's a minor version bump.

tanema: Major = Breaking changes
Minor = New features
Patch = bug fixes

ragalie:
No problem, @tanema, just let me know when it's good to go and I'll merge it!

tanema:
Good to go!

PR #722, opened 2020-04-22
Update clear_session not to clear the api-version

### LINKED ISSUE #640: clear_session clears the API version
I was investigating a Bugsnag issue for `You must set ShopifyAPI::Base.api_version before making a request.` and noticed that calling `ShopifyAPI::Base.clear_session` will reset the `api_version` to `null`:

https://github.com/Shopify/shopify_api/blob/24d2640c04fead8871097f23f5c1dadd4a9c298f/lib/shopify_api/resources/base.rb#L52-L58

My application was using a Rails initializer to set the version as hinted at by the [Getting Started](https://github.com/Shopify/shopify_api/blob/master/README.md#getting-started) notes so this error caught me by surprise.

Is this expected behaviour?

adfaris:
resolves #640 

Removes resetting the api_version when clearing the session. In the `test\base_test.rb` there is no logic the shows the session clearing should also clear the api_version.
@tanema

### DIFF
--- a/lib/shopify_api/resources/base.rb
+++ b/lib/shopify_api/resources/base.rb
@@ -53,7 +53,6 @@ module ShopifyAPI
         self.site = nil
         self.password = nil
         self.user = nil
-        self.api_version = nil
         self.headers.delete('X-Shopify-Access-Token')
       end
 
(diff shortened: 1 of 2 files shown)


tanema:
Thanks again!

Issue #729, opened 2020-05-11
Add utility for parsing bulk operations


nickmealey:
Since the output of bulk operations from GraphQL doesn't nest connections within each object, it can be tedious to correctly parse the file with the desired outcome. According to the docs, we can rely on the fact that children connections will always be listed _after_ their parent, not _directly after_. A tip I picked up from @eapache, looping through the file in reverse order can make reading the data reliable with efficient memory loading.

Here's an example of how nested connections are added:
```
{"id":"gid:\/\/shopify\/Customer\/123456"}
{"id":"gid:\/\/shopify\/Metafield\/987654","key":"some_key","__parentId":"gid:\/\/shopify\/Customer\/123456"}
```

The ideal output:
```
{"id":"gid:\/\/shopify\/Customer\/123456","metafields":{"edges":[{"node":{"id":"gid:\/\/shopify\/Metafield\/987654","key":"some_key"}}]}}
```

My suggestion is that we build a helper utility to parse through the data in reverse order and nest any connections within a hash. This would make it easier for developers to quickly handle the data in an intuitive way.

PR #730, opened 2020-05-11
Add bulk operation utility

nickmealey:
This pull request adds a utility for parsing bulk requests with nested connections inline with the parent object.

See `tests/bulk_operation_util_test.rb` for examples.

See issue #729 for more information.

Updated Readme and bumped version number

### DIFF
--- a/README.md
+++ b/README.md
@@ -461,6 +461,26 @@ jsonl.each_line do |line|
 end
 ```
 
+Take note that nested connections are listed on their own line in the file and although guaranteed to follow the parent resource in order, it's not a guarantee it will located directly after that resource. A trick is to read the file in reverse, so that any children nodes are tracked before the parent node is discovered. You can use the `BulkOperationUtil` class to parse and output nodes along with their children nested within each node:
+```
+url = "https://storage.googleapis.com/shopify/k524jdq...json"
+
+# Only output parent nodes with nested connections, in reverse order
+ShopifyAPI::BulkOperationUtil.with_nested_connections(url) do |node|
+  # A `__children` property is added to the parsed result, which will contain
+  # any child nodes that reference the node from "__parentId"
+  puts node["__children"]
+end
+```
+
+Using the JSON example above, this parses an output as such:
+```
+{"id"=>"gid://shopify/Product/1921569357880", "__children"=>[{"id"=>"gid://shopify/ProductVariant/19435459117112", "title"=>"47"}]}
+...
+{"id"=>"gid://shopify/Product/1921569226808", "__children"=>[{"id"=>"gid://shopify/ProductVariant/19435458986123", "title"=>"52"},
+                                                             {"id"=>"gid://shopify/ProductVariant/19435458986040", "title"=>"70"}]}
+```
+
 ## Pagination
 
 Shopify uses [Relative cursor-based pagination](https://shopify.dev/tutorials/make-paginated-requests-to-rest-admin-api) to provide more than a single page of results. 
--- a/README.md
+++ b/README.md
@@ -462,7 +462,7 @@ end
 ```
 
 Take note that nested connections are listed on their own line in the file and although guaranteed to follow the parent resource in order, it's not a guarantee it will located directly after that resource. A trick is to read the file in reverse, so that any children nodes are tracked before the parent node is discovered. You can use the `BulkOperationUtil` class to parse and output nodes along with their children nested within each node:
-```
+```ruby
 url = "https://storage.googleapis.com/shopify/k524jdq...json"
 
 # Only output parent nodes with nested connections, in reverse order
@@ -474,7 +474,7 @@ end
 ```
 
 Using the JSON example above, this parses an output as such:
-```
+```ruby
 {"id"=>"gid://shopify/Product/1921569357880", "__children"=>[{"id"=>"gid://shopify/ProductVariant/19435459117112", "title"=>"47"}]}
 ...
 {"id"=>"gid://shopify/Product/1921569226808", "__children"=>[{"id"=>"gid://shopify/ProductVariant/19435458986123", "title"=>"52"},
--- a/README.md
+++ b/README.md
@@ -466,10 +466,17 @@ Take note that nested connections are listed on their own line in the file and a
 url = "https://storage.googleapis.com/shopify/k524jdq...json"
 
 # Only output parent nodes with nested connections, in reverse order
-ShopifyAPI::BulkOperationUtil.open_with_nested_connections(url) do |node|
+ShopifyAPI::BulkOperationUtil.open_with_nested_connections(url) do |node, index|
   # A `__children` property is added to the parsed result, which will contain
   # any child nodes that reference the node from "__parentId"
   puts node["__children"]
+
+  # Note: since this parses the data in reverse, the index will also be reversed,
+  # and not correlate one-to-one per line, but rather per parent resource (i.e. 
+  # the last line will be index 0 and work forward toward the beginning).
+  # In most cases, the order of the line number isn't significant, but helpful to
+  # track which nodes you've processed.
+  puts "Index: #{index}"
 end
 ```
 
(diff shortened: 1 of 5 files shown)


nickmealey:
I've since signed the CLA

tanema:
I appreciate the work however I don't think we want to include utility classes like this within this library. This would be something that you would add to your own application to make integrating with the ShopifyAPI library easier for your application.

nickmealey:
> I appreciate the work however I don't think we want to include utility classes like this within this library. This would be something that you would add to your own application to make integrating with the ShopifyAPI library easier for your application.

@tanema 

Thanks for taking a look. Is there an appropriate place for something like this to live within the library? I'd like to make the case that the functionality here provides a much simpler interface for developers when working with the Bulk API. We've already implemented this into our own app as a utility. I realize it's important for this not to become a "dumping ground" of code, which I certainly can agree with. However, the reason libraries like this exist is to make working with an API simpler and quicker. After all, you don't need this gem at all in order to use the Shopify API with Ruby, but it sure is helpful.

The reason that the Bulk API exports connected data like it does it because of memory constraints, especially for mobile. However, in the context of this gem, these memory constraints are far less of a concern since the devices that host our apps have a well known memory amount. In the vast majority of cases, using the bulk API with connected nodes means you want to associate each child node directly with its parent. In fact, I can't really see a use case where you request a child node and not directly associate it with its parent. This takes a fair amount "glueing together" currently, and I think this library is a great place for a helpful tool as this to do that for you. In most cases, developers are going to have to implement something like this themselves anyway.

Possibly, the Bulk API deserves its own class. I know that currently we're relying on the graphql gem to handle most of resource logic, so I'm not exactly sure how this fits in. We had to do a fair amount of setup to properly validate queries being passed to the `query` param of the request. The Bulk API is very powerful and unlocks a lot of potential that was previously very slow with a per-request action, and that's why I'd love to see it be used in a simple reliable way.

Anyway, thanks for considering and understood if we just close this one, but I do think a variation of it could be a great service to other developers.

eapache:
cc @grcooper @swalkinshaw

grcooper:
I think one thing to consider is what is the core problem you are trying to solve, and how do we make a solution that doesn't make it too difficult for consumers to switch.

My initial thoughts are that the problem is (from your issue linked above) is that you don't know when the LAST object on an object is around. This ends up meaning that an object may have to live in memory for a long time while you add child objects to it.

Is this fixed by instead of nesting, to just make sure the lines are grouped correctly? ie, instead of nesting, just put all of the objects after a parent object?

You end up with a psuedo-nesting structure, and for objects with a large amount of nested objects, you don't need to load a large blob of JSON.


This also lends a familiar way of parsing the objects similar to the actual bulk API, just with an added layer of niceness, rather than making consumers change how they think about the Bulk API when working with the gem vs raw requests.

ShayneP:
> Anyway, thanks for considering and understood if we just close this one, but I do think a variation of it could be a great service to other developers.

I definitely don't think we should close this one. There's a lot of value here, we just need to figure out exactly where to put it ...

nickmealey:
@grcooper 

> I think one thing to consider is what is the core problem you are trying to solve, and how do we make a solution that doesn't make it too difficult for consumers to switch.

The core problem to solve here (at least from my perspective 😄) is that the Bulk API produces results that are inconsistent with the GraphQL API responses. I understand there are performance reasons (and likely others) for formatting the data the way it does. Being on the other side of the API itself, the best solution I see is to consume the response from the results and format it in such a way that makes clear sense of the data.

>My initial thoughts are that the problem is (from your issue linked above) is that you don't know when the LAST object on an object is around. This ends up meaning that an object may have to live in memory for a long time while you add child objects to it.

Yes. From a good tip from @eapache, parsing the file in reverse means you can load the least amount of data into memory because you know the moment you hit a line that has no `__parent_id`, there's no more children to discover and Mr. GC can do his work.

>Is this fixed by instead of nesting, to just make sure the lines are grouped correctly? ie, instead of nesting, just put all of the objects after a parent object?

If the API were to guarantee all nested children followed a parent on new lines _directly_ following the parent, this utility may not be as important (although I still see if helpful). It sounds like this would come with a performance impact on the API's end though, which sounds like a poor tradeoff.

---

My main motivation behind this is to make working with the Bulk API closer to working with a standard GraphQL response. Since there are limitations beyond my control on the API itself, a "post-processor" seems to make sense. Here's a compressed example:

Some basic GraphQL request:
```
{
  customers(first: 10) {
    edges {
      node {
        id
        metafields(first: 1) {
          edges {
            node {
              key
              namespace
              value
            }
          }
        }
      }
    }
  }
}
```

Response:
```
{
  "data": {
    "customers": {
      "edges": [
        {
          "node": {
            "id": "gid://Shopify/customer/571302832",
            "metafields": {
              "edges": [
                {
                  "node": {
                    "key": "says",
                    "namespace": "gandalf",
                    "value": "Don't you leave him Samwise Gamgee"
                  }
                }
              ]
            }
          }
        },
        ....
      ]
    }
  },
  ...
}
```

The same query, with the Bulk API:
```
    mutation {
      bulkOperationRunQuery(
        query: """
          customers(first: 10) { ... }
        """
      ) {
        bulkOperation {
          id
          status
        }
        userErrors {
          message
        }
      }
    }
```

But here's how the Bulk API might respond:
```
{ "id": "gid://Shopify/customer/571302832" }
{ "id": "gid://Shopify/customer/523432812" }
{ "id": "gid://Shopify/customer/984324394" }
{ "id": "gid://Shopify/metafield/984324394", "key": "says", "namespace": "gandalf", "value": "Don't you leave him Samwise Gamgee", "__parent_id": "gid://Shopify/customer/571302832" }
{ "id": "gid://Shopify/metafield/53482702423", "key": "says", "namespace": "samwise", "value": "I'm coming with you Mr. Frodo", "__parent_id": "gid://Shopify/customer/523432812" }
{ "id": "gid://Shopify/metafield/1023890323", "key": "says", "namespace": "frodo", "value": "Go home, Sam", "__parent_id": "gid://Shopify/customer/984324394" }
```

My first reaction was to just load the entire file into memory, locate each child, and process each customer. Fine for small example like this, but not a 2 million line JSON file. I'm inclined to think others might make that mistake. There are two challenges here:

**Challenge #1: Setting up a dynamic and syntactically valid `query`**

In our case, we needed to add variables to filter customers with a certain tag. You could interpolate this, but then you're not getting the benefit of GraphQL's client validation. So, we setup a method to generate a valid query. This will fail client side if the query is invalid (better than finding out from a `userError`):

```ruby
# Setup the request
BULK_REQUEST = ShopifyAPI::GraphQL.client.parse <<~GRAPHQL
  mutation($query: String!) {
    bulkOperationRunQuery(
      query: $query
    ) {
      bulkOperation {
        id
        status
      }
      userErrors {
        message
      }
    }
  }
GRAPHQL

# Dynamic query that is parsed client side
def bulk_query(tags: "")
  ShopifyAPI::GraphQL.client.parse <<~GRAPHQL
    {
      customers(after: null, query: "tags:#{tags}") {
        edges {
          node {
            id
            metafields { ... }
          }
        }
      }
    }
  GRAPHQL
end

# Run the query
ShopifyAPI::GraphQL.client.query(BULK_REQUEST, variables: {
  query: bulk_query(tags: "hobbit").document.to_query_string
})
```

I think there's an opportunity to create a class that lets you create these bulk requests and validated like this example client side. Maybe we just document the concept. I know dynamic queries are discouraged, but seeing that the `query` param is a string, I don't see a way around it.

**Challenge #2: Parsing the data correctly**
I think we've covered this plenty, but as a consumer I need to know which metafields exist for a customer, as soon as possible into the file read. If I don't have to think about it just like I don't have to think about it with a standard request, I'd say that's a win.

Sorry to be long winded 😆 I'm trying to clearly pass along my ideas to make this easier for everyone. I might conceptualize what a dedicated class might look like for this.

PR #731, opened 2020-05-14
Inventory param patch fix

marschattha:
This is a duplicate of https://github.com/Shopify/shopify_api/pull/655 but with the issue fixed, I posted the fix on that thread but no one paid attention for so long.

### DIFF
--- a/lib/shopify_api/resources/product.rb
+++ b/lib/shopify_api/resources/product.rb
@@ -15,6 +15,15 @@ module ShopifyAPI
         format % prices.min
       end
     end
+    
+    def total_inventory=(new_value)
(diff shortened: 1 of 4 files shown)


tanema:
This is great however it seems there is still a legitimate test failure and you need to sign the CLA

marschattha:
> This is great however it seems there is still a legitimate test failure and you need to sign the CLA

Sorry, just copied the tests from 9.0.3 version but they required some changes to work, fixed.

marschattha:
Can we push this out into a version onto ruby gems?
The deadline for removal of inventory_quantity params is coming soon.

PR #746, opened 2020-06-04
Prevent README and public docs inconsistencies

andyw8:
co-authored with @will-r-wang

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -21,7 +21,14 @@ rescue LoadError
   end
 end
 
-task :default => :test
+task :default => [:test, :verify_docs]
+
+require 'verify_docs'
+task :verify_docs do
+  unless VerifyDocs.call
+    abort("\nWARNING: docs/index.md and README.md no longer have identical content. Please correct this.")
+  end
+end
 
 require 'rdoc/task'
 Rake::RDocTask.new do |rdoc|
--- a/docs/index.md
+++ b/docs/index.md
@@ -106,7 +106,7 @@ For a private App you just need to set the base site url as follows:
    ```ruby
    shop_url = "https://#{API_KEY}:#{PASSWORD}@#{SHOP_NAME}.myshopify.com"
    ShopifyAPI::Base.site = shop_url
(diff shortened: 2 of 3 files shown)


docs/index.md
```diff
@@ -106,7 +106,7 @@ For a private App you just need to set the base site url as follows:
    ```ruby
    shop_url = "https://#{API_KEY}:#{PASSWORD}@#{SHOP_NAME}.myshopify.com"
    ShopifyAPI::Base.site = shop_url
-   ShopifyAPI::Base.api_version = '<version_name>' # find the latest stable api_version [here](https://shopify.dev/concepts/about-apis/versioning)
+   ShopifyAPI::Base.api_version = '<version_name>' # find the latest stable api_version here: https://shopify.dev/concepts/about-apis/versioning
```
andyw8, line 109: (This was already inconsistent - so fixing it here).

lib/verify_docs.rb
```diff
@@ -0,0 +1,7 @@
+class VerifyDocs
+  def self.call
+    readme_content = File.read("README.md").lines[2..-1]
+    docs_content = File.read("docs/index.md").lines[4..-1]
```
andyw8, line 4: The absolute line indices make this somewhat fragile but I think it's sufficient for now.

Rakefile
```diff
@@ -21,7 +21,14 @@ rescue LoadError
   end
 end
 
-task :default => :test
+task :default => [:test, :verify_docs]
+
+require 'verify_docs'
```
paulomarg, line 26: Minor: should this be required only when the `:verify_docs` task is called?

lib/verify_docs.rb
paulomarg: One viable option for this would be to add a comment to both files indicating where the shared content begins (which also serves as a reminder for people editing it) and ignore anything before that comment.

andyw8:
@tanema Travis defaults to running `rake`, and we've added `verify_docs` to the default Rake task.

https://docs.travis-ci.com/user/languages/ruby/

PR #747, opened 2020-06-04
Allow for executing client to be set dynamically

RichardBlair:
### Preface

First I want to thank you all for implementing a more memory efficient GraphQL client. We are really excited to delete ours. I also really appreciate the light wrapper approach.

### Problem

Throughout our application we use the graphql api and one of the first things we noticed was that we needed to check the result object returned from the graphql client to determine if an error occurred. This wasn't something we liked scattered throughout our code and as a result we enhanced the client to parse errors and raise exceptions instead.

However, we appreciate this may not be the approach everyone wants to take. We also really like the lightweight approach. It better enables developers to instrument heuristics they prefer.

### Solution

In order to maintain flexibility, respect the light wrapper approach, and enable cleaner code I'd like to propose the ability to set a custom client implementation. As much as this will allow us to easily handle errors without monkey patching, I'm far more excited to use this in our tests.

### DIFF
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -147,6 +147,32 @@ during your boot process.
 The goal is to have all clients created at boot so there's no schema loading,
 parsing, or client instantiation done during runtime when your app serves a request.
 
+## Using a custom client
+`ShopifyAPI::GraphQL::HTTPClient` inherits from `::GraphQL::Client::HTTP` and instruments
+the headers, url and api version for you. However, you may find that you wish to add
+additional functionality to the client responsible for executing and parsing queries.
+For instance, you may wish to create a client which loads responses from a file for your
+tests, or you may want to implement a client which raises errors instead of passing them
+back through the results object.
+
+To set a custom client set `ShopifyAPI::GraphQL.client_klass` to your client:
+```ruby
+class RaisingHTTPClient < ShopifyAPI::GraphQL::HTTPClient
+  def execute(document:, operation_name: nil, variables: {}, context: {})
+    result = super
+    do_work(result)
+  end
+
+  private
+
+  def do_work(result)
+    result
+  end
+end
+
+ShopifyAPI::GraphQL.client_klass = RaisingHTTPClient
+```
+
 ## Migration guide
 Prior to shopify_api v9.0 the GraphQL client implementation was limited and almost
 unusable due to the client making dynamic introspection queries to Shopify's API.
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -147,15 +147,16 @@ during your boot process.
 The goal is to have all clients created at boot so there's no schema loading,
 parsing, or client instantiation done during runtime when your app serves a request.
 
-## Using a custom client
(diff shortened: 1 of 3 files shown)


RichardBlair:
Signed CLA as requested by this fancy bot.

swalkinshaw:
Thanks for bringing this up @RichardBlair. Completely agreed that we should offer a way to customize the client. There's two issues I can see after a quick look:

1. what you're calling the "client" is not that accurate since the graphql-client has some confusing concepts in my opinion. It's really the execute "adapter" as they call it. So letting people customize the adapter is useful, but still fairly limited since the adapter doesn't do that much ([ref](https://github.com/github/graphql-client/blob/2bbe6a5f360521e31d5be3397ddef7ca0205585c/lib/graphql/client/http.rb)).
2. we could go one step further and actually allowing customize the client itself instead of the default (`GraphQL::Client`). Maybe we can come up with a higher-level way to achieve both of these with a single configuration option. Or at a minimum we just allow both to be set separately: `client` and `execute_adapter`.

I'll give this another look tomorrow after I think on it more.

RichardBlair:
The verbiage is very sub-optimal, I agree. "You can configure the client's client" really doesn't roll off the tongue and is a bit brain melting. Even "execute" is a bit weird, all the thing does is provide an abstraction for IO. That said, deviating from "execute" will probably just cause more confusion so we should probably stick with that?

One potential issue I see with allowing the client to be set is that there will need to be a defined interface that the client needs to adhere to. At that point I'm not sure there is that much of a benefit to using this library, the developer is already half way to hand rolling this implementation anyway. If this is a none issue though, then I'm all for it.

Personally I like the idea of allowing both to be set. It provides the greatest flexibility while still allowing this codebase to evolve. I'll whip it up and if we decide to go the other way no worries!

docs/graphql.md
```diff
@@ -147,6 +147,77 @@ during your boot process.
 The goal is to have all clients created at boot so there's no schema loading,
 parsing, or client instantiation done during runtime when your app serves a request.
 
+
+## Using a custom GraphQL Client
+By default `ShopifyAPI::GraphQL` wraps the Github GraphQL Client library. However, this client
+may not suitable for various reasons. If you wish to expand on the interface of the client or
+improve the required functions for your use case you can implement a client of your own.
+
+The client must conform to the following interface:
```
swalkinshaw, line 156: I'm a bit wary of detailing this interface in docs since it could change. I'm not sure it's reasonable (or common) for people to implement their own client entirely that conforms to this, maybe we should just mention/expect people to subclass `::GraphQL::Client`? If someone really wants to go beyond that then they're still free to.

RichardBlair: I'd hope that if someone has gone to this extent that they wouldn't need the doc, but also know to inherit from `GraphQL::Client`, totally agree. I was going with the mindset that it was easier to do too much and remove it.

RichardBlair:
Hey gang, curious what the process is for getting this merged and if there is anything else I need to do on my end?

tanema:
Sorry I was waiting for CI to pass again but it seems stuck. I will try to kick it

PR #748, opened 2020-06-05
Default sessions to Base api_version

### LINKED ISSUE #743: Setting Default Version for Public/Custom Apps
This:

```rb
ShopifyAPI::Base.api_version = "2020-07"
```

Works for private apps but for public apps you have to explicitly provide it **every time** (at least via `.temp`). 

Why not use this as the default when none is given? Otherwise one has to go about storing a default version elsewhere when there's a seemingly viable alternative.

paulomarg:
Paired with @carmelal for this one, to fix #743. We changed the way sessions (and temp sessions) are created so that they default to `Base.api_version` if no `api_version` is given, so that developers don't need to pass in the version every time they need a session.

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -20,7 +20,7 @@ module ShopifyAPI
         params.each { |k,value| public_send("#{k}=", value) }
       end
 
-      def temp(domain:, token:, api_version: nil, &block)
+      def temp(domain:, token:, api_version: ShopifyAPI::Base.api_version, &block)
         session = new(domain: domain, token: token, api_version: api_version)
 
         with_session(session, &block)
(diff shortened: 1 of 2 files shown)


test/session_test.rb
```diff
@@ -19,13 +19,23 @@ def setup
...
 
   test "not be valid without an api version" do
-    session = ShopifyAPI::Session.new(domain: "testshop.myshopify.com", token: "any-token", api_version: nil)
-    assert_not session.valid?
-
     session = ShopifyAPI::Session.new(domain: "testshop.myshopify.com", token: "any-token", api_version: ShopifyAPI::ApiVersion::NullVersion)
     assert_not session.valid?
   end
 
+  test "default to base api version" do
```
andyw8, line 26: Minor: Let's use proper capitalization for acronyms such as API?

lib/shopify_api/session.rb
```diff
@@ -84,9 +84,9 @@ def extract_current_session
       end
     end
 
-    def initialize(domain:, token:, api_version:, extra: {})
+    def initialize(domain:, token:, api_version: nil, extra: {})
```
andyw8, line 87: Not critical, but we could provide the default in the definition, i.e.:

`def initialize(domain:, token:, api_version: ShopifyAPI::Base.api_version, extra: {})`

PR #750, opened 2020-06-12
Remove old shopify binary from API

paulomarg:
With the upcoming CLI v2 changes, a new `shopify` executable will be generated and distributed, which will overwrite the current binary from this repo.

This is what happens when the current binary is executed:
```
% shopify
shopify command is no longer bundled with shopify_api.
if you need these tools, install the shopify_cli gem
```

Therefore, what this change does is simply fully remove the binary which no longer belongs to this gem so it doesn't accidentally overwrite the new CLI binary.

The version was bumped (patch bump since no function actually changed) so future installs don't bring the binary in over the new one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Version 9.1.2
+
+* Removes the `shopify` binary which will be used by the Shopify CLI
+
 ## Version 9.1.1
 
 * Make cursor based pagination return relative uri's when fetching next and previous pages. [#726](https://github.com/Shopify/shopify_api/pull/726)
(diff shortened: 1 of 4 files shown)


JuandGirald:
@paulomarg Hi! I'm using the `shopify_app` gem that use `shopify_api` as dependency and I'm getting the error you mentioned! I forked the shopify_app project and update the gem dependency to 9.2.0 but, still i'm getting 

`
shopify command is no longer bundled with shopify_api.
if you need these tools, install the shopify_cli gem
`

Can you help me with that?

mkevinosullivan:
@JuandGirald If you updated or installed the latest `shopify_api` gem, then you still have the old version in your path.

If you run `gem list shopify_api`, you should get a list of versions of the `shopify_api` gem that you have installed.  You want to keep `9.2.0`.

You can run `gem uninstall shopify_api -v x.y.z` for each of the other versions to remove them (obviously replacing x.y.z with the actual versions numbers!)

PR #751, opened 2020-06-15
Clean up Base user / password on temp sessions

### LINKED ISSUE #744: Basic auth can't be used together with ShopifyAPI::Session.temp
If you've set up basic auth for a private app with `ShopifyAPI::Base.site = ...`, you can't use `ShopifyAPI::Session.temp`, since `user` and `password` aren't cleared and restored on `ShopifyAPI::Base` in `ShopifyAPI::Session.with_session`.

I'm not sure what other implications it would have, but one solution might be to add `user` and `password` to `ShopifyAPI::Session`, and making sure they're cleared and restored accordingly.

Here's a monkey patch for anyone else running into this:

```ruby
module ShopifyAPI
  class Session
    class << self
      def with_session(session, &_block)
        original_session = extract_current_session
        original_user = ShopifyAPI::Base.user
        original_password = ShopifyAPI::Base.password

        begin
          ShopifyAPI::Base.clear_session
          ShopifyAPI::Base.activate_session(session)
          yield
        ensure
          ShopifyAPI::Base.activate_session(original_session)
          ShopifyAPI::Base.user = original_user
          ShopifyAPI::Base.password = original_password
        end
      end
    end
  end
end
```

paulomarg:
Fixes #744.
 
If an app was set up using `ShopifyAPI::Base.user/password`, temp sessions wouldn't work properly because those values would be set for them as well, which might not be desirable.

This change essentially ensures that we clear those values on `ShopifyAPI::Base` before running the block and restore them before exiting. That way, the temp session is indeed fresh.

**Concern**: while starting a fully fresh session makes sense and the suggested fix works, I wonder if it can be considered backward compatible. There could be a scenario where someone relies on the Basic auth settings being used for temp sessions, and those apps would be broken if they bumped their API version.

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -28,12 +28,17 @@ module ShopifyAPI
 
       def with_session(session, &_block)
         original_session = extract_current_session
+        original_user = ShopifyAPI::Base.user
+        original_password = ShopifyAPI::Base.password
 
         begin
+          ShopifyAPI::Base.clear_session
           ShopifyAPI::Base.activate_session(session)
           yield
         ensure
           ShopifyAPI::Base.activate_session(original_session)
+          ShopifyAPI::Base.user = original_user
+          ShopifyAPI::Base.password = original_password
         end
       end
 
--- a/test/session_test.rb
+++ b/test/session_test.rb
@@ -78,13 +78,17 @@ class SessionTest < Test::Unit::TestCase
     assert_equal "My test secret", ShopifyAPI::Session.secret
   end
 
-  test "#temp reset ShopifyAPI::Base.site to original value" do
+  test "#temp reset ShopifyAPI::Base values to original value" do
     session1 = ShopifyAPI::Session.new(domain: 'fakeshop.myshopify.com', token: 'token1', api_version: '2019-01')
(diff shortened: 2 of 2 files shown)


tanema:
Can you give an example of the temporary sessinos that would be broken? Should we provide a migration guide?

paulomarg:
To be honest, I'm having a bit of trouble figuring out an actual use case for this change. The entire scenario is a bit convoluted IMHO because I don't really see why you'd need multiple sessions in a single request - is there a valid reason for accessing multiple shops in one request?

Essentially, in my view what would happen here is that the developer would set up:

```ruby
Base.site = 'https://user:pass@store.shopify.com'

Session.temp(domain: 'another-store.shopify.com') do
  # This session used to use Base.user and Base.password
end
```

Now that I think about it, I'm not sure how much sense the use case from the issue makes, since the above behaviour actually seems right if you think about it.

tanema:
That is an interesting insight. I think it may be cross shop requests. These might be apps that sync data between shops.

paulomarg:
I see the benefit of this change if cross-shop requests are a valid concept, because otherwise we are essentially preventing apps from reaching both shops in this scenario.

PR #781, opened 2020-10-05
Filter unwanted product / variant attributes on creation

### LINKED ISSUE #775: Variant#serializable_hash in 9.2.0
Hi folks,

Outside of a `ShopifyApp` session block, I was trying to call `Variant#as_json`, and kept getting the very unhelpful error:
`TypeError (compared with non class/module)` with no traceback.

I was able to track it down to this recent addition by @marschattha in `lib/shopify_api/resources/variant.rb`:

```ruby
def allow_inventory_params?
  Base.api_version < ApiVersion.find_version('2019-10')
end
```

If you're using `ShopifyApp` and do something like:
```ruby
product = shop.with_shopify_session { ShopifyAPI::Product.find(XXX) } 
```

the `Base.api_version` the api version is only set within the block, and is otherwise `ShopifyAPI::ApiVersion::NullVersion` causing the `TypeError` on comparison.

It feels weird to not be able to serialize an object outside of a session block.  Not sure how/if you'd like to solve this issue, but thought I might share my research here in case others experience the same problem.

paulomarg:
Fixes #775 

## Why am I making these changes?

We had reports that the changes made in #731 were causing the JSON serialization of API objects to break when `Base.api_version` wasn't set, for instance outside the scope of a temp session.

The issue here was that we were filtering the fields out only when serializing the Product / Variant objects, rather than removing them altogether (even though they were no longer allowed in the API after version `2019-10`).

## What is this PR doing?

This PR changes those objects so that we remove the fields when the object is first created, so that we don't need to check later on.

### DIFF
--- a/lib/shopify_api/resources/product.rb
+++ b/lib/shopify_api/resources/product.rb
@@ -5,6 +5,11 @@ module ShopifyAPI
 
     early_july_pagination_release!
 
+    def initialize(*)
+      super
+      self.attributes.except!('total_inventory') unless allow_inventory_params?
+    end
+
     # compute the price range
     def price_range
       prices = variants.collect(&:price).collect(&:to_f)
@@ -19,10 +24,6 @@ module ShopifyAPI
     def total_inventory=(new_value)
       raise_deprecated_inventory_call('total_inventory') unless allow_inventory_params?
       super
-    end
-
-    def serializable_hash(options = {})
-      allow_inventory_params? ? super(options) : super(options).except('total_inventory')
     end
 
     def collections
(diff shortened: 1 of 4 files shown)


lib/shopify_api/resources/product.rb
```diff
@@ -5,6 +5,11 @@ class Product < Base
 
     early_july_pagination_release!
 
+    def initialize(*args, **kwargs)
+      super(*args, **kwargs)
```
andyw8, line 9: This should be fine as just `super`.

andyw8:
Since 2019-10 is no longer supported, would we be able to remove that check?

paulomarg:
> Since 2019-10 is no longer supported, would we be able to remove that check?

Seems we may not be able to fully drop this code yet, as the Admin API still returns these fields and explicitly forbids saving them. So if we remove this code to filter out these fields, we'll end up trying to save them and that may lead to API requests failing.

tanema (approved):

tanema (approved):

PR #783, opened 2020-10-05
Fix rubocop warnings

paulomarg:
This PR fixes (or ignores) rubocop warnings on the codebase, and removes the associated ignores from the TODO file.


lib/shopify_api/resources/line_item.rb
```diff
@@ -2,11 +2,11 @@ module ShopifyAPI
...
     class Property < Base
       def initialize(*args)
-        attributes = args[0] || {}
-        persisted = args[1] || false
+        self.attributes = args[0] || {}
+        self.persisted = args[1] || false
         super
       rescue NameError
-        attributes = attributes.to_hash
+        self.attributes = attributes.to_hash
```
paulomarg, line 9: I wasn't sure whether to do this or ignore this particular issue, but it doesn't seem like these variables were used at all in the previous code.

andyw8: Yeah, it seems flawed.

test/price_rule_test.rb
```diff
@@ -43,7 +43,7 @@ def test_create_price_rule
     )
 
     assert_equal '{"price_rule":{"target_type":"line_item","allocation_method":"across","value_type":"fixed_amount","value":-10.0,"customer_selection":"all","starts_at":"2017-01-19T00:00:00Z"}}', WebMock.last_request.body
-    assert_equal -10, price_rule.value
+    assert_equal(-10, price_rule.value)
```
andyw8, line 46: I don't understand why it's correcting this but not the other `assert_equal` calls?

test/session_test.rb
```diff
@@ -519,7 +519,7 @@ def make_sorted_params(params)
 
   def generate_signature(params)
     params = make_sorted_params(params) if params.is_a?(Hash)
-    OpenSSL::HMAC.hexdigest(OpenSSL::Digest::SHA256.new, ShopifyAPI::Session.secret, params)
+    OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new('SHA256'), ShopifyAPI::Session.secret, params)
```
andyw8, line 522: I would be cautious of how this might impact different Ruby versions.

paulomarg: We run the tests for every ruby version between 2.4 and 2.7, and the CI build worked. Is there anything else in particular we should look for?

test/price_rule_test.rb
paulomarg: This one in particular complained about the `-10` as it's not clear whether it's a parameter or not. We have ~600 infractions on the check for calling methods without parameters, but that's a lower severity warning. We'll get there 😄

test/session_test.rb
andyw8: Ok, then it should be fine. I'm just cautious of OpenSSL related things being fragile, based on past bad experience!

andyw8 (approved):

PR #787, opened 2020-10-14
Fix rubocop conflicts

mkevinosullivan:
### rubocop updates for
- Layout/LineLength
- Style/MethodCallWithArgsParentheses (some updates missed/different to those autocorrected in #791)


andyw8:
@mkevinosullivan Are these all auto-corrections?

mkevinosullivan:
No, went through them manually, running `rake test` after almost every file change!

paulomarg:
Seems like we're still failing for some test files.

mkevinosullivan:
Weird, they all passed on my local machine .. I've just pushed another commit.

paulomarg:
> Weird, they all passed on my local machine .. I've just pushed another commit.

That actually happened to me as well when I turned rubocop on. It seems that some of the rules behave differently for different Ruby versions.

andyw8:
One difficulty with this repo is that we don't commit Gemfile.lock, so there is no guarantee that you'll be running the same version of RuboCop as someone else, or as CI.

andyw8:
> One difficulty with this repo is that we don't commit Gemfile.lock, so there is no guarantee that you'll be running the same version of RuboCop as someone else, or as CI.

^ now fixed in https://github.com/Shopify/shopify_api/pull/788

paulomarg (approved):
Skipped checking the tests, but the other files LGTM!

andyw8 (approved):

PR #790, opened 2020-10-15
Use rubocop-shopify gem

andyw8:

PR #791, opened 2020-10-15
Autofix rubocop conflicts

mkevinosullivan:
### rubocop updates for
- Style/HashSyntax
- Style/MethodCallWithArgsParentheses

(used `rubocop` autocorrect capability)


.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config`
-# on 2020-10-15 13:52:14 UTC using RuboCop version 0.93.1.
```
paulomarg, line 3: This file will conflict nicely with the one from the PR to use the shopify rubocop gem. Otherwise it all looks good

paulomarg (approved):

test/webhook_test.rb
```diff
@@ -2,19 +2,19 @@
 
 class WebhookTest < Test::Unit::TestCase
   test 'create should create a webhook' do
-    fake 'webhooks', :method => :post, :status => 201, :body => load_fixture('webhook')
```
andyw8, line 5: It's strange the the auto-correct added parens in some in other places but not here.

andyw8 (approved):

.rubocop_todo.yml
andyw8: I already merged #790 so it doesn't seem to causing any problem.

PR #794, opened 2020-10-28
Enabling frozen string literal cop

paulomarg:
This PR adds the `# frozen_string_literal: true` comment to essentially all files in the code. The change is essentially the same everywhere.

The only file I'd like properly reviewed would be the changelog - I wonder if these comments may lead to apps using this gem breaking if they try and change the literals.


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Freezes all string literals in the library - possibly breaking apps that edit them
```
andyw8, line 3: Maybe something like:

> Freeze all string literals. This should have no impact unless your application is modifying ('monkeypatching') the internals of the library in a unusual way.

paulomarg: Yeah that's great! I wonder if this warrants a major version bump for the next release.

mkevinosullivan (approved):

CHANGELOG.md
andyw8: We're not changing the public interface, but perhaps out of caution we should have a major bump.

PR #797, opened 2020-11-02
Add API support for scheduled fulfillment orders

linzhao125:
Fixes: https://github.com/Shopify/shopify/issues/255067

Add `open` and `reschedule` APIs support for `scheduled` fulfillment orders.

- [x] POST /admin/api/2021-01/fulfillment_orders/{fulfillment_order_id}/open.json

- [x] POST /admin/api/2021-01/fulfillment_orders/{fulfillment_order_id}/reschedule.json

### DIFF
--- a/lib/shopify_api/resources/fulfillment_order.rb
+++ b/lib/shopify_api/resources/fulfillment_order.rb
@@ -52,6 +52,19 @@ module ShopifyAPI
       keyed_fulfillment_orders
     end
 
+    def open
+      load_attributes_from_response(post(:open, {}, only_id))
+    end
+
+    def reschedule(new_fulfill_at:)
+      body = {
+        fulfillment_order: {
+          new_fulfill_at: new_fulfill_at,
+        },
+      }
+      load_attributes_from_response(post(:reschedule, {}, body.to_json))
+    end
+
     def close(message: nil)
       body = {
         fulfillment_order: {
--- a/test/fulfillment_order_test.rb
+++ b/test/fulfillment_order_test.rb
@@ -227,6 +227,57 @@ class FulFillmentOrderTest < Test::Unit::TestCase
       end
     end
 
+    context "#open" do
+      should "be able to open fulfillment order" do
+        fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021)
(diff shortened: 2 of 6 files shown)


andyw8:
Can you add an update to the `CHANGELOG` please (in the Unreleased section).

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -51,6 +51,19 @@ def cancel
       keyed_fulfillment_orders
     end
 
+    def open
+      load_attributes_from_response(post(:open, {}, body.to_json))
```
c-johns, line 55: There is no `body` here, so I think this needs to be: `load_attributes_from_response(post(:open, {}, only_id))`

test/fulfillment_order_test.rb
```diff
@@ -227,6 +227,57 @@ def setup
       end
     end
 
+    context "#open" do
+      should "be able to open fulfillment order" do
+        fulfillment_order = ShopifyAPI::FulfillmentOrder.find(519788021)
+        fulfillment_order.status = 'scheduled'
+
+        opened = ActiveSupport::JSON.decode(load_fixture('fulfillment_order'))
+        opened['status'] = 'open'
```
andyw8, line 236: Why do we set the status here before making the `POST`?

andyw8: Oh, we're comparing it against the fixture, I see.

linzhao125: This fixture has the status `open` already actually, here I am trying to make sure it is `open`, it should be fine without it. Just clearer this way IMO.

andyw8 (approved):

c-johns (approved):

andyw8:
FYI, the version of shopify_api in core is a little behind.

I tested with the latest release recently and found a few failures, so we'd have to address those before upgrading:

https://github.com/shopify/shopify/commit/84c96b3404feb604403717dfbcf4418cf7aa0adb
https://buildkite.com/shopify/shopify-selective-tests/builds/270945

andyw8:
PR for the above: https://github.com/Shopify/shopify/pull/269163

PR #800, opened 2020-11-09
[Fulfillment Hold] Adding support for Fulfillment hold release on the  fulfillment order resource

glen-j-nkali:
Fixes https://github.com/Shopify/store/issues/19036

Add a new endpoint on the FulfillmentOrder resource to release the fulfillment hold on an order (all associated fulfillment orders)

- [ ] POST admin/api/2021-04/fulfillment_orders/release_hold.json
  `{ order_id: order_id }`


lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -22,6 +22,11 @@ def self.find(scope, *args)
       end
     end
 
+    def self.release_hold(order_id:)
```
glen-j-nkali, line 25: We've extensively discussed why this action has to live on the `fulfillment order` resource and why we accept an order id here merely as a convenience method. If you need more details please let me know and I'll share the docs with you.

MalazAlamir (approved):

lib/shopify_api/resources/fulfillment_order.rb
andyw8, line 25: I wonder this should raise an exception if called for an unsupported version of the API?

glen-j-nkali: good point, I am going to add a minimum supported version 👍

lib/shopify_api/resources/fulfillment_order.rb
```diff
@@ -22,6 +22,13 @@ def self.find(scope, *args)
...
+      ShopifyAPI::Base.version_validation!('unstable')
```
glen-j-nkali, line 26: you have to be on unstable for now but before merging this (have to wait for #796 first) I'll update to point to the exact version that will include the change.

The plan is to cut this into RC early next year so it will be in the `2021-04` version of the API.

andyw8: I think you've linked to the wrong PR above.

glen-j-nkali: true, I've updated 👍

andyw8:
@glen-j-nkali will you also be updating the python library?

glen-j-nkali:
> @glen-j-nkali will you also be updating the python library?

wasn't planning on doing so because I think @c-johns was doing the work to port all of the changes related to the Fulfillment platform that we have on this version to the python one first.
He's currently on vacation so I'll check with him when he's back to make sure we enforce parity between both gems.

Issue #801, opened 2020-11-12
Re-enable variant#inventory_quantity for serialization


axlekb:
First time venturing into inventory... based on the [docs](https://shopify.dev/docs/admin-api/rest/reference/products/product-variant?api%5Bversion%5D=2020-07)  it appears `variant.inventory_quantity` has re-emerged with a different meaning/use after being deprecated and renamed to `old_inventory_quantity`.  Therefore, I believe it should again be "serialized" with current API versions?

![image](https://user-images.githubusercontent.com/2972314/98892174-c6c88f80-246d-11eb-83c1-030c6c464cb2.png)
https://github.com/Shopify/shopify_api/blob/v9.2.0/test/variant_test.rb

```ruby
 def test_deprecated_inventory_fields_are_included_in_2019_07
    ShopifyAPI::Base.api_version = '2019-07'
    assert @variant.as_json.include?('inventory_quantity')
  end

  def test_deprecated_inventory_fields_are_removed_in_2020_01
    ShopifyAPI::Base.api_version = '2020-01'
    refute @variant.as_json.include?('inventory_quantity')
  end
```

https://github.com/Shopify/shopify_api/blob/v9.2.0/lib/shopify_api/resources/variant.rb
```ruby
def serializable_hash(options = {})
      if allow_inventory_params?
        super(options)
      else
        super(options).tap do |resource|
          (resource['variant'] || resource).except!('inventory_quantity', 'old_inventory_quantity')
        end
      end
    end
```

```ruby
  def allow_inventory_params?
      Base.api_version < ApiVersion.find_version('2019-10')
    end
```

paulomarg:
Hey @axlekb, thank you for reporting this, we'll look into it!

PR #802, opened 2020-11-13
Allow variant inventory quantity

mllemango:
As suggested in #801, Variant's`inventory_quantity` is now a read-only field, and should be readable from the object. 
The PR allows `inventory_quantity` to be filtered in as part of Variant's attributes. The field is only filtered out once variant is saved. User's will still be able to access the `inventory_quantity` field as an attribute of Variant and from the json object both before and after a save.

Also removed tests that are focused on deprecated versions, as well as the `allow_inventory_params?` method since that was also based on deprecated versions of the API.


lib/shopify_api/resources/variant.rb
```diff
@@ -26,10 +26,9 @@ def old_inventory_quantity=(new_value)
       super
     end
 
-    def serializable_hash(options = {})
-      super(options).tap do |resource|
-        (resource['variant'] || resource).except!('inventory_quantity', 'old_inventory_quantity')
-      end
+    def save
+      attributes.except!('inventory_quantity', 'old_inventory_quantity')
+      super
```
paulomarg: This is great, though we are still dropping the fields. Would it make sense to restore the original attributes in an `ensure` block?

test/variant_test.rb
```diff
@@ -35,30 +35,10 @@ def test_delete_variant
...
-    @variant.inventory_quantity_adjustment = 8
-  end
-
-  def test_setting_variant_inventory_quantity_adjustment_fails_in_2019_10_api
-    ShopifyAPI::Base.api_version = '2019-10'
-    refresh_variant(api_version: ShopifyAPI::Base.api_version)
-    assert_raises(ShopifyAPI::ValidationException) do
-      @variant.inventory_quantity_adjustment = 8
-    end
+    assert(10, @variant.inventory_quantity)
```
andyw8, line 41: Should be `assert_equal`? (below too)

mllemango: oops, yes it should be

lib/shopify_api/resources/variant.rb
mllemango: Resolved offline, original attributes are still present after the save!

paulomarg (approved):

PR #808, opened 2020-11-30
Move CI to Github Actions

mllemango:
We want to move away from running the CI on travis.org due the platform going into [read-only mode at the end of the year](https://docs.travis-ci.com/user/migrate/open-source-repository-migration/#frequently-asked-questions). 

Note: 
1. I couldn't emulate `allow_failures` that the Travis currently uses, due to this weird behaviour in Github actions where an extra failing test pops up (see [here](https://github.com/Shopify/shopify_api/runs/1468368626?check_suite_focus=true))

1. Not sure how good it is to remove `Gemfile.lock` in the CI steps but I couldn't get the pipeline to read in the test gemfiles otherwise

1. The additional `gem installs` were necessary for the pipeline to run, not sure why.


andyw8:
> I couldn't emulate allow_failures that the Travis currently uses, due to this weird behaviour in Github actions where an extra failing test pops up (see here)

Odd, I also notice it's not running `Ruby 2.7, Gemfile_ar41`.

andyw8:
But it _does_ appear in the list of Checks on GitHub...

mllemango:
> > I couldn't emulate allow_failures that the Travis currently uses, due to this weird behaviour in Github actions where an extra failing test pops up (see here)
> 
> Odd, I also notice it's not running `Ruby 2.7, Gemfile_ar41`.

Oops, the most recent run is actually incorrect. We should be excluding `Ruby 2.7, Gemfile_ar41` like travis

andyw8:
I'm not sure if this will solve any of the encountered issues, but on https://github.com/actions/setup-ruby/issues/80 is seems like there is to deprecate `actions/setup-ruby` in favour of `https://github.com/ruby/setup-ruby` – maybe worth trying?

mllemango:
Good catch! I keep switching back to actions/setup-ruby because its testable locally 😅

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Ruby ${{ matrix.version }}
+        uses: ruby/setup-ruby@v1
+        with: 
+          ruby-version: ${{ matrix.version }}
+      - name: Install dependencies
+        run: |
+          export BUNDLE_GEMFILE=${GITHUB_WORKSPACE}/${{ matrix.gemfile }}
+          rm -f ${GITHUB_WORKSPACE}/Gemfile.lock
```
andyw8, line 27: What error are you seeing when this isn't present?

I think it could be confusing if there are dependency-related failures and it's not clear the committed `Gemfile.lock` isn't being used.

mllemango: The workflow won't use the files: Gemfile_ar41, Gemfile_ar50, Gemfile_ar51, Gemfile_ar_master when this line isn't present :/ I'm not sure how else to go about it

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+          gem update --system
+          gem install bundler
+          gem install jekyll
```
andyw8, line 30: The reason this is needed is because `jekyll` is in a `optional` group in the `Gemfile`, so needs to be installed with `bundle --with docs`.

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Set up Ruby ${{ matrix.version }}
+        uses: ruby/setup-ruby@v1
+        with: 
+          ruby-version: ${{ matrix.version }}
+      - name: Install dependencies
+        run: |
+          export BUNDLE_GEMFILE=${GITHUB_WORKSPACE}/${{ matrix.gemfile }}
+          rm -f ${GITHUB_WORKSPACE}/Gemfile.lock
+          gem update --system
+          gem install bundler
```
andyw8, line 29: Does this not get handles automatically? https://github.com/ruby/setup-ruby#bundler

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up Ruby ${{ matrix.version }}
+        uses: ruby/setup-ruby@v1
+        with: 
+          ruby-version: ${{ matrix.version }}
+      - name: Install dependencies
+        run: |
+          export BUNDLE_GEMFILE=${GITHUB_WORKSPACE}/${{ matrix.gemfile }}
+          rm -f ${GITHUB_WORKSPACE}/Gemfile.lock
+          gem update --system
```
andyw8, line 28: Could we assume that the `ruby/setup-ruby` action will be kept up to date, so we wouldn't have to update rubygems ourselves?

.github/workflows/build.yml
mllemango: See https://github.com/Shopify/shopify_api/runs/1476242755#step:5:7 for the failure msg when `Gemfile.lock` isnt deleted. It for some reason defaults to activesupport-6.0.3.1?

andyw8:
@mllemango I'm seeing this deprecation message, should be an easy fix:

`[DEPRECATED] The `--with` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set with 'docs'`, and stop using this flag`

andyw8 (approved):

PR #812, opened 2020-12-03
Handle 422-500x errors when the response body is an array

vbyno:
For some reasons, the error responce could be an array. In that case
MessageEnricher raises TypeError:
  no implicit conversion of String into Integer
This commit fixes it


lib/shopify_api/message_enricher.rb
```diff
@@ -14,7 +14,7 @@ def message
             elsif parsed_body['errors']
               Array(parsed_body['errors']).join('; ')
             end
-          rescue JSON::ParserError
+          rescue JSON::ParserError, TypeError
```
paulomarg, line 17: I think catching `TypeError` is a bit dangerous because we may end up masking other errors. When the errors are an array, can we just loop over them and join them as we do with the elements of `errors`?

paulomarg (commented):
Thank you for contributing! This is quite worthwhile, though we probably want to catch as many of the errors as we can to make the messages as informative as we can.

PR #816, opened 2020-12-08
Add note about keeping Shopify/shopify up to date

andyw8:


mllemango (approved):
Is `shopify/shopify` the only place with this dependency?

andyw8:
That's the only one I know of but we can add more if discovered.

andyw8:
Added shopify_app

PR #818, opened 2020-12-11
Avoid depending on ActiveSupport

andyw8:
In #817 we ran into an issue relating to ActiveSupport not being loaded when updating to Rails 6.1.0. We only use it for some trivial time-related helpers – I think we can drop it completely.

### DIFF
--- a/test/session_test.rb
+++ b/test/session_test.rb
@@ -3,6 +3,8 @@ require 'test_helper'
 require 'timecop'
 
 class SessionTest < Test::Unit::TestCase
+  SECONDS_IN_A_DAY = 24 * 60 * 60
+
   def setup
     super
     ShopifyAPI::Session.secret = 'secret'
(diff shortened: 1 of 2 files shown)


test/session_test.rb
```diff
@@ -373,7 +375,7 @@ def setup
   end
 
   test "raise error if timestamp is too old" do
-    params = { code: "any-code", timestamp: Time.now - 2.days }
+    params = { code: "any-code", timestamp: Time.now - 2 * SECONDS_IN_A_DAY }
```
paulomarg, line 378: Extremely nitpicky, but would it make sense to use `Session.SECONDS_IN_A_DAY` here?

paulomarg (approved):
LGTM, I'm happy for my nitpick to be ignored.

test/session_test.rb
andyw8: There's a couple of reasons why I usually avoid referencing a implementation's constants in the tests:

- If the constant is wrong in the implementation, and the test references that, then you lose out on the 'double-entry bookkeeping' safeguard of writing a test.
- If I choose to rename the constant, or move it somewhere else, the tests shouldn't break since the implementation hasn't changed.

PR #821, opened 2020-12-15
Add logging based on env var

mllemango:
Logging is enabled through `export SHOPIFY_LOG_PATH={path}`

Is this the right way to go about logging? Current solution is to have users set an env variable SHOPIFY_LOG_PATH.
Should I also include some method inside the API to enable logging such as `ShopifyAPI::Logging=true`?

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -25,7 +25,7 @@ Tip: include an error message (in a `<details></details>` tag) if your issue is 
 1.
 
 ## Logs
-Enable logs with `export SHOPIFY_LOG=1` and paste them here
+Enable logs with `export SHOPIFY_LOG=1` and paste them here from "shopify_api.log"
 
 
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,36 @@
+# Issue summary
+
+Write a short description of the issue here ↓
+
+
+
+## Expected behavior
+
+What do you think should happen?
+
+
+
+## Actual behavior
+
+What actually happens?
+
+Tip: include an error message (in a `<details></details>` tag) if your issue is related to an error
+
+
+
+## Steps to reproduce the problem
+
+1.
+1.
+1.
+
+## Logs
+Enable logs with `export SHOPIFY_LOG=1` and paste them here
+
+
+
+## Specifications
+
+- `shopify_python_api` version:
+- `pyactiveresource` version:
+- Shopify API version used (e.g. `'2020-07'`):
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -25,12 +25,12 @@ Tip: include an error message (in a `<details></details>` tag) if your issue is 
 1.
 
 ## Logs
-Enable logs with `export SHOPIFY_LOG=1` and paste them here from "shopify_api.log"
+If applicable, enable the logs as described in the README, and paste the relevant portion here.
 
 
 
 ## Specifications
 
-- `shopify_python_api` version:
-- `pyactiveresource` version:
-- Shopify API version used (e.g. `'2020-07'`):
+- `shopify_api` version:
+- `activeresource` version:
+- Shopify API version used (example: `2020-07`):
--- a/README.md
+++ b/README.md
@@ -634,6 +634,8 @@ Enable ActiveResource's logger with
 
 `export SHOPIFY_LOG=1`
 
+This will create "shopify_api.log" in the current project directory.
+
 # Additional Resources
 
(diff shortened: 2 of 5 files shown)


andyw8:
Let's document this in the README.

lib/shopify_api.rb
```diff
@@ -32,3 +32,8 @@ module ShopifyAPI
 else
   require 'active_resource/connection_ext'
 end
+
+if ENV["SHOPIFY_LOG"] == '1'
+  ActiveResource::Base.logger = Logger.new("shopify_api.log")
+  ActiveResource::DetailedLogSubscriber
```
andyw8, line 38: Hmm, does referencing this without calling a method still have an affect?

mllemango: I'm able to get this working locally. I assumed that the entire file was executed

mkevinosullivan (approved):

lib/shopify_api.rb
andyw8: Hmm, this may be a side-effect of the autoloading. Let me check it locally...

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+## Steps to reproduce the problem
+
+1.
+1.
+1.
+
+## Logs
+Enable logs with `export SHOPIFY_LOG=1` and paste them here from "shopify_api.log"
```
andyw8, line 28:
```suggestion
Enable logs with `export SHOPIFY_LOG=1` and paste them here from "shopify_api.log"
```
If applicable, enable the logs as described in the README, and paste the relevant portion here.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+
+## Specifications
+
+- `shopify_python_api` version:
```
andyw8, line 34: Python? 😁

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+## Logs
+If applicable, enable the logs as described in the README, and paste the relevant portion here.
+
+
+
+## Specifications
+
+- `shopify_api` version:
+- `activeresource` version:
+- Shopify API version used (e.g. `'2020-07'`):
```
gfscott, line 36:
```suggestion
- Shopify API version used (example: `2020-07`):
```

gfscott (commented):
Shopify style is to just use `example` instead of latin abbreviations like `e.g.` or `i.e.`!

lib/shopify_api.rb
andyw8: Ok, so `DetailedLogSubscriber` currently does this:

https://github.com/Shopify/shopify_api/blob/master/lib/active_resource/detailed_log_subscriber.rb#L53

I think we should move this line out of `DetailedLogSubscriber` and into our logging setup code.

Otherwise it's not intuitive that loading the class will enable logging as this does.

cc @Shopify/platform-dev-tools-education for a second opinion, as the code has been this way for a [very long time](https://github.com/Shopify/shopify_api/commit/5b865286e83074e3020d0b3b36590b158368cf6f).

lib/shopify_api.rb
```diff
@@ -32,3 +32,8 @@ module ShopifyAPI
 else
   require 'active_resource/connection_ext'
 end
+
+if ENV["SHOPIFY_LOG"] == '1'
+  ActiveResource::Base.logger = Logger.new("shopify_api.log")
```
tanema, line 37: Should the env var be a path instead to make this easier to use in production? `Logger.new(ENV['SHOPIFY_LOG'])`

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
+# Issue summary
```
tanema, line 1: why is there an issue template in this PR?

lib/shopify_api.rb
mllemango: We could do that!

.github/ISSUE_TEMPLATE.md
mllemango: The issue template is added to remind users to paste in logs if applicable

README.md
```diff
@@ -632,9 +632,9 @@ bundle exec rake docker
 
 Enable ActiveResource's logger with 
 
-`export SHOPIFY_LOG=1`
+`export SHOPIFY_LOG_PATH={your_log_path}`
 
-This will create "shopify_api.log" in the current project directory.
+This will create file `your_log_path` in the current project directory.
```
andyw8, line 637:
```suggestion
This will log to a file at the given path, relative to the current project directory.
```

andyw8 (approved):

PR #829, opened 2021-01-13
Provide Scopes value object to encapsulate scope operations

rezaansyed:
This is a follow up on our discussion: https://github.com/Shopify/omniauth-shopify-oauth2/pull/96#issuecomment-714478697

## What does this PR solve?
There are common operations that are used for scopes on OmniAuth and apps consuming the gem. Such operations include serializing, deserializing and normalizing scopes. This PR introduces the `Scopes` value object that will be used to perform such operations. This helps ensure that apps do not potentially write duplicate code for scopes and delegating this responsibility to the Shopify OmniAuth gem.

### Simplifying valid scope checks
**Before**
```ruby
def valid_scope?(token)
  return false unless token && params[:scope] && token['scope']
  expected_scope = normalized_scopes(params[:scope]).sort
  (expected_scope == token['scope'].split(SCOPE_DELIMITER).sort)
end

def normalized_scopes(scopes)
  scope_list = scopes.to_s.split(SCOPE_DELIMITER).map(&:strip).reject(&:empty?).uniq
  ignore_scopes = scope_list.map { |scope| scope =~ /\A(unauthenticated_)?write_(.*)\z/ && "#{$1}read_#{$2}" }.compact
  scope_list - ignore_scopes
end
```

**After**
```ruby
def valid_scope?(token)
  return false unless token && params[:scope] && token['scope']
  expected_api_access = ShopifyAPI::ApiAccess.new(config[:scope])
  actual_api_access = ShopifyAPI::ApiAccess.new(token['scope'])

  actual_api_access == expected_api_access
end
```

### Delegating check for whether scopes are mismatched or subset of another
```ruby
def scopes_mismatch?(existing_scopes, requesting_scopes)
  existing_access = ShopifyAPI::ApiAccess.new(existing_scopes)
  requested_access = ShopifyAPI::ApiAccess.new(requesting_scopes)
  
  existing_access == requested_access
end

def scopes_subset?(existing_scopes, requesting_scopes)
  existing_access = ShopifyAPI::ApiAccess.new(existing_scopes)
  requested_access = ShopifyAPI::ApiAccess.new(requesting_scopes)

  requested_access.covers?(existing_access)
end
```


lib/shopify_api/api_access.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    alias :eql? :==
+
+    def hash
+      compressed_scopes.hash
+    end
+
+    protected
+
+    attr_reader :compressed_scopes, :expanded_scopes
```
NabeelAhsen, line 40: Do we need any tests that ensure `store_scopes` creates these two attributes as expected?

NabeelAhsen (approved):
Looks solid!

lib/shopify_api/api_access.rb
rezaansyed: We don't  because the behaviour `store_scopes` is a private method, while the `compressed_scopes` and `expanded_scopes` are protected. These will only be used within the context of the object itself to allow it to perform common operations that are public.

NabeelAhsen: Ah, I missed the `protected` line above!

test/api_access_test.rb
```diff
@@ -0,0 +1,132 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class ScopesTest < Minitest::Test
```
paulomarg, line 4: Nit: should this class be named `ApiAccessTest`?

test/api_access_test.rb
```diff
@@ -0,0 +1,132 @@
...
+  end
+
+  def test_to_s_outputs_scopes_as_a_comma_separated_list_without_implied_read_scopes
+    serialized_scopes = "read_products,write_orders"
+    expected_scopes = ShopifyAPI::ApiAccess.new(%w(read_products read_orders write_orders))
+    assert_equal expected_scopes.to_s, serialized_scopes
+  end
+
+  def test_to_a_outputs_scopes_as_an_array_of_strings_without_implied_read_scopes
+    serialized_scopes = %w(read_products write_orders)
```
paulomarg, line 104: Should this be a string?

paulomarg (approved):
LGTM! Couple of nitpicks only.

test/api_access_test.rb
rezaansyed: It can either be a string or an array of scopes so that we can support two types of inputs when instantiating ApiAccess.

paulomarg: I was just wondering if it made sense for `serialized_scopes` to be an array - I thought the tests was slightly off in that but it seems like it's just the variable name that is ''''wrong''''. No reason to change it 🙂

lib/shopify_api/api_access.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    private
+
+    def store_scopes(scope_names)
+      scopes = scope_names.map(&:strip).reject(&:empty?).to_set
+      implied_scopes = scopes.map { |scope| scope =~ /\A(unauthenticated_)?write_(.*)\z/ && "#{$1}read_#{$2}" }.compact
```
greatestape, line 46: I think it'd be nice to unpack this line a bit so readers understand what's happening here.

test/api_access_test.rb
```diff
@@ -0,0 +1,132 @@
...
+
+  def test_unauthenticated_is_not_implied_by_authenticated_access
+    unauthenticated_scopes = ShopifyAPI::ApiAccess.new(%w(unauthenticated_read_orders))
+    authenticated_read_scopes = ShopifyAPI::ApiAccess.new(%w(read_orders))
+    authenticated_write_scopes = ShopifyAPI::ApiAccess.new(%w(write_orders))
+    refute_equal unauthenticated_scopes, authenticated_read_scopes
+    refute_equal unauthenticated_scopes, authenticated_write_scopes
+  end
+
+  def test_scopes_includes_is_truthy_for_same_scopes
```
greatestape, line 43: Should this say "covers" instead of "includes"?

test/api_access_test.rb
```diff
@@ -0,0 +1,132 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class ScopesTest < Minitest::Test
+  def test_write_is_the_same_access_as_read_write_on_the_same_resource
+    full_scopes = ShopifyAPI::ApiAccess.new(%w(read_orders write_orders))
+    expected_scopes = ShopifyAPI::ApiAccess.new(%w(write_orders))
+    assert_equal expected_scopes, full_scopes
```
greatestape, line 8: How would you feel about having more descriptive variable names in these tests?

Like:

```suggestion
    read_write_orders = ShopifyAPI::ApiAccess.new(%w(read_orders write_orders))
    write_orders = ShopifyAPI::ApiAccess.new(%w(write_orders))
    assert_equal write_orders, read_write_orders
```

I know it's explained in the test name, but it feels like it'd be easier to read what the test is actually asserting this way.

greatestape:
It's sort of hard to picture what the big picture is with this change.

Would it make sense to update the description to show a preview of how we could use this new feature in `shopify_app` and/or `omniauth-shopify-oauth2`?

greatestape (approved):
LGTM

alexaitken (commented):
Small change to the test relying on the order of the elements from `to_a`.

PR #835, opened 2021-01-30
Feature: `respond_to?` to test attributes existance

hrdwdmrbl:
Use case: For example, sometimes LineItems have an origin_location property, while other times they don't.
In my code right now, I test for the existence of the origin_location property using the code `line_item.attributes[:origin_location]`.
If my code does not check for the existence before calling `line_item.origin_location`, then an exception will be raised.
The idiomatic way of checking for the existence of a method is to use `respond_to?(method_name)`.
IMHO LineItem should return nil when the origin_location property isn't defined, but this is a Pareto improvement.

If you would like a test to cover this new feature, please tell me where and how the test should be implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,6 @@ To follow these usage guides, you will need to:
 
 - have a working knowledge of ruby and a web framework such as Rails or Sinatra
 - have a Shopify Partner account and development store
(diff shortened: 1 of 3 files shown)


hrdwdmrbl:
Re-Check please!

PR #842, opened 2021-02-22
Remove access tokens in log headers

mllemango:
Removing sensitive information from the log
`INFO -- : Headers: {"Accept"=>"application/json", "User-Agent"=>"ShopifyAPI/9.3.0 ActiveResource/5.1.1 Ruby/2.7.1", "X-Shopify-Access-Token"=>"..."}`

becomes

`INFO -- : Headers: {"Accept"=>"application/json", "User-Agent"=>"ShopifyAPI/9.3.0 ActiveResource/5.1.1 Ruby/2.7.1"}`

### DIFF
--- a/lib/active_resource/detailed_log_subscriber.rb
+++ b/lib/active_resource/detailed_log_subscriber.rb
@@ -3,6 +3,9 @@ module ActiveResource
   class DetailedLogSubscriber < ActiveSupport::LogSubscriber
     VERSION_EOL_WARNING_HEADER = 'x-shopify-api-version-warning'
     VERSION_DEPRECATION_HEADER = 'x-shopify-api-deprecated-reason'
+    SHOPIFY_ACCESS_TOKEN = 'X-Shopify-Access-Token'
+    FILTERED = '[FILTERED]'
+
     def request(event)
       log_request_response_details(event)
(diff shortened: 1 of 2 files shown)


lib/active_resource/detailed_log_subscriber.rb
```diff
@@ -20,7 +20,7 @@ def log_request_response_details(event)
       request_body = data.first
 
       info("Request:\n#{request_body}") if request_body
-      info("Headers: #{headers.inspect}")
+      info("Headers: #{headers.except("X-Shopify-Access-Token").inspect}")
       info("Response:\n#{event.payload[:response].body}")
```
andyw8, line 24: I think it might be better to 'blank' the token rather than remove, otherwise it may cause confusion when debugging if it appears the header is missing.

mllemango: That makes sense, instead of a blank would it be clearer to put "..." or something like that?

andyw8: Yeah, or maybe `[FILTERED]` to match what Rails does with `filter_parameters`:

https://blog.appsignal.com/2013/04/18/filter-sensitive-data-from-params.html

andyw8 (approved):

PR #843, opened 2021-02-23
Add access_scopes attribute to session object

NabeelAhsen:
> Related: https://github.com/Shopify/shopify_app/pull/1192

This PR introduces an `access_scopes` attribute to the Session object. 

## Things to focus on

1. Do we determine the validity of a session through this attribute?
2. Is there immediate benefit in returning a `ShopifyAPI::ApiAccess` object? Will a string suffice?
3. ~~Is this level of error handling needed?~~

### DIFF
--- a/lib/shopify_api/session.rb
+++ b/lib/shopify_api/session.rb
@@ -13,7 +13,7 @@ module ShopifyAPI
     self.myshopify_domain = 'myshopify.com'
 
     attr_accessor :domain, :token, :name, :extra
-    attr_reader :api_version
+    attr_reader :api_version, :access_scopes
     alias_method :url, :domain
 
     class << self
@@ -92,10 +92,11 @@ module ShopifyAPI
       end
     end
 
-    def initialize(domain:, token:, api_version: ShopifyAPI::Base.api_version, extra: {})
+    def initialize(domain:, token:, access_scopes: nil, api_version: ShopifyAPI::Base.api_version, extra: {})
       self.domain = self.class.prepare_domain(domain)
       self.api_version = api_version
       self.token = token
+      self.access_scopes = access_scopes
       self.extra = extra
     end
 
@@ -179,6 +180,11 @@ module ShopifyAPI
     end
 
     private
+
+    def access_scopes=(access_scopes)
+      return unless access_scopes
+      @access_scopes = ShopifyAPI::ApiAccess.new(access_scopes)
+    end
 
     def parameterize(params)
       URI.escape(params.collect { |k, v| "#{k}=#{v}" }.join('&'))
--- a/test/session_test.rb
+++ b/test/session_test.rb
@@ -54,6 +54,39 @@ class SessionTest < Test::Unit::TestCase
     assert(session.valid?)
   end
 
+  test "be valid with nil access_scopes" do
+    session = ShopifyAPI::Session.new(
+      domain: "testshop.myshopify.com",
+      token: "any-token",
+      api_version: any_api_version,
+      access_scopes: nil
+    )
+
+    assert(session.valid?)
+  end
+
+  test "be valid with string of access_scopes" do
+    session = ShopifyAPI::Session.new(
+      domain: "testshop.myshopify.com",
+      token: "any-token",
+      api_version: any_api_version,
+      access_scopes: "read_products, write_orders"
+    )
+
+    assert(session.valid?)
+  end
+
+  test "be valid with a collection of access_scopes" do
+    session = ShopifyAPI::Session.new(
+      domain: "testshop.myshopify.com",
+      token: "any-token",
+      api_version: any_api_version,
+      access_scopes: %w(read_products write_orders)
+    )
+
+    assert(session.valid?)
+  end
+
   test "not raise error without params" do
     assert_nothing_raised do
       ShopifyAPI::Session.new(domain: "testshop.myshopify.com", token: "any-token", api_version: any_api_version)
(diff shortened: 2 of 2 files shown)


rezaansyed:
> 1. Do we determine the validity of a session through this attribute?
> 2. Is there immediate benefit in returning a `ShopifyAPI::ApiAccess` object? Will a string suffice?
> 3. Is this level of error handling needed?

1. Not sure what you mean here with this. But this is only a value object used to encompass the access token (and now access scopes)
2. `ShopifyAPI::ApiAccess` should be the preferred approach here. Anyone using this object has the ability to compare two objects for equality and subsets.
3. No. `access_scopes` should be `nil` if not specified.

rezaansyed (approved):
🎩 LGTM 👍🏼

lib/shopify_api/session.rb
```diff
@@ -143,6 +144,16 @@ def api_version=(version)
       end
     end
 
+    def access_scopes=(access_scopes)
+      return unless access_scopes
+
+      @access_scopes = begin
+        ShopifyAPI::ApiAccess.new(access_scopes)
+      rescue
```
ragalie, line 152: I wouldn't do this. Some other (better) options imo, not mutually exclusive:
1. Raise the error when the caller tries to access `access_scopes`, not when we set it. No use raising if it's never even accessed.
2. Raise the error when we set for the caller to handle.
3. Make `valid?` return false when the scopes are bad.

lib/shopify_api/session.rb
```diff
@@ -143,6 +144,16 @@ def api_version=(version)
       end
     end
 
+    def access_scopes=(access_scopes)
```
ragalie, line 147: Does this need to be public?

rezaansyed: yeah this shouldn't be public

lib/shopify_api/session.rb
rezaansyed: 1. This is so that we intentionally fail if anyone tries to get `access_scopes`?
2. If `access_scopes` are not passed in, then this should be nil. 
3. I don't think it should validate on `scopes` for this version because this is going to break existing apps.

lib/shopify_api/session.rb
ragalie: Yeah I guess I meant that it's public right now and we could probably make it private/protected.

lib/shopify_api/session.rb
ragalie: Sorry I think I'm misunderstanding something. The tests imply that this will fail if the scopes are "bad": https://github.com/Shopify/shopify_api/pull/843/files#diff-5e62e8ce4dc1b35b81a888833c02f9a424e3b8eb753fb2301bd8dc736a2eedfcR106

Wouldn't the caller want to know the difference between "scopes are bad" and "no scopes"?

rezaansyed: ohh yes you're right!

rezaansyed: I addressed the bad input cases and making sure that `valid?` handles it for access scopes as well.

lib/shopify_api/session.rb
```diff
@@ -180,6 +181,15 @@ def state
 
     private
 
+    def valid_access_scopes?
```
ragalie, line 184: Will this ever be false?

test/session_test.rb
```diff
@@ -84,6 +117,36 @@ def setup
...
+      domain: "testshop.myshopify.com",
+      token: "any-token",
+      access_scopes: "read_products",
+      api_version: any_api_version
+    )
+    assert_equal "read_products", session.access_scopes.to_s
+  end
+
+  test "session instantiation raises error if bad access scopes are provided" do
+    assert_raises do
```
ragalie, line 140: What does it raise?

lib/shopify_api/session.rb
rezaansyed: No because a bad input will cause `ShopifyAPI::ApiAccess` to raise an error

test/session_test.rb
rezaansyed: Currently it's `NoMethodError` because the `strip` method doesn't exist on the object passed in object passed in. I can specify `NoMethodError`

lib/shopify_api/session.rb
rezaansyed: Do you think we should even have the validity check for access scopes then?

test/session_test.rb
rezaansyed: Should we have a custom error for `ApiAccess` to throw when an invalid input is passed in?

lib/shopify_api/session.rb
NabeelAhsen: I think that's one thing I was uncertain about as well. Either `access_scopes=` will throw an error, or it will correctly be `ShopifyAPI::ApiAccess`. Should this determine whether the entire session is valid?

NabeelAhsen (commented):
Thanks for addressing comments @rezaansyed!

lib/shopify_api/session.rb
rezaansyed: I think given that `ShopifyAPI::ApiAccess` rejects it if you try to pass in a bad input, we won't need a `valid?` check. `valid_access_scopes?` currently will always resolve to true.

test/session_test.rb
NabeelAhsen: I think that because specifying `api_access` is opt-in, it's up to the user to specify good input. In this case, building a new session will either:

* parse access_scopes correctly or
* throw some error (automatically an invalid session)

For this test, we know that it's `NoMethodError` because `strip` doesn't exist on the hash, but really the bad input could throw anything.

mkevinosullivan (approved):
👍🏻

PR #847, opened 2021-03-16
Update method to use grant_options

mllemango:
`create_permission_url` should be able to take in a `grant_options` parameter.


paulomarg (approved):
Makes sense to me!

PR #859, opened 2021-04-23
Make ApiVersion::NullVersion Comparable

nicklozon:
# Problem Statement
`ApiVersion::NullVersion` can be compared with an instance of `ApiVersion` which can cause a `TypeError: compared with non class/module` error. For example: https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/resources/product.rb#L55

The purpose of this PR is to provide a more meaningful exception when `ApiVersion::NullVersion` is used as a comparable.

# Solution
Include `Comparable` statically and alias the spaceship operator to `raise_not_set_error`


nicklozon:
CLA Signed

nicklozon:
Closing in favour of v10

PR #878, opened 2021-09-25
replace deprecated api-version in graphql docs

uurcank:
I just ran into this issue, the sample rake command no longer works. new API version required

```
rake aborted!
ShopifyAPI::ApiVersion::UnknownVersion: ApiVersion.version_lookup_mode is set to `:raise_on_unknown`.
```


paulomarg (approved):
Thank for this!

andyw8:
Perhaps we can write this as `YYYY-MM` and indicate that the user should replace that with the desired version, otherwise the docs will need continually updated.

PR #887, opened 2021-11-10
Drop the dependency on webrick

casperisfine:
It made sense to use these utils when they were part of the stdlib but now it's a bit silly to pull a whole webserver for 5 lines of code.


flavorjones (approved):
LGTM

PR #891, opened 2022-01-07
Remove Active Resource upper bound version

rafaelfranca:
Upper bound versions should be avoided because they only make harder for applications to upgrade
the libraries.


JaKXz (approved):
Makes sense!

arkirchner:
@JaKXz The CI is failing because the main version of `activeresource` does not support Ruby 2.4 and 2.5 anymore.

I updated the CI to test the new version 6 and main with supported Ruby versions.
https://github.com/Shopify/shopify_api/pull/893

paulomarg:
> @JaKXz The CI is failing because the main version of `activeresource` does not support Ruby 2.4 and 2.5 anymore.
> 
> I updated the CI to test the new version 6 and main with supported Ruby versions. #893

Hey @arkirchner, thanks for that. Since those versions have reached EOL, we can just drop them from CI. I've updated the lock file to propagate the change, and updated the CI to exclude them.

teddyhwang (approved):

arkirchner:
Fixed by #891.

PR #919, opened 2022-04-06
Add save deny list to base REST resource

paulomarg:
### WHY are these changes introduced?

Some REST resources (currently only the Variant one) may throw errors if certain fields are included in PUT / POST requests.

### WHAT is this pull request doing?

This PR adds support for the resources to exclude arbitrary fields when saving data.

## Type of change

- [X] Patch: Bug (non-breaking change which fixes an issue)

## Checklist

- [X] I have added a changelog entry, prefixed by the type of change noted above
- [X] I have added/updated tests for this change


mllemango (approved):
NIT: is `save_deny_list` more understandable or is `read_only_list` more understandable? :D

paulomarg:
I was on the fence about that as well. I'll make it `read_only_attributes` then!

PR #920, opened 2022-04-06
Load unknown REST resource attributes from data

paulomarg:
## Description

Sometimes, the API will send back data that's not exactly as documented, such as the `optionX` fields in products / variants - those are documented as `option`, but they dynamically accept any number of options.

The current code only accepts fields that are in the "expected" set when loading API data, but in cases like the above, we should just take whatever the API is sending our way.

## How has this been tested?

Via a new unit test.

## Checklist:

### DIFF
--- a/lib/shopify_api/rest/base.rb
+++ b/lib/shopify_api/rest/base.rb
@@ -17,6 +17,8 @@ module ShopifyAPI
       @custom_prefix = T.let(nil, T.nilable(String))
       @read_only_attributes = T.let([], T.nilable(T::Array[Symbol]))
 
+      @aliased_properties = T.let({}, T::Hash[String, String])
+
       sig { returns(T::Hash[Symbol, T.untyped]) }
       attr_accessor :original_state
 
@@ -33,6 +35,7 @@ module ShopifyAPI
         @original_state = T.let({}, T::Hash[Symbol, T.untyped])
         @custom_prefix = T.let(nil, T.nilable(String))
         @forced_nils = T.let({}, T::Hash[String, T::Boolean])
+        @aliased_properties = T.let({}, T::Hash[String, String])
 
         session ||= ShopifyAPI::Context.active_session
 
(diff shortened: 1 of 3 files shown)


lib/shopify_api/rest/base.rb
```diff
@@ -313,6 +328,23 @@ def save(update_object: false)
         @errors.errors << e
         raise
       end
+
+      private
+
+      sig { params(var: T.any(String, Symbol), val: T.untyped).void }
+      def set_property(var, val)
+        clean = var.to_s.gsub(/[\?\s]/, "")
+        @aliased_properties[var.to_s] = clean if clean != var
```
paulomarg, line 337: Some of the fields returned by the API have special characters in them - I don't love this solution, but I couldn't figure out a better way of cleaning up the instance attributes.

lib/shopify_api/rest/base.rb
```diff
@@ -16,6 +16,8 @@ class Base
       @paths = T.let([], T::Array[T::Hash[Symbol, T.any(T::Array[Symbol], String, Symbol)]])
       @custom_prefix = T.let(nil, T.nilable(String))
 
+      @aliased_properties = T.let({}, T::Hash[String, String])
```
mllemango, line 19: does it make sense to add a comment here illustrating the purpose of ` @aliased_properties`?

mllemango (approved):
LGTM 
I'm having a hard time reading through some of these methods now that I've forgotten what a resource looks like. I wonder if it makes sense to unit test some of these methods as a way of documentation? For example, I had a hard time seeing what `create_instance` takes in as `data` or what to expect `method_missing` does.
(this seems outside of the scope of this PR)

lib/shopify_api/rest/base.rb
paulomarg: Yes, I think so.

paulomarg:
> I wonder if it makes sense to unit test some of these methods as a way of documentation? 

I think it may. Those methods are all indirectly tested by our suite, but if they're hard to grok having a test is probably useful.

teddyhwang (approved):

PR #926, opened 2022-04-11
Updating REST resources

paulomarg:
Updating the auto-generated REST resources after a couple of patch fixes were introduced.


mllemango (approved):
not sure whats up with the CLA :/

Issue #930, opened 2022-04-12
Why does `ShopifyAPI::Variant#to_hash` exclude `inventory_quantity` ?


koshigoe:
# Issue summary

I try to get variant's `inventory_quantity` via `ShopiryAPI::Product#to_hash`.
But `inventory_quantity` did not exist in `Hash` object.

```
> ShopifyAPI::Auth::Session.temp(shop: 'test.myshopify.com', access_token: 'token') do |session|
  variant = ShopifyAPI::Variant.create_instance(session: session, data: { inventory_item_id: 1234, inventory_quantity: 10 })
  pp variant.inventory_quantity
  pp variant.to_hash
end

10
{"inventory_item_id"=>1234}
=> {"inventory_item_id"=>1234}
```

## Expected behavior

I hope `ShopifyAPI::Variant#to_hash` include all API properties.

## Actual behavior

Some properties exclude?

## Specifications

- `shopify_api` version: 10.0.2
- Shopify API version used `2022-01`

paulomarg:
Hey @koshigoe, thanks for reporting this! The `inventory_quantity` field can't be sent when saving, so we started filtering it out when we call `to_hash` (because that is the method we use to serialize the object for saving). In your case, we shouldn't be ignoring that field.

We'll add a parameter to `to_hash` to make that behaviour optional. That way, when you call the method you'll get all the data.

PR #941, opened 2022-04-22
Fixes `to_hash` to include readonly attributes

mkevinosullivan:
## Description

Fixes issue #930

Adds parameter to `to_hash` so that default behaviour includes readonly attributes; setting the parameter to `true` indicates that `to_hash` is being used to serialize the object for saving (and thus excludes readonly attributes).

## How has this been tested?

Added tests to verify that readonly attributes are included when using the default `to_hash` and excluded when using `to_hash(true)` (i.e, `saving = true`).

## Checklist:

### DIFF
--- a/lib/shopify_api/rest/base.rb
+++ b/lib/shopify_api/rest/base.rb
@@ -116,10 +116,16 @@ module ShopifyAPI
           @has_many.include?(attribute)
         end
 
+        sig { returns(T::Hash[Symbol, Class]) }
+        attr_reader :has_many
+
         sig { params(attribute: Symbol).returns(T::Boolean) }
         def has_one?(attribute)
           @has_one.include?(attribute)
         end
+
+        sig { returns(T::Hash[Symbol, Class]) }
+        attr_reader :has_one
 
         sig { returns(T.nilable(T::Array[Symbol])) }
         def read_only_attributes
@@ -261,8 +267,8 @@ module ShopifyAPI
         match.nil? ? true : super
       end
 
-      sig { returns(T::Hash[String, T.untyped]) }
-      def to_hash
+      sig { params(saving: T::Boolean).returns(T::Hash[String, T.untyped]) }
+      def to_hash(saving = false)
         hash = {}
         instance_variables.each do |var|
           next if [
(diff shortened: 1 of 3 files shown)


lib/shopify_api/rest/base.rb
```diff
@@ -261,8 +261,8 @@ def respond_to_missing?(meth_id, *args)
         match.nil? ? true : super
       end
 
-      sig { returns(T::Hash[String, T.untyped]) }
-      def to_hash
+      sig { params(saving: T::Boolean).returns(T::Hash[String, T.untyped]) }
+      def to_hash(saving = false)
```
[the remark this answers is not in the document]
eliasrumley: Actually, thinking on this, I may be missing context though. If the intent is to actually change the default behaviour, then ignore my comment.

lib/shopify_api/rest/base.rb
```diff
@@ -116,11 +116,17 @@ def has_many?(attribute)
           @has_many.include?(attribute)
         end
 
+        sig { returns(T::Hash[Symbol, Class]) }
+        attr_reader :has_many
```
paulomarg, line 120: I wonder if it makes sense to declare all of the readers around L56 above so they're more grouped up?

lib/shopify_api/rest/base.rb
```diff
@@ -283,10 +289,16 @@ def to_hash
           end.to_sym
 
           if self.class.has_many?(attribute)
-            hash[attribute.to_s] = get_property(attribute).map(&:to_hash).to_a if get_property(attribute)
+            hash[attribute.to_s] = get_property(attribute).map do |element|
+              data = element
```
paulomarg, line 293: Nit: is it necessary to rename `element` here?

paulomarg (approved):
LGTM, just had nits.

lib/shopify_api/rest/base.rb
mkevinosullivan: I've refactored this to be a private method, so if you could take another look, that would be great (the Sorbet stuff confused me, so I'm not sure I've done it correctly).

paulomarg (approved):
Great refactorings!

PR #947, opened 2022-04-28
bad keyword

bashrj:
This is turning out to be an unreliable guide


mkevinosullivan:
Fixed in https://github.com/Shopify/shopify_api/pull/958

PR #951, opened 2022-05-09
Adds utility methods to find current shop, current recurring application charge

### LINKED ISSUE #923: Add `.current` method back to ShopifyAPI::Shop and ShopifyAPI::RecurringApplicationCharge
# Issue summary

With the new breaking api change, `ShopifyAPI::Shop.current`, `ShopifyAPI::RecurringApplicationCharge.current` are missing. 
Is there any plan in adding these back?

## Expected behavior

`ShopifyAPI::Shop.current` should return the current shop from the session. 

## Actual behavior

Right now it returns 
`NoMethodError: undefined method 'current' for ShopifyAPI::Shop:Class`

```
Shop.first.with_shopify_session do
  puts ShopifyAPI::Shop.current
end 
```

## Specifications

- `shopify_api` version: 10.0.0
- Shopify API version used (example: `2022-04`): 2022-04

mkevinosullivan:
## Description

Adds utility methods to find current shop and current recurring application charge, functionality that had been available in v9 of API when it used `ActiveResource`.

Fixes #923

## How has this been tested?

Unit tests added for each new method `current_shop`, `current_recurring_application_charge`

## Checklist:


test/utils_test.rb
```diff
@@ -0,0 +1,118 @@
...
+    def setup
+      super
+
+      test_session = ShopifyAPI::Auth::Session.new(
+        id: "id",
+        shop: "test-shop.myshopify.io",
+        access_token: "this_is_a_test_token"
+      )
+      ShopifyAPI::Context.activate_session(test_session)
+      modify_context(api_version: "2022-04")
```
paulomarg, line 19: Complete nit, but we probably want to do this before activating the session in case we ever change it to reset the session.

lib/shopify_api/utils.rb
```diff
@@ -0,0 +1,22 @@
...
+module ShopifyAPI
+  module Utils
+    extend T::Sig
+
+    sig do
+      params(
+        fields: T.untyped
+      ).returns(T.nilable(ShopifyAPI::Shop))
+    end
+    def self.current_shop(fields: nil)
```
paulomarg, line 13: We could also have the session be passed in as an optional argument to these, so we're not restricted to using `activate_session` only.

test/utils_test.rb
```diff
@@ -0,0 +1,118 @@
...
+        .to_return(status: 200, body: load_fixture("recurring_application_charges"), headers: {})
+
+      current_recurring_application_charge = ShopifyAPI::Utils.current_recurring_application_charge()
+
+      assert_requested(:get, "https://test-shop.myshopify.io/admin/api/2022-04/recurring_application_charges.json")
+      assert_equal(455696194, current_recurring_application_charge.id)
+    end
+
+    sig { void }
+    def test_current_recurring_application_charge_no_active
```
paulomarg, line 94: I like this test, we should probably have a similar one for the shop method!

schonert:
Any news on this? 👀

paulomarg:
Hey, thanks for raising this, I think you have a point in that this was an unannounced removal. I'll reopen and add this back to our tracking so we can finalize it.

PR #957, opened 2022-05-13
Gift card floating point values

### LINKED ISSUE #955: Gift card balance and initial_value in v10
It looks like `GiftCard` is using `T.let(nil, T.nilable(Float))` for `initial_value` and `balance` fields.

However, the API itself actually returns a `String` for those fields, not a `Float`. This causes `create_instance` to fail when using the response from the API as the `data`.

stevenou:
## Description

Fixes #955 

The gift card API returns string values for `balance` and `initial_value` (e.g. "25.00"). However, the `ShopifyAPI::GiftCard` class specifies `returns(T.nilable(Float))`. When an instance of `ShopifyAPI::GiftCard` is created using the response from the API, calling `gift_card.balance` or `gift_card.initial_value` results in an error like so:

`TypeError: Return value: Expected type T.nilable(Float), got type String with value "120.00"`

## How has this been tested?

Enhanced test suite to assert that `balance` and `initial_value` methods can returns the expected values.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+- [#957](https://github.com/Shopify/shopify_api/pull/957) Allow GiftCard to have String `balance` and `initial_value` (because API response is String)
+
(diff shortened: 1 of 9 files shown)


andyw8:
Closing for same reason as discussed in: https://github.com/Shopify/shopify_api/pull/956#issuecomment-1128054054

PR #958, opened 2022-05-15
Update argument in executing GraphQL query

remy727:
## Description
This is for fixing `GraphQL` usage documentation.


mllemango (approved):
good catch, thank you!
Looks like the CLA still needs to be signed

mllemango:
Thanks again!

Issue #952, opened 2022-05-12
REST Instructions lack reference to important configuration element.


Fladdermuz:
# Issue summary

Write a short description of the issue here ↓
Your example assumes in the docs for REST API that no one is using a private app. Had to work for hours trying to figure out I needed to put 
  private_shop: ENV.fetch('SHOPIFY_APP_PRIVATE_SHOP', nil), in the context set up.

## Expected behavior

What do you think should happen?
Would be really helpful to include that in the example. Not obvious what is needed.

## Actual behavior

What actually happens?

Tip: include an error message (in a `<details></details>` tag) if your issue is related to an error

## Steps to reproduce the problem

1.
1.
1.

## Logs
If applicable, enable the logs as described in the README, and paste the relevant portion here.

## Specifications

- `shopify_api` version:
- Shopify API version used (example: `2020-07`):

awd:
Why use an env for this, isn't the app locked into private/public mode - a simple true or false would be fine, and is in the documentation currently.

johnpash:
@awd, you might be thinking of the `is_private` attribute. `private_shop` is the (poorly named, imho) shop name such as shoppityshop.myshopify.com. The example given has left this out, but nothing will work without it.

Fladdermuz:
Adding this to your example for Context.setup  would be super helpful:
 private_shop: "myshop.myshopify.com", # Used for private app connection, etc

PR #991, opened 2022-07-05
Validate HMAC using both old and new API secrets

### LINKED ISSUE #990: Validate OAuth callback with both old and new API secrets
# Issue summary

Similar to https://github.com/Shopify/shopify-api-ruby/issues/978, this library needs the ability to verify the `hmac` signature of the query string in the OAuth 2.0 callback with a new and old API secrets in order to support [API key rotation](https://shopify.dev/apps/auth/oauth/rotate-revoke-api-credentials).

The verification logic is currently implemented in [`ShopifyAPI::Utils::HmacValidator.validate`](https://github.com/Shopify/shopify-api-ruby/blob/v11.0.0/lib/shopify_api/utils/hmac_validator.rb#L27) where it takes `ShopifyAPI::Context.api_secret_key` as the secret key. My suggestion is to add `old_api_secret_key` as a backup like [omniauth-shopify-oauth2 (v2.3.2)](https://github.com/Shopify/omniauth-shopify-oauth2/blob/v2.3.2/lib/omniauth/strategies/shopify.rb#L73) does so.

## Expected behavior

`hmac` signature can be verified with both the new and old API secrets.

## Actual behavior

`hmac` signature is only valid if it's signed with the new API secret.

- `shopify_api` version: 11.0.0
- Shopify API version used (example: `2020-07`): Any

tatsuya:
## Description

Fixes #990

Similar to #952 978, this library needs the ability to verify the `hmac` signature of the query string in the OAuth 2.0 callback using both ew and old API secrets in order to support [API key rotation](https://shopify.dev/apps/auth/oauth/rotate-revoke-api-credentials).

## How has this been tested?

Added new test cases to validate the signature signed with the old secret.

## Checklist:

### DIFF
--- a/test/utils/hmac_validator_test.rb
+++ b/test/utils/hmac_validator_test.rb
@@ -6,47 +6,10 @@ require_relative "../test_helper"
 module ShopifyAPITest
   module Utils
     class HmacValidatorTest < Test::Unit::TestCase
-      def test_invalid_signature
-        auth_query = ShopifyAPI::Auth::Oauth::AuthQuery.new(
-          code: query[:code],
-          shop: query[:shop],
-          timestamp: query[:timestamp],
-          state: query[:state],
-          host: query[:host],
-          hmac: "invalid"
-        )
-        refute(ShopifyAPI::Utils::HmacValidator.validate(auth_query))
-      end
+      def setup
+        super
 
-      def test_valid_signature
-        auth_query = ShopifyAPI::Auth::Oauth::AuthQuery.new(
-          code: query[:code],
-          shop: query[:shop],
-          timestamp: query[:timestamp],
-          state: query[:state],
-          host: query[:host],
-          hmac: hmac(query, ShopifyAPI::Context.api_secret_key)
-        )
-        assert(ShopifyAPI::Utils::HmacValidator.validate(auth_query))
-      end
-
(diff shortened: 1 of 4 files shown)


test/utils/hmac_validator_test.rb
```diff
@@ -7,39 +7,60 @@ module ShopifyAPITest
...
       def test_invalid_signature
-        query = ShopifyAPI::Auth::Oauth::AuthQuery.new(
-          code: "somecode",
-          shop: "some-shop.myshopify.com",
-          state: "1234",
-          timestamp: "123456",
-          hmac: "invalid",
-          host: "host"
+        auth_query = ShopifyAPI::Auth::Oauth::AuthQuery.new(
+          code: query[:code],
```
paulomarg, line 11: Not really a problem since this is just test code, but we're calling the `query` method quite a few times here. I wonder if it would be worth assigning the value to a variable so we're not calling the method several times for each test.

paulomarg (approved):
Good stuff - only had a nit but not necessarily blocking!

Also, quick question: do we need to do the same for webhook hmac validation? Or will webhooks never be signed with the old key?

test/utils/hmac_validator_test.rb
tatsuya: Updated!

PR #1000, opened 2022-07-31
Shipping zone fix

hoppergee:
## Description

- Update `ShippingZone#price_based_shipping_rates`'s type from Hash to Array
- Update `ShippingZone#weight_based_shipping_rates`'s type from Hash to Array

Otherwise, I will get below error when access those two attributes.

```
Return value: Expected type T.nilable(T::Hash[T.untyped, T.untyped]), got type Array with value ...
```


mkevinosullivan:
@hoppergee, thanks for bringing this to our attention.

However, as mentioned at the top of these files, they are auto-generated.  Therefore, it will require internal schema file changes for this to propagate properly to the API.

Please create an issue for tracking purposes.

PR #1002, opened 2022-08-05
Add new method to construct host app URL

mkevinosullivan:
## Description

For performance soon we will be recommending that apps favour server redirects within the OAuth process. This will replace client side redirects using AppBridge. We want to control that code that constructs the URL for the app. That way it's easier for developers, and if that URL changes we can update the logic with a version bump.

This utility will be used to get the URL to redirect to when the store does not have a token for the app.

Ruby API equivalent of https://github.com/Shopify/shopify-api-node/pull/419
Closes https://github.com/Shopify/first-party-library-planning/issues/392

## What is this pull request doing?

Add a new utility called `ShopifyAPI::Auth.embedded_app_url`. This utility constructs the host URL for an embedded app.

## How has this been tested?

Unit tests added to test green path as well as error paths.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Note: For changes to the API, see https://shopify.dev/changelog?filter=api
 
 ## Unreleased
 
-- [#1002](https://github.com/Shopify/shopify-api-ruby/pull/1002) Add new method to construct the host app URL
+- [#1002](https://github.com/Shopify/shopify-api-ruby/pull/1002) Add new method to construct the host app URL for an embedded app, allowing for safer redirect to app inside appropriate shop admin
 
 ## Version 11.0.1
 


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ Note: For changes to the API, see https://shopify.dev/changelog?filter=api
 
 ## Unreleased
 
-N/A
+- [#1002](https://github.com/Shopify/shopify-api-ruby/pull/1002) Add new method to construct the host app URL
```
paulomarg, line 7: Should we mention that this enables apps to redirect back to the appropriate Shopify surface, so it's safer than `shop`?

lib/shopify_api/auth.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module ShopifyAPI
+  module Auth
+    extend T::Sig
+
+    class << self
+      extend T::Sig
+
+      sig { params(host: T.nilable(String)).returns(String) }
+      def embedded_app_url(host)
```
paulomarg, line 12: I think `ShopifyAPI::Auth.embedded_app_url` is a sane name, wondering if others feel differently.

paulomarg (approved):
🔥

PR #1008, opened 2022-08-22
Increase session token JWT leeway

paulomarg:
### WHY are these changes introduced?

We currently have a 5s grace period for validating JWTs from session token, but there are cases where that is not long enough and we still end up failing to validate the token.

### WHAT is this pull request doing?

Increase the period to 10s which seems to eliminate the problem for good.

## Type of change

- [x] Patch: Bug (non-breaking change which fixes an issue)

## Checklist

- [x] I have added a changelog entry, prefixed by the type of change noted above

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Note: For changes to the API, see https://shopify.dev/changelog?filter=api
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


tatsuya (approved):
🚀

mkevinosullivan (approved):

PR #1017, opened 2022-09-21
Add `ShopifyAPI::Context.host` to allow for localhost support

nelsonwittwer:
## Description

The default CLI is to use `localhost` for OAuth instead of using a tunnel 🎉 . We have an assumption throughout the codebase that the `https` scheme is used. In order to get `localhost` to work with this gem, we need remove that assumption.

This adds the following:
- Adds the `ShopifyAPI::Context.host` which will include the scheme, name, and port of the host app
- Adds the `ShopifyAPI::Context.host_scheme` helper to return `http` or `https` for the few edge cases when we'd need it

The `host` is set by the configuration, ENV variable, or deduced from `host_name`.

This will work with backward compatibility and won't need updated configs to work. We will update [shopify_app](https://github.com/Shopify/shopify_app)'s generated config to include `host`.

## How has this been tested?

I'll be developing another PR in the [shopify_app gem](https://github.com/Shopify/shopify_app) in tandem with this. I'll ensure that I can get localhost as well as tunnel with https to work with this gem.

## Checklist:


lib/shopify_api/webhooks/registrations/http.rb
```diff
@@ -9,12 +9,12 @@ class Http < Registration
 
         sig { override.returns(String) }
         def callback_address
-          if @path.match?(%r{^https://})
+          if @path.match?(%r{^https://}) || @path.match?(%r{^http://})
```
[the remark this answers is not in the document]
paulomarg: Also, pedantry:
```suggestion
          if @path.match?(%r{^https?://})
```

test/context_test.rb
```diff
@@ -111,6 +112,24 @@ def test_unsupported_api_version
...
+      ShopifyAPI::Context.setup(
+        api_key: "key",
+        api_secret_key: "secret",
+        api_version: "unstable",
+        host_name: "host",
+        scope: ["scope1", "scope2"],
+        is_private: true,
+        is_embedded: true,
+        session_storage: ShopifyAPI::Auth::FileSessionStorage.new,
+        is_https: false,
```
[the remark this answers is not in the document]
paulomarg: That's true, when using the CLI we'll be fetching these 2 values from the `HOST` env var, so it'd be cleaner to just parse the scheme we get itself rather than checking it for a boolean.

Also, because this came back to bite us on the PHP template - will `URI().host` include the port in an address like `localhost:1234`?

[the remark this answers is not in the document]
nelsonwittwer: Ya, I went back and forth on this one. I initially had it as `hots_scheme` as you've outlined, but then I had to think about validation logic for making sure the provided value was valid (was one of `http://` or `https://`) and thought the boolean config was easier. The drift is a pain though..

I do like the idea of parsing from `HOST`, as that is fewer things to configure / validate. Can we assume that env variable will always be there? I'm not seeing that in the readme for the API like it is for `shopify_app`.

lib/shopify_api/auth.rb
```diff
@@ -19,7 +19,7 @@ def embedded_app_url(host)
         end
 
         decoded_host = Base64.decode64(host)
-        "https://#{decoded_host}/apps/#{Context.api_key}"
+        "#{Context.host_scheme}://#{decoded_host}/apps/#{Context.api_key}"
```
paulomarg, line 22: I think this needs to stay as https - this is us redirecting back to Shopify at this point, which would always be https.

lib/shopify_api/context.rb
```diff
@@ -131,7 +134,7 @@ def embedded?
 
       sig { returns(T::Boolean) }
       def setup?
-        !(api_key.empty? || api_secret_key.empty? || host_name.empty?)
+        !(api_key.empty? || api_secret_key.empty? || host_name.empty? || T.must(host).empty?)
```
paulomarg, line 137: Just checking that `host` will always be present here - even if only `host_name` is passed in?

lib/shopify_api/context.rb
```diff
@@ -51,6 +52,7 @@ def setup(
         is_embedded:,
         session_storage:,
         logger: Logger.new($stdout),
+        host: ENV["HOST"] || "https://#{host_name}",
```
andyw8, line 55: Hmm, the naming doesn't feel right here, I wouldn't expect `host` to include a protocol.

lib/shopify_api/context.rb
```diff
@@ -150,6 +153,11 @@ def activate_session(session)
       def deactivate_session
         @active_session.value = nil
       end
+
+      sig { returns(String) }
+      def host_scheme
+        T.must(URI.parse(T.must(host)).scheme)
```
paulomarg, line 159: Would it make sense to store `host` as a `URI` to make it easier to handle later?

paulomarg (approved):
Just nits and questions!

lib/shopify_api/context.rb
```diff
@@ -64,6 +66,7 @@ def setup(
         @api_secret_key = api_secret_key
         @api_version = api_version
         @host_name = host_name
+        @host = T.let(host, T.nilable(String))
```
andyw8, line 69: Can `@host` ever be nil here?

lib/shopify_api/context.rb
andyw8, line 137: I know it's not knew but my brain can't parse conditionals like that, may we can De Morgan it to:

`!api_key.empty? && !api_secret_key.empty? && !host_name.empty? && !T.must(host).empty?)`

which reduces to:

`api_key.present? && api_secret_key.present? && host_name.present? && T.must(host).present?)`

or even

`[api_key, api_secret_key, host_name, T.must(host)].all?(&:present?)`

lib/shopify_api/webhooks/registrations/http.rb
```diff
@@ -9,12 +9,12 @@ class Http < Registration
 
         sig { override.returns(String) }
         def callback_address
-          if @path.match?(%r{^https://})
```
andyw8, line 12: Just an observation, not about your PR, but it's odd that we call this `@path` when it seems to be a URI.

lib/shopify_api/context.rb
andyw8: (maybe call it `host_with_scheme`?)

andyw8 (commented):
All looks good in terms of code – I think some of the names could be more precise, but I understand we may locked into some things for backwards compatibility.

lib/shopify_api/context.rb
nelsonwittwer: That's right, in the context we'll assign the default for `host` like so:

```ruby
host: ENV["HOST"] || "https://#{host_name}"
```

lib/shopify_api/context.rb
nelsonwittwer: In the initializer, yes but the default gets set in the `setup` method to build it from the `host_name` if it is nil.

lib/shopify_api/context.rb
nelsonwittwer: good call on this goofy conditional callout. I don't love it either...

`present?` is an active support helper which if I recall, we are trying to remove from this library as a dependency.

lib/shopify_api/webhooks/registrations/http.rb
nelsonwittwer: ya, as I worked on this I noticed a lot of inconsistencies like that 🤦

lib/shopify_api/context.rb
nelsonwittwer: I agree, the naming here isn't as precise as I'd hope. We use `host` name to imply the needed scheme. For that consistency I think we should leave it as is :(

lib/shopify_api/context.rb
andyw8: Ah, I had misread because of the way GitHub had shown the diff.

resistorsoftware:
This is busted in my life. Here is why. I use Puma Dev to name my localhost. So I have a name and port. It works! So to Shopify App, my App is configured as https://myapp.test and that works fine, as under the hood, this gets converted to an ngrok tunnel and away we go. But you cannot use this host for webhooks. Nope. For webhooks to work you need to provide an actual reachable URI which is https://myapp.ngrok.io 

Unfortunately, this change with webhooks is now writing all my localhost webhooks as https://myapp.test//webooks/app_uninstalled as an example, which is not reachable. So I am screwed now when my older working pattern is used with this new code. 

Why is it this way? I dunno. But it is. Is there a way to register a webhook and not have it play havoc with the provided callback URL? 

In the Shopify App if I use this:

     config.webhooks = [
        {topic: "app/uninstalled", path: "webhooks/app_uninstalled"}
      ]

The webhook I find registered in the store is https://myapp.test/webhooks/app_uninstalled which is unreachable. That is because my host is indeed myapp.test. But the reality is, webhooks work perfect when I use a manual setup and set the callback URL to be https://myapp.ngrok.io  

This is frustrating.

PR #1023, opened 2022-09-28
Allow custom scopes during the auth process

andyw8:
## Description

First party apps such as https://shopify-graphiql-app.shopifycloud.com have scopes which are set during installation, rather than being preconfigured.

There is a corresponding update to `shopify_app`: https://github.com/Shopify/shopify_app/pull/1540

## How has this been tested?

Tested as part of the upgrade of https://shopify-graphiql-app.shopifycloud.com

## Checklist:

- [x] ~I have updated the project documentation~ N/A

### DIFF
--- a/lib/shopify_api/auth/oauth.rb
+++ b/lib/shopify_api/auth/oauth.rb
@@ -16,9 +16,18 @@ module ShopifyAPI
             shop: String,
             redirect_path: String,
             is_online: T.nilable(T::Boolean),
+            scope_override: T.nilable(T.any(ShopifyAPI::Auth::AuthScopes, T::Array[String], String)),
           ).returns(T::Hash[Symbol, T.any(String, SessionCookie)])
         end
-        def begin_auth(shop:, redirect_path:, is_online: true)
+        def begin_auth(shop:, redirect_path:, is_online: true, scope_override: nil)
+          scope = if scope_override.nil?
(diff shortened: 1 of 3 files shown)


lib/shopify_api/auth/oauth.rb
```diff
@@ -16,9 +16,10 @@ class << self
             shop: String,
             redirect_path: String,
             is_online: T.nilable(T::Boolean),
+            scope: AuthScopes,
```
andyw8, line 19: This should really be called `scopes` but we're already using `scope` to refer to the collection in multiple places, so I've followed that convention.

lib/shopify_api/auth/oauth.rb
```diff
@@ -16,9 +16,10 @@ class << self
             shop: String,
             redirect_path: String,
             is_online: T.nilable(T::Boolean),
+            scope: ShopifyAPI::Auth::AuthScopes,
```
andyw8, line 19: This should really be called scopes but we're already using scope to refer to the collection in multiple places, so I've followed that pattern.

paulomarg (approved):
LGTM, only non-blocking comments.

lib/shopify_api/auth/oauth.rb
andyw8: 👍 updated

PR #1025, opened 2022-10-02
Update README.md for Custom Apps and Setup

caseyprovost:
## Description

Custom apps have a different credential structure than the deprecated private apps. The AccessToken has replaced the old `api_secret_key` value.

Private apps were deprecated, and instructions on configuring Setup properly for regular custom apps were never added.

## How has this been tested?
* Tested by integrating with Shopify with both Rest and GQL on version `2022-07`, gem version: `11.1.0`

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ ShopifyAPI::Context.setup(
 )
 ```
 
+*Important*: If you are using custom apps you will need to use the value of your access_token instead of api_secret_key
+
 ### Setup a Session Store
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,6 +61,8 @@ ShopifyAPI::Context.setup(
 )
 ```
 
+*Important*: If you are using custom apps you will need to use the value of your access_token instead of api_secret_key
```
caseyprovost, line 64: When I was forced to migrate from private to custom apps this weekend, this little booger caused hours of confusion.

PR #1057, opened 2022-11-11
Clarify statement about packages

andyw8:
Since there are still a number of dependencies in the gemspec, this seemed a little misleading.


andyw8:
CI is failing due to https://github.com/Shopify/shopify-api-ruby/issues/1058 but since this only impacts the docs it's safe to override.

PR #1069, opened 2022-11-22
Add Logger to library

klenotiw:
## Description

Fixes #<issue-number>

## How has this been tested?

## Checklist:

### DIFF
--- a/lib/shopify_api/context.rb
+++ b/lib/shopify_api/context.rb
@@ -12,7 +12,8 @@ module ShopifyAPI
     @is_private = T.let(false, T::Boolean)
     @private_shop = T.let(nil, T.nilable(String))
     @is_embedded = T.let(true, T::Boolean)
-    @logger = T.let(Logger.new($stdout), Logger)
+    @logger = T.let(::Logger.new($stdout), ::Logger)
+    @log_level = T.let(:error, Symbol)
     @notified_missing_resources_folder = T.let({}, T::Hash[String, T::Boolean])
     @active_session = T.let(Concurrent::ThreadLocalVar.new { nil }, Concurrent::ThreadLocalVar)
     @session_storage = T.let(nil, T.nilable(ShopifyAPI::Auth::SessionStorage))
@@ -32,7 +33,8 @@ module ShopifyAPI
           scope: T.any(T::Array[String], String),
           is_private: T::Boolean,
           is_embedded: T::Boolean,
-          logger: Logger,
+          log_level: Symbol,
+          logger: ::Logger,
           session_storage: T.nilable(ShopifyAPI::Auth::SessionStorage),
           host_name: T.nilable(String),
           host: T.nilable(String),
@@ -48,7 +50,8 @@ module ShopifyAPI
         scope:,
         is_private:,
         is_embedded:,
-        logger: Logger.new($stdout),
+        log_level: :error,
+        logger: ::Logger.new($stdout),
         session_storage: nil,
         host_name: nil,
         host: ENV["HOST"] || "https://#{host_name}",
@@ -70,15 +73,16 @@ module ShopifyAPI
         @is_embedded = is_embedded
         @session_storage = session_storage
         if @session_storage
-          ::ShopifyAPI::Context.logger.warn("The use of SessionStorage in the API library has been deprecated. " \
+          ::ShopifyAPI::Logger.deprecated("The use of SessionStorage in the API library has been deprecated. " \
             "The ShopifyAPI will no longer have responsibility for session persistence. " \
             "Upgrading to `shopify_app` 21.3 will allow you to remove session_storage" \
-            " from the API library Context configuration.")
+            " from the API library Context configuration.", "13.0.0")
         end
         @logger = logger
         @private_shop = private_shop
         @user_agent_prefix = user_agent_prefix
         @old_api_secret_key = old_api_secret_key
+        @log_level = log_level
 
         load_rest_resources(api_version: api_version)
       end
@@ -120,8 +124,11 @@ module ShopifyAPI
       sig { returns(T.nilable(ShopifyAPI::Auth::SessionStorage)) }
       attr_reader :session_storage
 
-      sig { returns(Logger) }
+      sig { returns(::Logger) }
       attr_reader :logger
+
+      sig { returns(Symbol) }
+      attr_accessor :log_level
 
       sig { returns(T::Boolean) }
       def private?
--- a/lib/shopify_api/errors/log_level_not_found_error.rb
+++ b/lib/shopify_api/errors/log_level_not_found_error.rb
@@ -1,0 +1,9 @@
+# typed: strict
+# frozen_string_literal: true
+
+module ShopifyAPI
+  module Errors
+    class LogLevelNotFoundError < StandardError
+    end
+  end
+end
--- a/lib/shopify_api/logger.rb
+++ b/lib/shopify_api/logger.rb
@@ -1,0 +1,54 @@
+# typed: false
+# frozen_string_literal: true
+
+module ShopifyAPI
+  class Logger
+    LOG_LEVELS = { debug: 0, info: 1, warn: 2, error: 3, off: 6 }
+    DEFAULT_LOG_LEVEL = :info
+
+    class << self
+      def send_to_logger(log_level, message)
+        return unless enabled_for_log_level?(log_level)
+
+        full_message = "#{context(log_level)} #{message}"
+
+        ShopifyAPI::Context.logger.send(log_level, full_message)
+      end
+
+      def debug(message)
+        send_to_logger(:debug, message)
+      end
+
+      def info(message)
+        send_to_logger(:info, message)
+      end
+
+      def warn(message)
+        send_to_logger(:warn, message)
+      end
+
+      def error(message)
+        send_to_logger(:error, message)
+      end
+
+      def deprecated(message, version)
+        return unless enabled_for_log_level?(:warn)
+
+        raise Errors::FeatureDeprecatedError unless ShopifyAPI::VERSION < version
+
+        send_to_logger(:warn, message)
+      end
+
+      def context(log_level)
+        current_shop = ShopifyAPI::Context.active_session&.shop || "Shop Not Found"
+        "[ ShopifyAPI | #{log_level.to_s.upcase} | #{current_shop} ]"
+      end
+
+      def enabled_for_log_level?(log_level)
+        raise(Errors::LogLevelNotFoundError,
+          "Invalid Log Level - #{log_level}") unless LOG_LEVELS.keys.include?(log_level)
+        LOG_LEVELS[log_level] >= LOG_LEVELS[ShopifyAPI::Context.log_level || DEFAULT_LOG_LEVEL]
+      end
+    end
+  end
+end
--- a/lib/shopify_api/utils/session_utils.rb
+++ b/lib/shopify_api/utils/session_utils.rb
@@ -81,6 +81,7 @@ module ShopifyAPI
           if Context.embedded?
             if auth_header
               matches = auth_header.match(/^Bearer (.+)$/)
+              #add log here
               raise Errors::MissingJwtTokenError,
                 "Missing Bearer token in authorization header" unless matches
 
--- a/test/context_test.rb
+++ b/test/context_test.rb
@@ -17,6 +17,7 @@ module ShopifyAPITest
         is_private: true,
         is_embedded: true,
         session_storage: ShopifyAPI::Auth::FileSessionStorage.new,
+        log_level: :warn,
         logger: Logger.new(writer),
         private_shop: "privateshop.myshopify.com",
         user_agent_prefix: "user_agent_prefix1",
--- a/test/logger_test.rb
+++ b/test/logger_test.rb
@@ -1,0 +1,34 @@
+# typed: true
+# frozen_string_literal: true
+
+require_relative "./test_helper"
+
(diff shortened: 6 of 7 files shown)


test/logger_test.rb
```diff
@@ -0,0 +1,34 @@
+# typed: true
+# frozen_string_literal: true
+
+require_relative "./test_helper"
+
+module ShopifyAPITest
+  class LoggerTest < Minitest::Test
+    def test_call_warn
```
klenotiw, line 8: Is this a useful test? It kind of just tests if `send_to_logger` was called the correct message and log_level. I didn't make more of these because I wanted to see what we thought of this. (#still-a-noob)

lib/shopify_api/logger.rb
```diff
@@ -0,0 +1,54 @@
+# typed: false
+# frozen_string_literal: true
+
+module ShopifyAPI
+  class Logger
```
nelsonwittwer, line 5: I think we are missing a `private` keyword here to establish a public/private interface. Where do you think a good place to draw that line is?

lib/shopify_api/utils/session_utils.rb
```diff
@@ -81,6 +81,7 @@ def current_session_id(auth_header, cookies, online)
           if Context.embedded?
             if auth_header
               matches = auth_header.match(/^Bearer (.+)$/)
+              #add log here
```
nelsonwittwer, line 84: *jon_travolta_looking_around.gif* 😂

test/logger_test.rb
nelsonwittwer: Good question! IMHO `send_to_logger` should be a private method so this would break my rule of expecting messages from ourselves. I'd put the expectation on the context logger:

```ruby
ShopifyAPI::Context.logger.expects(:warn, "test")
```

The behavior we are looking at testing is that our logger should be warning with the string of "test" so that is what the test is here.

If you ended up leaving `send_to_logger` as a public method, then this test would be fine :) I'd argue it is a private function because you wouldn't call that method outside of the context of that class because it would require users of this class to know exactly how logging is wired together.

nelsonwittwer: Happy to pair on this if you'd like :) Testing is an abstract art that takes some experience shooting yourself in the foot to really get a grasp on

test/logger_test.rb
```diff
@@ -0,0 +1,34 @@
...
+      ShopifyAPI::Logger.expects(:send_to_logger).with(:warn, "test")
+      ShopifyAPI::Logger.warn("test")
+    end
+
+    def test_deprecated_should_raise_an_error_if_version_is_lower_or_equal_to_current
+      parent_log_level = ShopifyAPI::Context.log_level
```
nelsonwittwer, line 14: This would be a great use of stubbing `ShopifyAPI::Context.stubs(:log_level).returns(:warn)`. With that stub you don't have to worry about setting the log level back to level you found it in. Once it exits this method the stub will be ignored by the rest of the test.

I typically only capture current settings and put them back how I found them when working with environment variables since they don't have that same contextual rollback a stub has.

lib/shopify_api/logger.rb
```diff
@@ -0,0 +1,54 @@
...
+      end
+
+      def context(log_level)
+        current_shop = ShopifyAPI::Context.active_session&.shop || "Shop Not Found"
+        "[ ShopifyAPI | #{log_level.to_s.upcase} | #{current_shop} ]"
+      end
+
+      def enabled_for_log_level?(log_level)
+        raise(Errors::LogLevelNotFoundError,
+          "Invalid Log Level - #{log_level}") unless LOG_LEVELS.keys.include?(log_level)
```
nelsonwittwer, line 49: Should we do this check during `ShopifyAPI::Context.setup` instead?? 🤔 Since that is where the config lives, that feels like the right place to check to make sure it was setup correctly.

lib/shopify_api/errors/log_level_not_found_error.rb
```diff
@@ -0,0 +1,9 @@
+# typed: strict
+# frozen_string_literal: true
+
+module ShopifyAPI
+  module Errors
+    class LogLevelNotFoundError < StandardError
```
nelsonwittwer, line 6: I wonder if we should be raising an error if log level is misconfigured? Do we want to potentially break an API call or server instance because of a bad config? I'd think we'd want to just send a warning to logger but keep the app running. I'd hate to bring down someone's app because they couldn't tell they misconfigured something until it started processing requests

nelsonwittwer (commented):
Glad to see this functionality in the API! 🎉 🍻 Hopefully these changes don't take too long and make sense. If not, let me know and I'll be happy to help ya!

lib/shopify_api/errors/log_level_not_found_error.rb
andyw8: Perhaps only raise in dev/test environments?

lib/shopify_api/logger.rb
klenotiw: True I think `send_to_logger` and `enabled_for_log_level?` should probably be private. `deprecated` and `context` gets overridden in `shopify_app` so I think that needs to stay public.

lib/shopify_api/logger.rb
```diff
@@ -0,0 +1,54 @@
...
+      def deprecated(message, version)
+        return unless enabled_for_log_level?(:warn)
+
+        raise Errors::FeatureDeprecatedError unless ShopifyAPI::VERSION < version
+
+        send_to_logger(:warn, message)
+      end
+
+      def context(log_level)
+        current_shop = ShopifyAPI::Context.active_session&.shop || "Shop Not Found"
```
nelsonwittwer, line 43: I'm still looking into this, but `active_session` isn't used everywhere and I'd like to potentially deprecate. I'd think it would be fine to log it if we have it, but I wouldn't call attention to it by including "Shop Not Found" I'd just keep it nil.

lib/shopify_api/utils/session_utils.rb
```diff
@@ -81,9 +81,10 @@ def current_session_id(auth_header, cookies, online)
           if Context.embedded?
             if auth_header
               matches = auth_header.match(/^Bearer (.+)$/)
-              # add log here
-              raise Errors::MissingJwtTokenError,
-                "Missing Bearer token in authorization header" unless matches
+              unless matches
+                ShopifyAPI::Logger.info("Missing Bearer token in authorization header")
```
klenotiw, line 85: We wanted to add this because the error gets rescued later on

lib/shopify_api/context.rb
```diff
@@ -32,7 +33,8 @@ class << self
           scope: T.any(T::Array[String], String),
           is_private: T::Boolean,
           is_embedded: T::Boolean,
-          logger: Logger,
+          log_level: Symbol,
```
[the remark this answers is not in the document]
klenotiw: I don't think sorbet has an opinion. Would you rather this be a String?

lib/shopify_api/errors/log_level_not_found_error.rb
klenotiw: Is there a dev/test environment in ruby? Or is that just relevant for rails? I did a quick google and could only find things for rails. So I am just sending a log: 
```ruby
      sig { params(log_level: Symbol).returns(T::Boolean) }
      def valid_log_level?(log_level)
        return true if ::ShopifyAPI::Logger::LOG_LEVELS.keys.include?(log_level)

        ShopifyAPI::Logger.warn("#{log_level} is not a valid log_level. "\
          "Valid options are #{::ShopifyAPI::Logger::LOG_LEVELS.keys.join(", ")}")

        false
      end
```
Since the log level isn't set it may have to rely on the previous `setup` call or our default inside the logger which is `info`

test/logger_test.rb
klenotiw: Really good point I'll add that in!

lib/shopify_api/context.rb
```diff
@@ -167,6 +179,18 @@ def host_scheme
       def host_name
         T.must(URI(T.must(host)).host)
       end
+
+      private
+
+      sig { params(log_level: Symbol).returns(T::Boolean) }
+      def valid_log_level?(log_level)
+        return true if ::ShopifyAPI::Logger::LOG_LEVELS.keys.include?(log_level)
```
andyw8, line 187:
```
::ShopifyAPI::Logger:.valid_level?(log_level)`
```

andyw8, line 187:
```suggestion
        return true if ::ShopifyAPI::Logger:.valid_level?(log_level)`
```

lib/shopify_api/context.rb
```diff
@@ -167,6 +179,18 @@ def host_scheme
...
+
+        ShopifyAPI::Logger.warn("#{log_level} is not a valid log_level. "\
+          "Valid options are #{::ShopifyAPI::Logger::LOG_LEVELS.keys.join(", ")}")
```
andyw8, line 190:
```suggestion
          "Valid options are #{::ShopifyAPI::Logger.levels.join(", ")}")
```
This an the above change help to decouple from the internals of the logger.

lib/shopify_api/errors/log_level_not_found_error.rb
andyw8: > Is there a dev/test environment in ruby? 

Ah, no 🤦

test/context_test.rb
```diff
@@ -131,13 +132,33 @@ def test_with_host_name_and_no_host_env
...
       )
       assert_equal("https", ShopifyAPI::Context.host_scheme)
       assert_equal("https://tunnel-o-security.com", ShopifyAPI::Context.host)
       assert_equal("tunnel-o-security.com", ShopifyAPI::Context.host_name)
       ENV["HOST"] = old_host
     end
 
+    def test_send_a_warning_if_log_level_is_invalid
+      ShopifyAPI::Context.stubs(:log_level).returns(:warn)
+      ShopifyAPI::Logger.expects(:warn).with("not_a_level is not a valid log_level. "\
```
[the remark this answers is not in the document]
klenotiw: yeah I tried that, now its saying that warn is expecting a regexp that looks `/not a valid log_level/`. Maybe theres another syntax.

test/logger_test.rb
```diff
@@ -0,0 +1,20 @@
+# typed: true
+# frozen_string_literal: true
+
+require_relative "./test_helper"
```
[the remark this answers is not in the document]
klenotiw: yeah it doesn't load like that. The api uses "zeitwerk" so there might be some fancy things going on.

lib/shopify_api/utils/session_utils.rb
cquemin: It feels a bit strange to me that a raised error is translated into a log level `INFO`? I'd expect a `warn` at least if not more no?

lib/shopify_api/logger.rb
```diff
@@ -0,0 +1,70 @@
...
+      sig { params(message: String).void }
+      def error(message)
+        send_to_logger(:error, message)
+      end
+
+      sig { params(message: String, version: String).void }
+      def deprecated(message, version)
+        return unless enabled_for_log_level?(:warn)
+
+        raise Errors::FeatureDeprecatedError unless ShopifyAPI::VERSION < version
```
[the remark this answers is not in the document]
klenotiw: nvm "3.10" > "3.2" doesn't work

test/context_test.rb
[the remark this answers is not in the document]
cquemin: I think we test this way then we need to have this type of message or part of them define in a standard way. For consistency but also for testability. Even with the regexp here, someone might decide to change how the message is phrased and that should not break the test.

test/logger_test.rb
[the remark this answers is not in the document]
cquemin: Do you think we could find out why it is not already on the path?

lib/shopify_api/context.rb
```diff
@@ -69,16 +72,22 @@ def setup(
...
         @user_agent_prefix = user_agent_prefix
         @old_api_secret_key = old_api_secret_key
+        @log_level = if valid_log_level?(log_level)
+          log_level
+        else
+          :info
+        end
+
+        if @session_storage
+          ::ShopifyAPI::Logger.deprecated("The use of SessionStorage in the API library has been deprecated. " \
```
cquemin, line 86: Same point I made on the[ other repo](https://github.com/Shopify/shopify_app/pull/1581#discussion_r1030843509)

cquemin (commented):
Some question around consistency of deprecated log message/configuration as I already mentioned on [your other PR ](https://github.com/Shopify/shopify_app/pull/1581#discussion_r1030843509)

test/logger_test.rb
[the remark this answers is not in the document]
klenotiw: Yeah totally, do you think this is a blocker? Maybe we should make this a separate issue since this PR has been getting slowed down a bit?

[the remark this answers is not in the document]
cquemin: Mo, I don't think it's a blocker at all!
I agree with you to investigate this separately.

lib/shopify_api/context.rb
nelsonwittwer: I'm a fan of iterative improvement. This is the first time we've really used deprecation logging in our api / app gem. Can we ship this as is for now and if we want to invest in more feature rich IDE tooling in the future, we can do so then?

cquemin (approved):
Awesome work guys! This looks pretty good!

PR #1080, opened 2022-12-20
Resolve ActionView::Template::Error on Boolean Metafields values access using REST client

### LINKED ISSUE #1079: ActionView::Template::Error: Return value: Expected type T.nilable(T.any(Float, Integer, String)), got type TrueClass
# Issue summary
On accessing Object.value of boolean metafields we are getting https://d.pr/i/TGyM8r 

## Specifications
- `shopify_api` version: 12.2.1
- Shopify API version used (example: `2022-10`):
```[tasklist]

```

jaimina-hulkapps:
## Description
REST API client Metafields with boolean type values are not accessible due to Class type restrictions. Please review [this screenshot](https://d.pr/i/TGyM8r) for the reference.

Fixes #1079


nelsonwittwer:
Shoot, I forgot that these models are generated and these changes will be replaced next time we generate rest resources. I'll need to apply this change upstream at our generator logic. I'll keep this PR is open until I do so and I'll close the loop with you.

nelsonwittwer:
This will be fixed with #1137 as the problem needed to be solved upstream where we generate these files. Thanks for your help!

PR #1104, opened 2023-01-21
Allow api version override in Graphql query

ewalk153:
## Description

In order to develop again unstable features, it's important to be able to make particular queries against the unstable version. Furthermore, it's helpful to be able to adopt the latest version of the ApiVersion query by query while migrating versions.

## How has this been tested?

There is an automated test to confirm this has the desired behavior.

## Checklist:


ewalk153:
If we align on this approach, I’ll update the release notes.

mkevinosullivan (approved):
Some nits w.r.t. docs.

Also added some suggestions to log a debug message if the version is being overridden (informational for dev), or to log a debug message if it's being overridden by the same value as in `Context` (i.e., when it's unnecessary) ... aligns with Node library

Issue #1101, opened 2023-01-17
ShopifyAPI::GiftCard balance and initial_value fail with a TypeError


matthewlein:
# ShopifyAPI::GiftCard balance and initial_value fail with a TypeError

After fetching a GiftCard through `ShopifyAPI::GiftCard.find(id: id)`, I can see the balance and initial value are both strings
`<ShopifyAPI::GiftCard:0x0000000114166b78 ... @balance="1.29" @initial_value="1.29">`

Accessing those via `gift_card.initial_value` or `gift_card.balance` will cause and error. Sorbet (or whatever) expects them to be floats and raises

```
TypeError (Return value: Expected type T.nilable(Float), got type String with value "1.29")
```

- `shopify_api` version: 12.3.0
- Ruby version: 2.7.2
- Operating system: MacOS

## Expected behavior

I would expect to be able to access a gift card's balance or initial_value as either a string or float without an error

## Actual behavior

Errors are raised

## Steps to reproduce the problem

1. `gift_card = ShopifyAPI::GiftCard.find(id: id)`
1. `gift_card.balance`

matthewlein:
The not-ideal workaround is converting to json and then parsing to get the values.

jaredbeck:
I can reproduce this issue in 12.4.0. The workaround (I used `as_json`) worked for me.

matthewlein:
guh, this one was filed in May (!).

https://github.com/Shopify/shopify-api-ruby/issues/955

I wouldn't count on it getting fixed.

nelsonwittwer:
Fixed with https://github.com/Shopify/shopify-api-ruby/pull/1137

PR #1120, opened 2023-03-06
Fix Sorbet type signatures for gift cards

### LINKED ISSUE #955: Gift card balance and initial_value in v10

gavinballard:
## Description

Fixes #955, #1101

- As per the issues described in #955 and #1101, the `GiftCard` and `GiftCardAdjustment` resources incorrectly have value-related attributes typed as `Float` when they are handled as `String` within the REST API.
- While this PR is somewhat ineffective as the REST resources are automatically generated, opening it in the hope we can get this addressed upstream.

## How has this been tested?

Tested locally.

## Checklist:

### DIFF
--- a/lib/shopify_api/rest/resources/2023_01/gift_card_adjustment.rb
+++ b/lib/shopify_api/rest/resources/2023_01/gift_card_adjustment.rb
@@ -16,7 +16,7 @@ module ShopifyAPI
     def initialize(session: ShopifyAPI::Context.active_session)
       super(session: session)
 
-      @amount = T.let(nil, T.nilable(Float))
+      @amount = T.let(nil, T.nilable(String))
       @api_client_id = T.let(nil, T.nilable(Integer))
       @created_at = T.let(nil, T.nilable(String))
       @gift_card_id = T.let(nil, T.nilable(Integer))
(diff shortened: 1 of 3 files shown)


lib/shopify_api/rest/resources/2023_01/gift_card_adjustment.rb
```diff
@@ -16,7 +16,7 @@ class GiftCardAdjustment < ShopifyAPI::Rest::Base
     def initialize(session: ShopifyAPI::Context.active_session)
       super(session: session)
 
-      @amount = T.let(nil, T.nilable(Float))
+      @amount = T.let(nil, T.nilable(String))
```
nelsonwittwer, line 19: I believe float is the right type, we just need to cast it as such if we are provided a string as the argument.

gavinballard: @nelsonwittwer Thanks for taking a look at this!

Agree a numeric type would make sense - would `BigDecimal` be preferable to `Float` though for a current amount?

nelsonwittwer:
Digging through all the open issues yesterday, this problem of taking a string value from the API and casting it into a number type is breaking more resources than just gift cards :( 

I'm looking at solving this at the base layer to try and solve it everywhere

nelsonwittwer:
Thanks for your patience with this! I've tracked down the root problem and we are working on a fix! Unfortunately, the problem seems to be a type misunderstanding upstream with our API at Shopify so I'm not in complete control of the fix. I see this root issue affecting several other issues within our REST resources here which is a bummer :( 

I'm going to close this PR because the fix will be regenerating all of the rest resources. Thanks for being proactive with trying to provide a solution here!

PR #1138, opened 2023-04-11
Packaging for release 12.5.0

nelsonwittwer:
Shipping `v12.5.0`!

🎩 - I created a new app with the shopify CLI and pointed to this branch and ran `bundle install`. Verified the branch version of the gem was being used. Was able to install my test app on a store and make an API call.


paulomarg (approved):
🚢

Issue #1145, opened 2023-04-24
REST API Shop resource definitions TypeError


LovroM:
# Issue summary

When trying to access 'taxes_included' and 'tax_shipping' fields on the shop resource it fails because of the wrong sorbet type definition.
The Shopify documentation states that both fields should be booleans but in shop resource are defined as strings.
https://shopify.dev/docs/api/admin-rest/2023-04/resources/shop

## Expected behavior
```
>shopify_shop.taxes_included
>true

>shopify_shop.tax_shipping
>true

```

## Actual behavior
```
> shopify_shop.taxes_included
> TypeError (Return value: Expected type T.nilable(String), got type TrueClass)

> shopify_shop.tax_shipping
> TypeError (Return value: Expected type T.nilable(String), got type TrueClass)
```

PR #1146, opened 2023-04-24
Fix 'taxes_included' and 'tax_shipping' type definitions

LovroM:
## Description

Fixes [#1145 ](https://github.com/Shopify/shopify-api-ruby/issues/1145)


nelsonwittwer:
Thanks for providing this fix! Sadly, we won't be able to merge this in because this code is auto generated by a schema up stream that we need to fix. We've got it on our radar and are making the fix for the next release :) Thanks for your help!

Issue #1037, opened 2022-10-12
Useless HashDiff comparison when saving resource


kaarelss:
# Issue summary

When updating/saving any resource, there is useless HashDiff comparison. It looks like the intended logic is to not send attributes which are not changed, but all attributes are sent always, because comparison result contains duplicate attributes as symbols and as strings.
 
## Expected behavior

original_state should return hash with same attribute types as to_hash methods result, so the HashDiff comparison works correctly.

## Actual behavior

original_state return hash with keys as symbols, but to_hash return hash with keys as strings. Therefore the HashDiff comparison have both sets of attributes.

## Steps to reproduce the problem

1. Execute the code below
```
blog = ShopifyAPI::Blog.new(session: ShopifyAPI::Context.active_session)
blog.title = "My Blog"
blog.handle = "my-blog"
blog.save!

blog.handle = "my-blog-changed"
blog.save!
```

Look at the image, you can see that there are duplicate set of attributes after HashDiff comparison.
<img width="1345" alt="Screenshot at Oct 12 09-58-32" src="https://user-images.githubusercontent.com/11540218/195273145-0d3a82ab-0e94-4964-baa5-56d3c19c3987.png">

## Logs
```
D, [2022-10-12T10:05:40.261857 #41270] DEBUG -- : Pool shutdown finished
D, [2022-10-12T10:05:42.024169 #41270] DEBUG -- : [httplog] Connecting: excelify-karlis-2.myshopify.com:443
D, [2022-10-12T10:05:42.570946 #41270] DEBUG -- : [httplog] Sending: POST http://excelify-karlis-2.myshopify.com:443/admin/api/2022-04/blogs.json
D, [2022-10-12T10:05:42.571025 #41270] DEBUG -- : [httplog] Data: {"blog":{"handle":"my-blog","title":"My Blog"}}
D, [2022-10-12T10:05:42.571044 #41270] DEBUG -- : [httplog] Status: 201
D, [2022-10-12T10:05:42.571062 #41270] DEBUG -- : [httplog] Benchmark: 0.508879 seconds
D, [2022-10-12T10:05:42.571085 #41270] DEBUG -- : [httplog] Response:
[...]

PR #1149, opened 2023-04-28
Fix dirty attribute changes on update

### LINKED ISSUE #1133: Model save erroneously sends unchanged attributes to API
# Issue summary

There are a number of issues coming from the shopify gem erroneously sending the entire object when saving as outlined here almost 6 months ago: https://github.com/Shopify/shopify-api-ruby/issues/1037

The gem should only send the changed attributes to the API, but it sends all fields, leading to API errors.

Changing/Saving any part of a Customer created without an email (phone number only) through the checkout will return
`{"email_marketing_consent"=>["An email address is required when setting the email marketing consent."]}`

Changing/Saving any part of a Customer with a nil email_marketing_consent.opt_in_level will return
`{"email_marketing_consent"=>["Opt in level must exist"]}`

Changing/Saving any part of an Order (maybe from POS?) without a shipping address will return
`{"shipping_address"=>[": Enter a last name."]}`

- `shopify_api` version: 12.3.0
- Ruby version: 2.7.2
- Operating system: All

## Expected behavior

The gem should only send the changed attributes on save, as outlined in https://github.com/Shopify/shopify-api-ruby/issues/1037

## Actual behavior

The gem sends all attributes leading to api errors.

see  https://github.com/Shopify/shopify-api-ruby/issues/1037

nelsonwittwer:
## Description

Fixes #1133 and #1037 which both report a substantial issue with tracking what attributes were actually modified by the user since the rest resources were initialized.

The root cause of this issue, as pointed out by @kaarelss, was symbols vs hashes when comparing drift on attributes 🤦 . I also needed to 1) set associations in `original_state` since our API allows us to modify associated resources and 2) needed to ignore read only attributes in `original_state`.

## How has this been tested?

- Wrote a unit test that reproduced the issue. The fix was then applied to the base rest resource which passed the test.
- In a local environment using this branch of the ruby API library, I found an existing product and updated a single attribute. I compared all other attributes and verified only the single attribute was modified.

## Checklist:


wayne692 (commented):
It's easy

lib/shopify_api/rest/base.rb
```diff
@@ -234,16 +234,20 @@ def create_instance(data:, session:, instance: nil)
...
               end
               instance.public_send("#{attribute}=", attr_list)
             elsif has_one?(attr_sym) && value
               # force a hash if core returns values that instantiate objects like "USD"
               data_hash = value.is_a?(Hash) ? value : { attribute.to_s => value }
-              instance.public_send("#{attribute}=",
-                T.unsafe(@has_one[attr_sym]).create_instance(data: data_hash, session: session))
+              child = T.unsafe(@has_one[attr_sym]).create_instance(data: data_hash, session: session)
+              instance.public_send("#{attribute}=", child)
+              instance.original_state[attr_sym] = child.to_hash(true)
```
DuncanUszkay1, line 250: This was a breaking change for our application that went unreported in the changelog. We were creating an instance and immediately calling save on it, which with this change resulted in an empty array being sent to Shopify.

DuncanUszkay1: Not sure if we were using the library improperly since I didn't see mention of this is the docs, but if not I can go update the changelog for others who might run into this.

PR #1161, opened 2023-05-18
Update CLI env variable names

paulomarg:
## Description

With https://github.com/Shopify/cli/pull/1789, the CLI will deprecate some env var names. While they will still be present so previous apps won't break, we should migrate new apps as soon as possible to the new format.

This PR adds support for the new variable names, but falls back to the current ones so it won't affect existing apps.

## Checklist:


mkevinosullivan (approved):

paulomarg:
We've decided to close this for now since we might want to take a slightly different approach in the future!

PR #1188, opened 2023-07-27
Add config variable passed in to all methods/classes using Context

sterrym:
## Description

Fixes #<issue-number>

## How has this been tested?

## Checklist:


sterrym:
> Hi there 👋
> 
> Could you give us a bit more context on the use case for this 👇
> 
> > desire/need to create more than one Shopify App within a single Rails application.

Absolutely. This was initially an internal need that we figured other partners would be interested in as well. Specifically, we are building multiple payments apps within an existing Shopify rails app. In our specific case, we are actually building a single Rails engine that supports three different types of payments apps. Each of them needs to be configured in a different way.

Internally, we ended up going another direction and are just using internal Shopify tools but I still suspect there's a valid use-case here.

Issue #970, opened 2022-06-02
CustomerAddress.all doesn't return results


agramichael:
# Issue summary

`CustomerAddress.all` doesn't return results, using Rest API with a private app.

## Expected behavior

Returns list with customer's addresses.

## Actual behavior

Returns empty list for a customer with an address.

## Steps to reproduce the problem

1. Create a customer with an address on Shopify.
2. Initialise gem with private app config:
```ruby
ShopifyAPI::Context.setup(
  api_key: "",
  host_name:"",
  scope: "",
  session_storage: ShopifyAPI::Auth::FileSessionStorage.new,
  is_embedded: false,
  is_private: true,
  api_version: Rails.application.credentials.shopify.api_version!,
  api_secret_key: Rails.application.credentials.shopify.password!,
  private_shop: "#{Rails.application.credentials.shopify.shop_name!}.myshopify.com"
)
```
4. Perform `ShopifyAPI::CustomerAddress.all(customer_id: x)` and receive `[]`
5. Perform `ShopifyAPI::Customer.find(id: x).addresses` and confirm the customer actually has an address.

## Specifications

- `shopify_api` version: 10.1.0
- Shopify API version used: 2022-04

ilyabyar:
Seeng exactly the same issue on both 10.0.3 and 10.1.0

Trimakas:
And so what's the resolution here? I'm having this issue with `AssignedFulfillmentOrder`

mllemango:
https://github.com/Shopify/shopify-api-ruby/pull/1208 should address this

@Trimakas I'll look into implementing something similar for AssignedFulfillmentOrder`

PR #1192, opened 2023-08-08
Fix CustomerAddress.all returning empty array

### LINKED ISSUE #1018: ShopifyAPI::CustomerAddress.all always returns an empty array
# Issue summary

ShopifyAPI::CustomerAddress.all always returns an empty array.

## Expected behavior

It should return an array of CustomerAddress objects

## Actual behavior

This happens because create_instance_from_response method is unable to create instances if normalized resource class name is different than parent attribute in API response. It tries to find normalized class_name inside response body, but it could not find it because response from API contains parent attribute `addresses` instead of `customer_addresses`. So the resource class should be named `ShopifyAPI::Addresses` or API response should contain `customer_addresses` as parent attribute.  This is same issue just with different resource - https://github.com/Shopify/shopify-api-ruby/issues/992

So this results as no objects created and returned an empty array.

```
def create_instances_from_response(response:, session:)
  objects = []

  body = T.cast(response.body, T::Hash[String, T.untyped])

  if body.key?(class_name.pluralize) || (body.key?(class_name) && body[class_name].is_a?(Array))
    (body[class_name.pluralize] || body[class_name]).each do |entry|
      objects << create_instance(data: entry, session: session)
    end
  elsif body.key?(class_name)
    objects << create_instance(data: body[class_name], session: session)
  end

  objects
end
```

1. Create Customer with addresses
2. Find addresses for that customer with - `ShopifyAPI::CustomerAddress.all(customer_id: customer_id)`

## Specifications
- `shopify_api` version: 11.1.0
- Shopify API version used (example: `2022-04`):

@itissible/matrixify

cdmwebs:
## Description

Fixes #1018, #970

When fetching CustomerAddresses, the results are returned as `addresses`. This PR adds `json_response_body_name` to the class and updates tests.

## How has this been tested?

Assertions added to count the returned results.

## Checklist:

### DIFF
--- a/test/rest/2023_07/customer_address_test.rb
+++ b/test/rest/2023_07/customer_address_test.rb
@@ -46,12 +46,15 @@ class CustomerAddress202307Test < Test::Unit::TestCase
     )
 
     assert_requested(:get, "https://test-shop.myshopify.io/admin/api/2023-07/customers/207119551/addresses.json?limit=1")
+    assert_equal(1, response.count)
 
     response = response.first if response.respond_to?(:first)
 
     # Assert attributes are correctly typed preventing Sorbet errors downstream
     if response.respond_to?(:original_state)
       response&.original_state&.each do |key, value|
+        next if key == :default
+
         begin
           response.send(key)
         rescue TypeError => error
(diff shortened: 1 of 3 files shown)


test/rest/2023_07/customer_address_test.rb
```diff
@@ -46,12 +46,15 @@ def test_1()
...
 
     assert_requested(:get, "https://test-shop.myshopify.io/admin/api/2023-07/customers/207119551/addresses.json?limit=1")
+    assert_equal(1, response.count)
 
     response = response.first if response.respond_to?(:first)
 
     # Assert attributes are correctly typed preventing Sorbet errors downstream
     if response.respond_to?(:original_state)
       response&.original_state&.each do |key, value|
+        next if key == :default
```
cdmwebs, line 56: I don't like this, but I'm not sure how to better handle. `default` is both an attribute and a method.

cdmwebs:
Looking for comments here. If this makes sense, I'll update the other API versions as well.

Issue #1164, opened 2023-05-25
Updating assets is broken in v13.0.0


ClaytonPassmore:
# Issue summary

In v12.5.0, you could update the value of an asset by fetching it, setting a new value, and then saving it:

```ruby
asset = ShopifyAPI::Asset.all(asset: { key: "layout/theme.liquid" }, theme_id: theme.id).first
asset.value = asset.value + "<p>Something new</p>\n"
asset.save!
```

In v13.0.0, updating an asset results in a 406 Not Acceptable error.

I'm fairly confident this is broken because of the changes made in #1149. We're no longer sending up the required `key` parameter (because it hasn't changed).

With access to the asset API being removed in API version 2023-04, I'm not super hopeful that this is actually going to get fixed 🙈 

- `shopify_api` version: v13.0.0
- Ruby version: 3.2.2
- Operating system: Ubuntu 22.04 LTS

## Expected behavior

Updating an asset should behave similar to the way it did in v12.5.0 (in that it should work 😄). Including the `key` when we save the model would fix the problem.

v12.5.0 was obviously sending up more attributes than it should have, but it did work. See the screenshot below. Excuse the large code block - I'm replicating how the gem (incorrectly) determined which attributes changed in v12.5.0.

<img width="949" alt="Screenshot 2023-05-25 at 3 10 58 PM" src="https://github.com/Shopify/shopify-api-ruby/assets/6856391/72629344-fe88-4e11-ae16-c25acc0f127e">

## Actual behavior

In v13.0.0, updating an asset results in a 406 Not Acceptable error. Note that I'm using the new private methods introduced in #1149 for determining the changed attributes.

<img width="953" alt="Screenshot 2023-05-25 at 3 28 18 PM" src="https://github.com/Shopify/shopify-api-ruby/assets/6856391/9667c8f9-87f9-4030-a172-ac465a01d124">

## Steps to reproduce the problem

1. Fetch an asset
1. Change the asset's `value`
1. Try to save the asset

garethson:
https://github.com/Shopify/shopify-api-ruby/pull/1149 also appeared to break saving metafields. Example:
```ruby
metafield = ShopifyAPI::Metafield.all(limit: 1, namespace: "foo", key: "bar").first
metafield.value = "new value"
metafield.save
```
This returns `{"errors":{"metafield":"Required parameter missing or invalid"}` an error because neither `key` nor `namespace` are sent, as those fields did not change.

[...]

PR #1197, opened 2023-08-18
Patch bug with newly added attributes after resource creation

nelsonwittwer:
## Description
As mentioned in #1164, change tracking for associated records is prone to errors. Especially when modifying a `has_many` association. 

The root cause was an erroneous implementation of `HashDiff#left_diff` when removing or adding associated objects. When the key wasn't present in the original, which `HashDiff#left_diff` would return a `{0=> HashDiff::NO_VALUE}` as the value, we will now return whatever the current state of the object is since the original state didn't have that key.

## How has this been tested?
Unit test + recreating reports in console. 

## Checklist:

### DIFF
--- a/test/clients/base_rest_resource_test.rb
+++ b/test/clients/base_rest_resource_test.rb
@@ -203,7 +203,7 @@ module ShopifyAPITest
         assert_equal("some-value", resource.to_hash["unknown"])
       end
 
-      def test_saves_removing_children_zzz
+      def test_saves_removing_children
         draft_order_data = {
           "id" => 1124601987358,
           "line_items" => [{


nelsonwittwer:
@ mvanholstyn - I'm going to do more testing to verify this fix, but you would mind testing this solution with your workflow as well?

nelsonwittwer:
Looking at the reported issue, it looks like we need to also make sure we are sending the required IDs. I'll add that fix as well here before getting review

test/clients/base_rest_resource_test.rb
```diff
@@ -203,6 +203,32 @@ def test_loads_unknown_attribute
         assert_equal("some-value", resource.to_hash["unknown"])
       end
 
+      def test_saves_removing_children_zzz
```
zzooeeyy, line 206:
```suggestion
      def test_saves_removing_children
```

lib/shopify_api/rest/base.rb
```diff
@@ -365,10 +365,18 @@ def attributes_to_update
...
 
-        HashDiff::Comparison.new(
+        diff = HashDiff::Comparison.new(
           deep_stringify_keys(original_state_for_update),
           deep_stringify_keys(to_hash(true)),
         ).left_diff
+
+        diff.each do |attribute, value|
+          if value.is_a?(Hash) && value[0] == HashDiff::NO_VALUE
+            diff[attribute] = send(attribute)
```
zzooeeyy, line 375: What does `send(attribute)` do in this case? I'm not sure why it returns `[]` in this case?

nelsonwittwer: `send[attribute]` is how to invoke an instance method from the context of an instance. In this case, we are saying call your method that is named `attribute` in this loop. 

In the context of the test it is kinda like saying `diff["line_items"] = draft_order.line_items` but instead of having a variable named `draft_order` you have the current instance context.

nelsonwittwer: https://apidock.com/ruby/Object/send might do a better job than me 😅

zzooeeyy: THANK YOU

Issue #1202, opened 2023-08-23
Bug in error handling in Shopify::Rest::Base#save


martinsp:
# Issue summary

Since v13.0.0 [ShopifyAPI::DiscountCode#errors](https://github.com/Shopify/shopify-api-ruby/commit/2b5cb782e6d600f38a62520c52cf4b71eabd66ba#diff-f380d1c32c5bcaf9c16b9e4abe19f0aa04652404221286a328a5153e10c83e58) signature has been changed to `Hash`, and #errors is initialized as `Hash` for `DiscountCode` resources.

This causes exception in error handling in [ShopifyAPI::Rest::Base#save](https://github.com/Shopify/shopify-api-ruby/blob/v13.1.0/lib/shopify_api/rest/base.rb#L356) if `ShopifyAPI::Errors::HttpResponseError` is raised during save operation for `DiscountCode` resource,  because `Hash` does not implement `#errors`

- `shopify_api` version: 13.1.0
- Ruby version: 2.7.7

## Expected behavior

Rescuing from `ShopifyAPI::Errors::HttpResponseError` does not raise exception

## Actual behavior

#<NoMethodError: undefined method \`errors' for {}:Hash>` is raised

martinsp:
Please find the test case and test results below

<details>
   <summary>test_case.rb, run with: ruby test_case.rb</summary>

   ```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "minitest"
  gem "shopify_api", "~> 13.1.0"
  gem "webmock", "3.14"
end

require "minitest/autorun"
require "webmock/minitest"

require "shopify_api"

class DiscountCode202204Test < Minitest::Test
  def setup
    ShopifyAPI::Context.setup(
      api_key: "API_KEY",
      api_secret_key: "API_SECRET_KEY",
      api_version: "2023-07",
      host: "https://app-address.com",
      scope: ["scope1", "scope2"],
      is_private: false,
      is_embedded: false,
      logger: ::Logger.new(T.let(StringIO.new, StringIO)), # comment line to see logging on stdout
      user_agent_prefix: nil,
      old_api_secret_key: nil,
      log_level: :off,
    )

    test_session = ShopifyAPI::Auth::Session.new(id: "id", shop: "test-shop.myshopify.io", access_token: "this_is_a_test_token")
    ShopifyAPI::Context.activate_session(test_session)
  end

  def teardown
[...]

PR #1205, opened 2023-08-28
Fixes invalid typing of discount code errors

garethson:
## Description

Fixes [#1202](https://github.com/Shopify/shopify-api-ruby/issues/1202)

It appears as though when https://github.com/Shopify/shopify-api-ruby/pull/1153 was shipped, it set the `errors` type the `ShopifyAPI::DiscountCode` class to be `T.nilable(T::Hash[T.untyped, T.untyped]))` when it should have just been left as the type [defined in ShopifyAPI::Rest::Base](https://github.com/Shopify/shopify-api-ruby/blob/95a1847398e97de5c02af2a9ebd55423342e5c7a/lib/shopify_api/rest/base.rb#L45).

As reported by @martinsp, this results in an error when `ShopifyAPI::Rest::Base` attempts to `@errors.errors << e`.

It's unclear to why this was added in https://github.com/Shopify/shopify-api-ruby/pull/1153 but I suspect it was just an unintended regression.

## How has this been tested?

Tested manually and also added a test here.

## Checklist:


test/rest/2023_07/discount_code_test.rb
```diff
@@ -324,4 +324,26 @@ def test_9()
     end
   end
 
+  sig do
```
garethson, line 327: I only put this one tests in the latest API version, I doubt this is where this test belongs as it's supposed to be auto-generated. Seems to me that the auto-generation of these tests are missing some error cases.

nelsonwittwer:
Thanks for your contribution! Unfortunately, these REST resource files are auto generated upstream so we won't be able to merge this in as the next time we generate these resources your fixes will be reverted. Having said that, we can apply your fix upstream! We are working on a few other fixes in this area and will add these to our list!

garethson:
@nelsonwittwer Thanks! I'll watch for updates on this issue, appreciate you pulling this in. 👍

garethson:
@nelsonwittwer  This has been rebased and everything is passing, is it possible to get this looked at? 

Thanks!

PR #1212, opened 2023-09-11
Improve V9->v10 Upgrade Docs

zzooeeyy:
## Description
Add more examples and reference for upgrading this gem from v9.

### New
[New dedicated page - "Breaking change notice for version 10.0.0"](https://github.com/Shopify/shopify-api-ruby/blob/421e1823fcda3d4cd34b966ddc96683795ebd2e1/BREAKING_CHANGES_FOR_V10.md#breaking-change-notice-for-version-1000)
[Link from README](https://github.com/Shopify/shopify-api-ruby/blob/421e1823fcda3d4cd34b966ddc96683795ebd2e1/README.md#breaking-change-notice-for-version-1000)

### Existing
[Existing docs on v10 upgrade](https://github.com/Shopify/shopify-api-ruby/blob/712c960d453aeb14e5422c7e80413917bb62a8bb/README.md#breaking-change-notice-for-version-1000)

### DIFF
--- a/BREAKING_CHANGES_FOR_V10.md
+++ b/BREAKING_CHANGES_FOR_V10.md
@@ -2,6 +2,18 @@
 
 We've rewritten this library for v10, so that it provides all essential features for a Shopify app without depending on the [Active Resource](https://github.com/rails/activeresource) or [graphql-client](https://github.com/github/graphql-client) libraries.
 
+#### Table of Contents
+- [New Features](#new-features)
+- [Upgrade Guide](#upgrade-guide)
+  - [1. Configuration Changes](#1-configuration-changes)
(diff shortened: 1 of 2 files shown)


nelsonwittwer (approved):
Love it! Thanks for taking the time to put this together :) 

I wonder if we should include some background on WHY we needed the big rewrite. This was before my time but from what I've been able to gather it largely had to do with 1) removing 3rd party cookies and 2) so we could consolidate code bases for REST resources with other languages. @paulomarg do you have any more history you could share with the class here 🙏

nelsonwittwer:
Thanks @paulomarg ! @zzooeeyy - would you mind mixing in that context into your upgrade doc so the community understands why we did what we did? Thank you both!

PR #1236, opened 2023-10-26
fix intermittent test failures

nelsonwittwer:
Fixes intermittent test failures [like this one](https://github.com/Shopify/shopify-api-ruby/actions/runs/6654959771/job/18084102541) by not hard coding versions anymore into stubs.


zzooeeyy (approved):
+10000 points for the GIF

nelsonwittwer:
After removing `unstable` contexts from test was able to run ~25 pipelines locally without error. I think that should do it! 🤞

Issue #1235, opened 2023-10-24
Add the ability to modify base URI for 1st party shopify apps


nelsonwittwer:
For 1st party Shopify apps that use the API library, adding the ability to use the internal domain vs public domain can reduce latency by 80ms by avoiding the public internet for API calls.

If we modified the [base_uri](https://github.com/Shopify/shopify-api-ruby/blob/main/lib/shopify_api/clients/http_client.rb#L16) in our http client, we'd need to make sure the session's shop worked well with unified admin domains and legacy patterns. We'd need to validate this change by testing with 3rd party and 1st party applications before releasing.

PR #1239, opened 2023-11-06
Use ShopifyAPI::Clients::HttpClient instead of HTTParty directly in OAuth.validate_auth_callback

brendo:
RE: #1235

## Description

This PR replaces a direct call to HTTParty with the HttpClient included by this library. Whilst this doesn't change much from a consumer's perspective, internally it means a monkey patch on HttpClient to use the internal domain covers all HTTP requests originating from this library.

## How has this been tested?

At the moment, just the unit tests in this project.

## Checklist:


nelsonwittwer (approved):
Thanks for this little refactor! Like the consistency / ability to patch this with 1st party apps.

PR #1244, opened 2023-11-23
Add expired? method to check if session is expired

gbzodek:
### What is the problem it is solving?

There isn't currently a method available to check if an auth session is expired (for online tokens mainly). It's up to each app to implement the method if they want to do such check. This PR adds an `expired?` method to the `Session` that will check the `expires` attribute if it exists.

### What is the context of these changes?

We are planning to add a check on the expiry date in `shopify_app` to trigger an early re-auth, and having that method will make the check cleaner

### What is the impact of this PR?

A new method can be used to check if the session is expired. No impact as long as it's not used

## How has this been tested?

Unit tests provided in the PR

## Checklist:

- [X] My commit message follow the pattern described in [here](https://chris.beams.io/posts/git-commit/)
- [X] I have added tests that prove my fix is effective or that my feature works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Note: For changes to the API, see https://shopify.dev/changelog?filter=api
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


paulomarg (approved):
This is great, thanks!

PR #1245, opened 2023-11-23
Fixes/type errors

hrdwdmrbl:
Since you completely ignored [my last PR](https://github.com/Shopify/shopify-api-ruby/pull/1072) which contained one of these fixes, I assume that you will also ignore this one. But alas I submit this PR anyway. That one even had full tests and everything! In this one I didn't bother updating the tests with the latest fixes because I don't trust you to even look at this so why bother? Why bother having code of conduct and PR templates if no one even reviews the PRs? 🙄

Anyway, your types are wrong, please update them. Your own backend does not always return the types that you use here. You can pretend like it does, but it does not. 

Shopify:  🙈🙉🙊
Me: 🤯

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Note: For changes to the API, see https://shopify.dev/changelog?filter=api
 
 ## Unreleased
 
+- [#1071](https://github.com/Shopify/shopify-api-ruby/issues/1071) Fix FulfillmentEvent class types
+- Fix: InventoryItem class `harmonized_system_code` attribute type which can be either integer, string or nil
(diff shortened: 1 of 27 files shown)


lizkenyon:
Hi there 👋

Sorry for the sluggish responses on this. Tldr; we do want to get this merged. And we do appreciate you taking the time to find all these type errors.

We have a ton of upstream friction with the REST resources, and we would ideally make these changes upstream, like we outline in our contribution guidelines.

Though we are planning on merging this when the January API release goes out, and apply these changes.

hrdwdmrbl:
@lizkenyon I already did sign the CLA. Just rebased and made the same changes to 2024-01

paulomarg:
Thanks for your contribution!

PR #1268, opened 2024-01-19
Add webhook _id and api_version to webhook handler

### LINKED ISSUE #1247: Include X-Shopify-Webhook-Id to webhook handlers
## Overview/summary

As per [Shopify's best practices for handling Webhooks](https://github.com/Shopify/shopify-api-ruby/blob/459b04e4cc58942866c0b53a5ece0bac73d90167/lib/shopify_api/webhooks/request.rb#L1-L56), we should be checking the webhook id for duplicates when processing. For example, we could use it to ensure only a single background job is enqueued for a specific webhook id at any one time.

Unfortunately the ShopifyAPI does not include the `X-Shopify-Webhook-Id` to our webhook handlers, and so we have no way of following these best practices when using the gem.

https://github.com/Shopify/shopify-api-ruby/blob/459b04e4cc58942866c0b53a5ece0bac73d90167/lib/shopify_api/webhooks/request.rb#L1-L56

Would a PR be accepted for this?

lizkenyon:
## Description

Fixes #1247 

Shopify recommends as a [best practice](https://shopify.dev/docs/apps/webhooks/best-practices#ignore-duplicates) to check the `x-shopify-webhook-id` to verify you are not processing duplicate webhooks. We currently are not passing this information into the webhook handler. This PR will add the webhook_id and api_version to the webhook handler to bring the ruby library in line with the Javascript library in terms of information the handler has access to.

This could be resolved in a couple of approaches outlined below

### Add `webhook_id` and `api_version` to handle function
This is what is currently implemented in this PR.
The `webhook_id` and `api_version` are added to the webhook request object from the headers. They are then explicitly passed to the handle function.

This would be a breaking change, but would be a straightforward fix. Developers would need to add `webhook_id` and `api_version` to their `handle` implementation. 

### Add metadata
If we are concerned, that in the future we may need to add even more information to the handle function and would require more breaking changes we could instead change the method signature to include the body and metadata. Where metadata could be a catch all of all the additional information. This would make any future additions easier but would require slight more initial refactoring for developers on the initial change.

```ruby
      sig do
        abstract.params(body: T::Hash[String, T.untyped], metadata: T::Hash[String, T.untyped]).void
      end
      def handle(body:, metadata); end
    end
```

### Add new handle function

If we are concerned about introducing a breaking change, we could introduce a new handle function. (eg. `handle_with_webhook_and_version` then developers would be able to implement the new function if they wanted access to the `webhook_id` and `api_version`. This could get complicated if we need to add more information in the future. (We would need to add another new function?)

## How has this been tested?

## Checklist:

### DIFF
--- a/docs/usage/webhooks.md
+++ b/docs/usage/webhooks.md
@@ -7,17 +7,14 @@ If using in the Rails framework, we highly recommend you use the [shopify_app](h
 
 ## Create a Webhook Handler
 
-### New webhook handler as of Version 13.5.0
-If you want to register for an http webhook you need to implement a webhook handler which the `shopify_api` gem can use to determine how to process your webhook. You can make multiple implementations (one per topic) or you can make one implementation capable of handling all the topics you want to subscribe to. To do this simply make a module or class that includes or extends `ShopifyAPI::Webhooks::WebhookHandler` and implement the `handle_webhook` method which accepts the following named parameters: data: `Hash[Symbol, untyped]`. An example implementation is shown below:
-
-You will also have to define a method called `use_handle_webhook?` which returns a boolean. If this method returns `true` then the `handle_webhook` method will be called when a webhook is received. If this method returns `false` then the `handle` method will be called when a webhook is received. The `handle` method is the old way of handling webhooks and will be deprecated in a future version of the gem. When using this method you will not have access to the `webhook_id` or `api_version` of the webhook.
+If you want to register for an http webhook you need to implement a webhook handler which the `shopify_api` gem can use to determine how to process your webhook. You can make multiple implementations (one per topic) or you can make one implementation capable of handling all the topics you want to subscribe to. To do this simply make a module or class that includes or extends `ShopifyAPI::Webhooks::WebhookHandler` and implement the `handle` method which accepts the following named parameters: data: `WebhookMetadata`. An example implementation is shown below:
 
 `data` will have the following keys
-- `:topic` - The topic of the webhook
-- `:shop` - The shop domain of the webhook
-- `:body` - The body of the webhook
-- `:webhook_id` - The id of the webhook event to [avoid duplicates](https://shopify.dev/docs/apps/webhooks/best-practices#ignore-duplicates)
-- `:api_version` - The api version of the webhook
+- `topic`, `String` - The topic of the webhook
+- `shop`, `String` - The shop domain of the webhook
+- `body`, `T::Hash[String, T.untyped]`- The body of the webhook
+- `webhook_id`, `String` - The id of the webhook event to [avoid duplicates](https://shopify.dev/docs/apps/webhooks/best-practices#ignore-duplicates)
+- `api_version`, `String` - The api version of the webhook
 
 ```ruby
 module WebhookHandler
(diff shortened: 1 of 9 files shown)


paulomarg:
I think we can make this a minor change by adding a static value to the webhook handler - if we add

```rb
module Handler
  FLAG_TO_USE_ALL_FIELDS = T.let(false, T::Boolean)
```

and then do

```rb
module TestHelpers
  class FakeWebhookHandler
    include ShopifyAPI::Webhooks::Handler
    FLAG_TO_USE_ALL_FIELDS = true
```

we can then use `if handler.class::USE_ALL_FIELDS` in the registry to call the appropriate version of the function.

We can add deprecation logs when using the old format to give folks time to fix it before a major version. We'd need to declare the abstract class in a way that could deal with both cases, but I think that should be possible with some creative typing?

Do you think that would work?

I agree that there is a bit of a deeper problem here. I see two ways that apps can make this future-proof:
- declare their handlers with a `**metadata` argument
- change the function to take in a PORO

I'd lean towards the second option because the first one is somewhat easy to forget and can trip people up - the second option is more reliable IMO.

nelsonwittwer (commented):
I'd only introduce a new breaking change if we have plans for major version on the horizon. If not, I'd prefer to allow users to opt-in to this option in the initializer. 

Besides that roll out question, I think everything looks great!

lib/shopify_api/webhooks/handler.rb
```diff
@@ -4,12 +4,29 @@
...
     module Handler
+      include Kernel
       extend T::Sig
       extend T::Helpers
-      interface!
+      abstract!
 
-      sig { abstract.params(topic: String, shop: String, body: T::Hash[String, T.untyped]).void }
+      sig do
+        params(data: T::Hash[Symbol, T.untyped]).void
```
paulomarg, line 13: Should we:
- explicitly type `data`?
- make `data` an object instead of a hash to keep the API similar to what it was before? Can we use `ShopifyAPI::Webhooks::Request`?

lib/shopify_api/webhooks/handler.rb
```diff
@@ -4,12 +4,29 @@
...
+        end
+      end
+
+      sig do
+        abstract.params(topic: String, shop: String, body: T::Hash[String, T.untyped]).void
+      end
       def handle(topic:, shop:, body:); end
+
+      sig { returns(T::Boolean) }
+      def use_handle_webhook?
```
paulomarg, line 27: I wonder if this name is clear enough - is there something we can name it that indicates what it does (i.e. that it takes in the metadata object, or the full metadata)?

We could even take it a bit further: if we need a new method, we could simply test for the presence of that method in the class when we're calling, so `use_handle_webhook?` wouldn't be necessary.

lib/shopify_api/webhooks/registry.rb
```diff
@@ -193,7 +193,16 @@ def process(request)
...
 
-          handler.handle(topic: request.topic, shop: request.shop, body: request.parsed_body)
+          if handler.use_handle_webhook?
+            handler.handle_webhook({ topic: request.topic, shop: request.shop, body: request.parsed_body,
+              webhook_id: request.webhook_id, api_version: request.api_version, })
+          else
+            handler.handle(topic: request.topic, shop: request.shop, body: request.parsed_body)
+            ShopifyAPI::Logger.warn(<<~WARNING)
+              DEPRECATED: Use ShopifyAPI::Webhooks::Handler#handle_webhook
+              instead of ShopifyAPI::Webhooks::Handler#handle
```
paulomarg, line 203: Can we add a link to the great docs you added here to help folks understand why we did it?

test/webhooks/registry_test.rb
```diff
@@ -359,6 +384,77 @@ def do_registration_test(delivery_method, path, fields: nil, metafield_namespace
         assert_equal(queries[delivery_method][:register_update_response], update_registration_response.body)
       end
 
+      def do_registration_test_new_handler(delivery_method, path, fields: nil, metafield_namespaces: nil)
```
paulomarg, line 387: Is this a test, or a helper function? I don't think it's called anywhere?

docs/usage/webhooks.md
```diff
@@ -7,10 +7,42 @@ If using in the Rails framework, we highly recommend you use the [shopify_app](h
 
 ## Create a Webhook Handler
 
+### New webhook handler as of Version 13.5.0
```
paulomarg, line 10: This is a bit of a nit, but I think we can just make this the "right" way of doing it in this doc, and then just add a quick note with an example on how it used to be. WDYT?

paulomarg (commented):
This looks great! I had some thoughts mostly around naming things, but the logic makes sense to me.

lib/shopify_api/webhooks/handler.rb
paulomarg: Hmm right - I guess we can't have an interface with an optional method, which complicates things indeed. Let's go ahead with this approach then.

lizkenyon: As discussed, I refactored the handler to have multiple handler modules. 
Developers can implement the new handler if they want they additional data.

nelsonwittwer (approved):
I like this method of rolling this out, well done!

lib/shopify_api/webhooks/registry.rb
```diff
@@ -193,7 +193,17 @@ def process(request)
             raise Errors::NoWebhookHandler, "No webhook handler found for topic: #{request.topic}."
           end
 
-          handler.handle(topic: request.topic, shop: request.shop, body: request.parsed_body)
+          if handler.is_a?(WebhookHandler)
+            handler.handle(data: WebhookMetadata.new(topic: request.topic, shop: request.shop,
+              body: request.parsed_body, api_version: request.api_version, webhook_id: request.webhook_id))
+          else
+            handler.handle(topic: request.topic, shop: request.shop, body: request.parsed_body)
+            ShopifyAPI::Logger.warn(<<~WARNING)
```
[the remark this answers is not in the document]
lizkenyon: Added! Can you verify what I did is correct

lib/shopify_api/webhooks/registry.rb
```diff
@@ -193,7 +193,18 @@ def process(request)
...
+          if handler.is_a?(WebhookHandler)
+            handler.handle(data: WebhookMetadata.new(topic: request.topic, shop: request.shop,
+              body: request.parsed_body, api_version: request.api_version, webhook_id: request.webhook_id))
+          else
+            handler.handle(topic: request.topic, shop: request.shop, body: request.parsed_body)
+            ShopifyAPI::Logger.deprecated(
+              "DEPRECATED: Use ShopifyAPI::Webhooks::WebhookHandler#handle
+              instead of ShopifyAPI::Webhooks::Handler#handle.
+              https://github.com/Shopify/shopify-api-ruby/blob/main/docs/usage/webhooks.md#create-a-webhook-handler
+            WARNING", "14.0.0"
```
paulomarg, line 205: Looks good, but is this `WARNING` bit supposed to be here?

lizkenyon:
Tested this and verified that: 
1) The deprecation warning is shown when you use the old handler
2) When using the new handler you have access to the additional fields.

image_1706209679580_0
image_1706213906411_0

PR #1289, opened 2024-03-12
Update webhooks documentation

jsaubry:
## Description

I updated my repositories to use the new webhook handler, but the documentation was slightly wrong.

- The `data` param is passed by name not position ([source](https://github.com/Shopify/shopify-api-ruby/blob/main/lib/shopify_api/webhooks/handler.rb#L35)).
- Use the new module rather than the old one ([deprecation](https://github.com/Shopify/shopify-api-ruby/blob/main/lib/shopify_api/webhooks/registry.rb#L202-L203)).

## Checklist:

### DIFF
--- a/docs/usage/webhooks.md
+++ b/docs/usage/webhooks.md
@@ -18,10 +18,10 @@ If you want to register for an http webhook you need to implement a webhook hand
 
 ```ruby
 module WebhookHandler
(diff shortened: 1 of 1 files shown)


paulomarg (approved):
Great catch, thank you!

PR #1309, opened 2024-04-11
Add `Session#copy_attributes_from` method

rachel-carvalho:
## Description

Adding this method to support changes in https://github.com/Shopify/shopify_app/pull/1822

## How has this been tested?

Added a unit test.

## Checklist:

### DIFF
--- a/lib/shopify_api/auth/session.rb
+++ b/lib/shopify_api/auth/session.rb
@@ -121,6 +121,17 @@ module ShopifyAPI
         end
       end
 
+      sig { params(other: Session).returns(Session) }
+      def copy_attributes_from(other)
+        JSON.parse(other.serialize).keys.each do |key|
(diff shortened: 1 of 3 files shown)


lib/shopify_api/auth/session.rb
```diff
@@ -121,6 +121,19 @@ def deserialize(str)
         end
       end
 
+      sig { params(other: Session).returns(Session) }
+      def copy_attributes_from(other)
```
paulomarg, line 125: Is this method equivalent to `serialize` + `deserialize`?

rachel-carvalho: I considered using serialize/deserialize but `deserialize` is a class method that creates a new object, and we need to mutate the existing one.

paulomarg (approved):

PR #1312, opened 2024-04-16
Use same leeway for `exp` and `nbf` when parsing JWT

rachel-carvalho:
## Description

Right now we specify a leeway of 10 seconds when validating JWT payload expiration, but no leeway for `nbf` (not valid before). The `shopify-api` JS [library](https://github.com/Shopify/shopify-app-js/blob/0fd6a80c0af45138240b6b508210d3ef6b9292ee/packages/apps/shopify-api/lib/session/decode-session-token.ts#L25C11-L25C25) does allow for `clockTolerance` in both fields, and I think it makes sense for the Ruby gem to do the same.

## How has this been tested?

I added unit tests.

## Checklist:

### DIFF
--- a/lib/shopify_api/auth/jwt_payload.rb
+++ b/lib/shopify_api/auth/jwt_payload.rb
@@ -73,8 +73,7 @@ module ShopifyAPI
 
       sig { params(token: String, api_secret_key: String).returns(T::Hash[String, T.untyped]) }
       def decode_token(token, api_secret_key)
-        JWT.decode(token, api_secret_key, true,
-          { exp_leeway: JWT_EXPIRATION_LEEWAY, algorithm: "HS256" })[0]
+        JWT.decode(token, api_secret_key, true, leeway: JWT_EXPIRATION_LEEWAY, algorithm: "HS256")[0]
       rescue JWT::DecodeError => err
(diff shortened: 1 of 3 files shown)


lib/shopify_api/auth/jwt_payload.rb
```diff
@@ -73,8 +73,7 @@ def ==(other)
 
       sig { params(token: String, api_secret_key: String).returns(T::Hash[String, T.untyped]) }
       def decode_token(token, api_secret_key)
-        JWT.decode(token, api_secret_key, true,
-          { exp_leeway: JWT_EXPIRATION_LEEWAY, algorithm: "HS256" })[0]
+        JWT.decode(token, api_secret_key, true, leeway: JWT_EXPIRATION_LEEWAY, algorithm: "HS256")[0]
```
rachel-carvalho, line 76: I would prefer to rename the constant to `JWT_LEEWAY` since it doesn't only apply to expiration anymore, but it's a public constant and might make this a breaking change, so I decided to keep it.

rachel-carvalho: Oh yeah good call, I'll do that then 👍

PR #1343, opened 2024-10-04
Add default value to Context::scope parameter to make it optional

zzooeeyy:
## Description
Fixes:
- https://github.com/Shopify/shopify_app/issues/1912
#### Background
The `scope` parameter was required as a part of the legacy OAuth flow (Authorization code grant). It's a required parameter during the `ShopifyAPI::Context` setup process. It can be set to `[]` or `""`. 

We don't have to enforce `scope` in the setup with Shopify managed install since scopes are configured in the `shopify.app.*.toml` files already. 

missing-keyword

#### Solution
Making this field to have a default value so it can be omitted during setup.

## How has this been tested?
- Linked a local version of `shopify-api-ruby` with my changes to my app
- Removed `scope` setup in the `ShopifyAPI::Context.setup` 
- Start server an observe no errors

#### Configuration without `scopes`
configuration

## Checklist:


zzooeeyy:
> What is the functionality when an app is not using managed install, and they omit scopes? Is OAuth successful, and you just get the default store properties scope granted?

Yup if they omit scopes, `shopify-api-ruby` will default their "configuration" to be an empty list and request the auth code grant flow for installation with an empty list of scopes, and this will prompt the shop for the default "store data"

Here's an example request:
```
https://admin.shopify.com/store/mystore/oauth/authorize?client_id=this-is-a-valid-client_id&scope=&redirect_uri=https%3A%2F%2Fsomething-something-something-missions.trycloudflare.com%2Fapi%2Fauth%2Fcallback&state=P5F5J74FDCYREpy&grant_options%5B%5D=
```
image

lizkenyon (approved):
Thanks @zzooeeyy ⭐

PR #1344, opened 2024-10-30
Fix: ShopifyAPI::Webhooks::Registry doesn't update webhooks if metafield_namespaces has changed

### LINKED ISSUE #1323: ShopifyAPI::Webhooks::Registry doesn't update webhooks if metafield_namespaces has changed
## Overview/summary

`ShopifyAPI::Webhooks::Registry#webhook_registration_needed?` only checks that the `address` of the webhook is the same.
If anything else changed (metafield_namespaces, fields, private_metafield_namespaces, format) the webhook will be considered up to date and will be skipped.

The (Graphql and Rest) API supports `update` operations on Webhooks, so it should be trivial for the `ShopifyAPI::Webhooks::Registry` to implement updates for existing webhooks.

Basically the behaviour should be as follow : 
- No webhook existing for that address + topic ? -> create it
- If webhook exist, check if other parameters (fields, metafields, ...) are up to date. If not -> update it
- Otherwise all is good, we don't have to do anything

andy-liuu:
TODO: 
- [x] missing edgecase: arrays have same contents but different orders
- [x] missing edgecase: nil and empty array check should be good
- [x] make solution more ruby-like?
- [x] try to reduce duplicate code?
- [x] add more unit tests to not update when nothing changes
- [x] tophat
- [x] changelog

## Description

Fixes #1323 

### What is the problem it is solving?

- Previously, the ruby api would only update webhooks if the topic already existed but the callback url / endpoint was different. It did not consider the scenario where the same topic and callback url were supplied, but the fields or metafield namespaces were updated. I chose to only check fields and metafield namespaces because those are the only attributes accepted by the ruby api (despite the actual gql object having other (deprecated) fields.

### What is the context of these changes?
- refactor the existing unit tests so that the shared functions are more flexibly called via adding new parameters
    - Now, each test defines it's own expected queries, responses and values, instead of having an if statement in the helper function to deal with it
- remove unused code in the unit testing file
- write tests and add to existing ones for ...
    - having an update mutation called when just path, fields, or metafieldNamespaces is changed
    - adding a new webhook with each of these params
    - not updating when an existing identical webhook exists
    - & hardcode expected queries in the testing helper file
- update the check for existing webhook query for each of http, pubsub, and eventbridge
- update the parsing method for each of http, pubsub, eventbridge
    - pulling shared code into registration superclass. allowing subclass to implement a specific detail
- update the logic in webhook_registration_needed to check for diffs in fields and metafieldNamespaces

### What is the impact of this PR?

Fixing #1323 

## How has this been tested?

- Made new unit tests and strengthened existing tests to account for the new scenarios
    - the previous tests were running the exact same "update path" test regardless of parameters in do_register_test. 

## Checklist:

### DIFF
--- a/test/webhooks/registry_test.rb
+++ b/test/webhooks/registry_test.rb
@@ -40,7 +40,7 @@ module ShopifyAPITest
 
       VALID_PROTOCOL_ADDRESSES.each do |protocol, addresses|
         addresses.each do |address|
-          define_method("test_#{protocol}_no_registration_if_identical_webhook_exists_on_#{address}") do
+          define_method("test_#{protocol}_no_registration_if_identical_webhook_exists_with_address_#{address}") do
             # Given
             setup_queries_and_responses(
               [queries[protocol][:check_query]],


test/webhooks/registry_test.rb
andy-liuu, line 1: In ruby, is it better to have code in unit tests be very copy-paste like (like right now)? Or better to introduce logic to make it more elegant e.g. looping thru each of http, pubsub, eventbridge?

Pros:
- easier to tell if a specific scenario fails
- less mental overhead on tests (less likely that test logic is buggy

Cons:
- More code to change if these tests become outdated
- looks dumb

lib/shopify_api/webhooks/registration.rb
```diff
@@ -48,8 +48,28 @@ def mutation_name(webhook_id); end
...
-      def parse_check_result(body); end
+      sig do
+        params(body: T::Hash[String, T.untyped]).returns({
+          webhook_id: T.nilable(String),
+          current_address: T.nilable(String),
+          fields: T::Array[String],
+          metafield_namespaces: T::Array[String],
+        })
+      end
+      def parse_check_result(body)
```
andy-liuu, line 59: Is this the "ruby way" to remove duplicate code i.e. with a superclass & subclass implementation. Is there a good way to ensure this never gets called by itself?

PR #1356, opened 2025-01-03
Autoload resource classes

gmcgibbon:
## Description

#### What is the problem it is solving?

Implements autoloading on Active Resources classes so they don't load early unnecessarily. This improves load times for Rails applications, especially in instances where they are not used.

#### What is the context of these changes?

I don't want this gem to load Active Resource early, it causes unnecessary speed regressions in my application's boot time.

#### What is the impact of this PR?

Shaves about 40ms off boot time. The remainder is caused by other gems we're requiring like GraphQL, but it implements autoloading in the latest version. The time savings would be much greater in apps due to the amount of load hooks for Active Resource in place.

## How has this been tested?

Test this with:
```rb
#!/usr/bin/env ruby
# frozen_string_literal: true

require 'bundler/setup'
require "benchmark"
# require "ostruct"

time = 1000 * Benchmark.realtime do
  require 'shopify_api'
end

puts "Took #{time}ms"
```

Note, you might have to add racc to the gemfile  and require ostruct before the gem require if you're running on Ruby 3.3+.

#### Please, describe the tests that you ran to verify your changes.

I verified eager loading does in fact load the autoloaded constants. I didn't add anything to the test suite because this is a very old gem version that's on its way out, so we probably don't need continued assurance that autoloading works.

## Checklist:

cc @sle-c


sle-c (approved):
CI builds failed on Ruby v2x which we don't support anymore. I've confirmed this change works with the current test suite in Shopify core.

PR #1369, opened 2025-03-28
Make sub and sid jwt claims optional

BaggioGiacomo:
## Description

The generated session token on the checkout ui extensions doesn't contain:
- The `sid` claim
- The `sub` claim if the customer is not logged in

Having them mandatory causes an `Expected type String, got type NilClass` error

## Checklist:


BaggioGiacomo:
> One question I have is how are you (or how are you planning to) currently validating your session token in your app? ie. Are you using a controller concern from shopify_app?

Yes, I've an `AuthenticatedController` that includes `ShopifyApp::EnsureHasSession` concern. `new_embedded_auth_strategy` is set to true, token exchange is used

PR #1370, opened 2025-03-28
Support internal Shopify hosts

Shinomix:
## Description
Fixes https://github.com/Shopify/shop-server/issues/101259

In this PR, we add support for new Shopify internal hosts by rewriting the host to a generic one and passing the storefront domain as a `X-Forwarded-Host`. This change enables the installation of Shopify Apps, internally, in development.

We only filter on Shopify internal local development hosts so impact should be null in any other situation.

## How has this been tested?
This was successfully tested. Please check the testing instructions in the linked issue.

## Checklist:

### DIFF
--- a/lib/shopify_api/auth/oauth.rb
+++ b/lib/shopify_api/auth/oauth.rb
@@ -46,8 +46,8 @@ module ShopifyAPI
           }
 
           query_string = URI.encode_www_form(query)
+          auth_route = auth_base_uri(shop) + "/oauth/authorize?#{query_string}"
 
-          auth_route = "https://#{shop}/admin/oauth/authorize?#{query_string}"
           { auth_route: auth_route, cookie: cookie }
         end
 
@@ -105,6 +105,20 @@ module ShopifyAPI
           end
 
           { session: session, cookie: cookie }
+        end
+
+        private
+
+        sig { params(shop: String).returns(String) }
+        def auth_base_uri(shop)
+          return "https://#{shop}/admin" unless defined?(DevServer)
+
+          # For first-party apps in development only, we leverage DevServer to build the admin base URI
+          admin_web = T.unsafe(Object.const_get("DevServer")).new("web") # rubocop:disable Sorbet/ConstantsFromStrings
+          admin_host = admin_web.host!(nonstandard_host_prefix: "admin")
+          shop_name = shop.split(".").first
+
+          "https://#{admin_host}/store/#{shop_name}"
         end
       end
     end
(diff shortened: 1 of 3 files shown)


lib/shopify_api/auth/oauth.rb
```diff
@@ -47,7 +47,7 @@ def begin_auth(shop:, redirect_path:, is_online: true, scope_override: nil)
 
           query_string = URI.encode_www_form(query)
 
-          auth_route = "https://#{shop}/admin/oauth/authorize?#{query_string}"
+          auth_route = "https://#{shop.sub(".my.shop.dev", ".dev-api.shop.dev")}/admin/oauth/authorize?#{query_string}"
```
GoodForOneFare, line 50: I'll still work towards making this unnecessary.

GoodForOneFare (approved):
👍 One less paper cut for so many developers.  Thank you!

lib/shopify_api/auth/oauth.rb
GoodForOneFare: Sorbet makes it clumsier than it should be, but here's an alternative that uses standard endpoints:

```rb
          auth_route = if defined?(DevServer)
            admin_web = T.unsafe(Object.const_get("DevServer")).new("web") # rubocop:disable Sorbet/ConstantsFromStrings
            admin_host = admin_web.unsafe_unchecked_host(nonstandard_host_prefix: "admin")

            "https://#{admin_host}/store/#{shop.split(".").first}/oauth/authorize?#{query_string}"
          else
            "https://#{shop}/admin/oauth/authorize?#{query_string}"
          end
```

Shinomix: @GoodForOneFare Re-tested the installation flow with this new solution, it works perfectly.
Pushed a commit to use this solution and moved it to a dedicated method with a comment for clarity.

lib/shopify_api/auth/oauth.rb
```diff
@@ -106,6 +106,20 @@ def validate_auth_callback(cookies:, auth_query:)
...
         end
+
+        private
+
+        sig { params(shop: String).returns(String) }
+        def auth_base_uri(shop)
+          return "https://#{shop}/admin" unless defined?(DevServer)
+
+          # For first-party apps in development only, we leverage DevServer to build the admin base URI
+          admin_web = T.unsafe(Object.const_get("DevServer")).new("web") # rubocop:disable Sorbet/ConstantsFromStrings
```
gbzodek, line 117: this assumes that all 1P apps running locally want to target a local Core, which is not always the case as they can also target production Core (for example POS channel). This breaks the oauth flow for those apps

PR #1374, opened 2025-04-22
Fix edge cases for internal hosts

gbzodek:
Follow up to https://github.com/Shopify/shopify-api-ruby/pull/1370

## Description

This fixes 2 issues when targeting internal hosts

### Several requests with same HTTP client

The HTTP client can be reused for several HTTP calls. The current logic is flawed because it will update the headers for every request, ending with an invalid configuration:

1. first request is done to URI `https://shop1.my.shop.dev/admin/api/unstable/graphql.json`
 -> Host header is set to `app.shop.dev`and X-Forwarded-Host is set to `shop1.my.shop.dev`
2. second request is done, using the same client, and same URI
 -> X-Forwarded-Host is set to the value of the Host header, so `app.shop.dev` from previous request
 -> we end up with the Host and X-Forwarded-Host headers set to `app.shop.dev`, which fails

### Oauth with local DevServer and production host

Apps (like the POS channel or Flows afaik) can run locally but target production APIs, inside the production admin . The Oauth flow when installing them currently targets the internal Admin Web, whereas we should still target Production web when the shop does not have an internal host.

## How has this been tested?

This has been tested locally with POS channel

## Checklist:

- [X] My commit message follow the pattern described in [here](https://chris.beams.io/posts/git-commit/)
- [X] I have performed a self-review of my own code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 Note: For changes to the API, see https://shopify.dev/changelog?filter=api
 ## Unreleased
+- [#1374](https://github.com/Shopify/shopify-api-ruby/pull/1374) Fix edge cases for Shopify internal hosts
 
 ## 14.9.0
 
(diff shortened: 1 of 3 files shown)


GoodForOneFare (approved):

lizkenyon (approved):
👍

PR #1443, opened 2026-04-09
Add shop validation

gonzaloriestra:
## Description

https://docs.google.com/document/d/1c0iXhKBpm-yhvff0iSgq1XwBSPL39fTBVajIJ3svaxU/edit?usp=sharing 

## Checklist:

### DIFF
--- a/lib/shopify_api/auth/token_exchange.rb
+++ b/lib/shopify_api/auth/token_exchange.rb
@@ -21,12 +21,12 @@ module ShopifyAPI
 
         sig do
           params(
-            shop: String,
             session_token: String,
             requested_token_type: RequestedTokenType,
+            shop: T.nilable(String),
           ).returns(ShopifyAPI::Auth::Session)
         end
-        def exchange_token(shop:, session_token:, requested_token_type:)
+        def exchange_token(session_token:, requested_token_type:, shop: nil)
           unless ShopifyAPI::Context.setup?
             raise ShopifyAPI::Errors::ContextNotSetupError,
               "ShopifyAPI::Context not setup, please call ShopifyAPI::Context.setup"
@@ -36,10 +36,19 @@ module ShopifyAPI
           raise ShopifyAPI::Errors::UnsupportedOauthError,
             "Cannot perform OAuth Token Exchange for non embedded apps." unless ShopifyAPI::Context.embedded?
 
-          # Validate the session token content
-          ShopifyAPI::Auth::JwtPayload.new(session_token)
+          # Validate the session token and use the shop from the token's `dest` claim
+          jwt_payload = ShopifyAPI::Auth::JwtPayload.new(session_token)
+          dest_shop = jwt_payload.shop
 
-          shop_session = ShopifyAPI::Auth::Session.new(shop: shop)
+          if shop
+            ShopifyAPI::Logger.deprecated(
+              "The `shop` parameter for `exchange_token` is deprecated and will be removed in v17. " \
+                "The shop is now always taken from the session token's `dest` claim.",
+              "17.0.0",
+            )
+          end
+
+          shop_session = ShopifyAPI::Auth::Session.new(shop: dest_shop)
           body = {
             client_id: ShopifyAPI::Context.api_key,
             client_secret: ShopifyAPI::Context.api_secret_key,
@@ -74,7 +83,7 @@ module ShopifyAPI
           session_params = T.cast(response.body, T::Hash[String, T.untyped]).to_h
 
           Session.from(
-            shop: shop,
+            shop: dest_shop,
             access_token_response: Oauth::AccessTokenResponse.from_hash(session_params),
           )
         end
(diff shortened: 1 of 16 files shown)


lib/shopify_api/auth/token_exchange.rb
```diff
@@ -36,10 +36,11 @@ def exchange_token(shop:, session_token:, requested_token_type:)
...
             "Cannot perform OAuth Token Exchange for non embedded apps." unless ShopifyAPI::Context.embedded?
 
-          # Validate the session token content
-          ShopifyAPI::Auth::JwtPayload.new(session_token)
+          # Validate the session token and use the shop from the token's `dest` claim
+          jwt_payload = ShopifyAPI::Auth::JwtPayload.new(session_token)
+          dest_shop = jwt_payload.shop
 
-          shop_session = ShopifyAPI::Auth::Session.new(shop: shop)
+          shop_session = ShopifyAPI::Auth::Session.new(shop: dest_shop)
```
lizkenyon, line 51: If we are no longer using the shop param. We could deprecate it, so it can be removed in the next major version.

```
  jwt_payload = ShopifyAPI::Auth::JwtPayload.new(session_token)
  dest_shop = jwt_payload.shop

  if shop != dest_shop
    ShopifyAPI::Context.logger.warn(
      "shop parameter (#{shop}) does not match session token dest claim (#{dest_shop}). " \
      "The dest claim will be used. The shop parameter is deprecated for exchange_token " \
      "and will be removed in a future major version."
    )
  end
```

lib/shopify_api/utils/shop_validator.rb
```diff
@@ -0,0 +1,31 @@
+# typed: strict
+# frozen_string_literal: true
+
+module ShopifyAPI
+  module Utils
+    class ShopValidator
```
lizkenyon: I think we need to modify the ShopValidator check to protect against URLs like "attacker.com/.myshopify.com". This passes validation since it ends with .myshopify.com, but URI.parse resolves the host to
  attacker.com

lizkenyon: This could probably be module similar to [hmac ](https://github.com/Shopify/shopify-api-ruby/blob/main/lib/shopify_api/utils/hmac_validator.rb)

zzooeeyy: There's also a similar sanitize method from [shopify_app](https://github.com/Shopify/shopify_app/blob/main/lib/shopify_app/utils.rb) gem, I wonder if the logic can be extracted so we don't have to maintain 2 packages for validating Shopify URLS

lib/shopify_api/auth/token_exchange.rb
zzooeeyy: I also agree! We should remove `shop` from this method's arguments

lib/shopify_api/utils/shop_validator.rb
```diff
@@ -0,0 +1,31 @@
...
+      SHOPIFY_OWNED_SUFFIXES = T.let([
+        ".myshopify.com",
+        ".myshopify.io",
+      ].freeze, T::Array[String])
+
+      class << self
+        extend T::Sig
+
+        sig { params(shop: String).returns(String) }
+        def validate!(shop)
```
zzooeeyy: `validate` might be a misleading method name since it does more than validation and returns the sanitized shop name

lib/shopify_api/auth/token_exchange.rb
gonzaloriestra: Done!

lib/shopify_api/utils/shop_validator.rb
gonzaloriestra: I've finally copied the code from there. Once we release a new version, we could use it from `shopify_app`, as it's already using `shopify-api-ruby`. Makes sense?

lib/shopify_api/auth/token_exchange.rb
```diff
@@ -21,12 +21,11 @@ class << self
 
         sig do
           params(
-            shop: String,
             session_token: String,
             requested_token_type: RequestedTokenType,
           ).returns(ShopifyAPI::Auth::Session)
         end
-        def exchange_token(shop:, session_token:, requested_token_type:)
```
lizkenyon, line 29: I would prefer we don't remove this in this PR. But rather log a warning that this will be remove in a future version. i.e we should deprecate now remove later.

This will allow us to group our breaking changes together.

```
  ShopifyAPI::Logger.deprecated(
    "The `shop` parameter for `exchange_token` is deprecated and will be removed in v17. " \
    "The shop is now always taken from the session token's `dest` claim.",
    "17.0.0"
  )
```

gonzaloriestra: Oh, sorry, I misread your previous comment. Updated!

lizkenyon: all good!

lizkenyon (approved):
TY for this! ⭐

gonzaloriestra:
@lizkenyon @zzooeeyy thanks a lot for your help! Do you want me to create a new release or how do you usually handle this?

PR #1459, opened 2026-08-24
Add App Events logging API

kristianpd:
## Description

Adds the Ruby SDK counterpart to `shopify-app-js` App Events logging:

- `ShopifyAPI.log` validates and posts one App Event to the Global API.
- `ShopifyAPI::Auth::GlobalApiClientCredentials.global_api_client_credentials` mints and caches the app-level JWT.
- `ShopifyAPI::Context.global_api_url` supports HTTPS-only non-production Global API hosts.
- Tokens refresh within a 60-second expiry skew and once after an event `401`.
- Concurrent stale-token failures reuse a replacement token, while rejected tokens clear conditionally from the cache.
- Documented `expires_in` and `scope` response fields take precedence over JWT claim fallbacks.

The request contract uses snake_case JSON, string shop IDs, the `2026-07` API version floor, and both actual and documented idempotent replay headers.

JavaScript reference: https://github.com/Shopify/shopify-app-js/pull/3368

## How has this been tested?

- `bundle exec ruby -Ilib -Itest test/app_events_test.rb`
- `bundle exec ruby -Ilib -Itest test/app_events/event_payload_test.rb`
- `bundle exec ruby -Ilib -Itest test/auth/global_api_client_credentials_test.rb`
- `bundle exec ruby -Ilib -Itest test/context_test.rb`
- `bundle exec rake test:library` — 396 runs, 791 assertions
- `bundle exec rubocop` — 119 files, no offenses

`bundle exec srb tc` cannot run on this Apple Silicon host because the locked `sorbet-static` binary is x86_64-only: `Bad CPU type in executable`. CI must provide the typecheck result.

Local Core smoke validation minted tokens, ingested events, and reported idempotent replays for `2026-07`, `2026-10`, and `unstable`. The local installation cache was primed because the Query Engine `apps.app_installations` view is unavailable in this environment.

## Checklist:

- [x] My commit message follows the documented pattern.
- [x] I have added tests that prove my feature works.


lib/shopify_api/context.rb
```diff
@@ -57,6 +63,8 @@ def setup(
         api_version:,
         is_private:,
         is_embedded:,
+        global_api_version: DEFAULT_GLOBAL_API_VERSION,
+        global_api_url: DEFAULT_GLOBAL_API_URL,
```
kristianpd, line 67: advice welcome here

PR #1463, opened 2026-09-20
Expose request ID after token refresh

### LINKED ISSUE #1453: Provide a way to access the request-id when refreshing an access token
## Background

When using `ShopifyAPI::Auth::RefreshToken.refresh_access_token` (`lib/shopify_api/auth/refresh_token.rb`) to refresh an offline access token (calling `POST /admin/oauth/access_token` with `grant_type=refresh_token`), we've run into an issue where **the request itself succeeds, but the `refresh_token` returned in the response has an expiry that is far shorter than the documented 90 days**.

- Reported on the Shopify Community forum:
  https://community.shopify.dev/t/refresh-token-returned-with-an-expiry-far-shorter-than-the-documented-90-days/35880

In that thread, a Shopify staff member asked whether we had the request-id from the response (since it would help speed up their investigation). However, the return value of `refresh_access_token` — a `ShopifyAPI::Auth::Session` — provides no way to obtain the request-id, so we were unable to provide it.

Looking at the current implementation of `refresh_access_token`:
https://github.com/Shopify/shopify-api-ruby/blob/8c72cd222d346472a29e51c7a7dfb6b7239f7456/lib/shopify_api/auth/refresh_token.rb#L18-L54

The `response` object returned by `client.request` presumably includes the response headers (which contain the request-id), but only `response.body` is used to build the `Session` — the response itself is never exposed to the caller. As a result, there's no way for callers to retrieve the request-id.

## Request

- Provide a way for callers to obtain the request-id when `refresh_access_token` is used to refresh an access token
  - e.g. include the request-id on the return value (e.g. `Session`)
  - or return an object that includes the response headers, allow passing a callback, etc. — any approach along these lines would work
- This should be available for both successful and failed responses (including cases like this one, where the HTTP request succeeds but the returned data itself is problematic)

najibna:
## Description

Fixes #1453

Keeps the existing `Session` return type and attaches Shopify's `x-request-id` to the session after a successful refresh. Failed refreshes expose the same ID on `HttpResponseError#request_id`.

This preserves backward compatibility instead of introducing a new result object.

## How has this been tested?

- `bundle exec ruby -Itest test/auth/refresh_token_test.rb`
- `bundle exec srb tc`

## Checklist:


najibna:
I have signed the CLA!

